Issue #5, opened 2015-01-07
Offer a choice of 32bit vs 64bit downloads


danrabbit:
Current only 64bit is available. Should get both choices

cassidyjames:
Two buttons, or a radio button, or what? :D

cassidyjames:
After discussion, we're going to hold off on 32-bit, at least on the homepage. 64-bit is the recommended download, and we can update our docs to reflect that.

Issue #1, opened 2015-01-06
Design responsive Freya download page


cassidyjames:
First step. :)

cassidyjames:
@danrabbit says he's going to basically play this by ear and tackle it issue-by-issue.

PR #7, opened 2015-01-07
single-page foundation

### LINKED ISSUE #2: Get a site favicon
Gotta have one of those

KeitIG:
Just minor but useful commits before we can seriously work on #1 

fix #2 and #5 

When merged we should continue committing on this branch instead of deleting it.


danrabbit:
Let's revert the changes to "a". Changing the buttons to an anchor with button class is fine. But I think we'll want to eventually have regular links in addition to buttons. So I don't think we should use inherit color or anything like that.

CSS selectors should each have their own line.

Let's also please use 4 spaces instead of tabs to be consistent with desktop dev style.

KeitIG:
As @danrabbit said, this branch contains too much unrelated things, closing it.

PR #13, opened 2015-01-07
add basic footer

danrabbit:
This branch adds a very simple footer. It's required to add meta charset utf-8 to use the copyright character


lewisgoddard:
Looks good, although I usually encode © as `&copy;`.

Also, most countries recommend the format **Copyright © YYYY Owner**.

PR #14, opened 2015-01-07
link font Open Sans

danrabbit:
Link "Open Sans" since we use that

### DIFF
--- a/index.html
+++ b/index.html
@@ -4,8 +4,8 @@
         <meta name="viewport" content="width=device-width, initial-scale=1.0">
         <title>elementary</title>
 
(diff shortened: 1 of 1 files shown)


KeitIG:
Well, It's in https://github.com/elementary/mvp/commit/a9d8134f39bb4928ab04bfb0fd36202b88738b12

lewisgoddard:
Everything else uses `"` rather than `'`.

Also, you might want to match the `rel="stylesheet" type="text/css" media="all"` from above, and perhaps put it before the internal styles? Should save rendering time having the font set before referencing it.

PR #16, opened 2015-01-07
Add Favicon and Link. Also fixes relative style reference.

### LINKED ISSUE #2: Get a site favicon

lewisgoddard:
Fixes #2


KeitIG:
Well, I'm not sure using a .ico instead of a .png for the favicon is the good way nowadays, it is a bit "old"

lewisgoddard:
ico is and alway has been the standard. Older IEs will appreciate it and modern browsers won't care either way. GIFs are also well supported, that doesn't make it a better format. PNGs only really come into their own for large files with text that needs to be sharp.

PR #29, opened 2015-01-09
Remove outline for button

### LINKED ISSUE #28: Implement custom focus outline style for button
https://github.com/elementary/mvp/blob/master/styles/main.css#L106 - Not good for people navigating with a keyboard. I suggest we remove this line.
```css
outline: none
```

KeitIG:
fix #28

### DIFF
--- a/styles/main.css
+++ b/styles/main.css
@@ -103,7 +103,6 @@ button {
     transition: all 100ms ease-in;
 
     -webkit-appearance: none;
-    outline: none;
 }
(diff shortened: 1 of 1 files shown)


danrabbit:
I don't want to just re-add the outline as it is default in the browser. It looks bad. We should come up with a focus outline style.

danrabbit:
Closing this since we fixed it already by creating a custom focus style

PR #46, opened 2015-01-10
Carousel

jacobp100:
Needs designer direction.


fabianthoma:
Animation with clip-path does not work with Firefox 34.0.5 (newest available). Also I'm not sure we are looking for that kind of effect, but I'll leave the design input to @danrabbit

jacobp100:
Yep. The animation only works in Chrome. It still works as a regular carousel in FF, Safari, and IE9+. Functions as a wall of images in IE8-.

danrabbit:
Please fix merge conflicts with trunk
As this contains dummy content, it is not ready to be proposed for merging
I would also suggest merging small unrelated changes (like putting the logo in a <heading>) separately.

jacobp100:
How are there merge conflicts when it’s not even being merged‽
I don't have any screenshots to replace the dummy content with. We were waiting for them on the old website too.
It's only the miscellaneous title thing that's changed—it’s probably fine to just include.

danrabbit:
Well this has been proposed to merge as a PR. You can see on the PR page that this branch cannot be merged due to conflicts.

Right, which is what I'm saying. It's not acceptable to push dummy/incomplete content into master.

Only related changes should be merged together. It's not really okay to create PRs with unrelated changes.

index.html
```diff
@@ -11,10 +11,39 @@
     </head>
     <body>
         <div class="grid-margin">
-            <img alt="elementary OS" id="logotype" src="images/logotype.svg">
+            <heading>
```
lewisgoddard: `heading` is not a valid, semantic, html element.

index.html
```diff
@@ -24,5 +53,6 @@
                 <li><a href="https://github.com/elementary" target="_blank"><i class="fa fa-github"></i></a></li>
             </ul>
         </footer>
+        <script src="js/carousel.min.js"></script>
```
lewisgoddard: If you're only now loading the script, what does the content look like before? How does it look without JavaScript?

styles/main.css
```diff
@@ -34,8 +34,6 @@ h3 {
 **********/
 
 footer {
-    position: absolute;
```
lewisgoddard: Why the need to stop positioning the footer?

index.html
jacobp100: Displays all images at once for IE8 and below.

styles/main.css
jacobp100: Got in the way. Should be solved by flex box (http://philipwalton.github.io/solved-by-flexbox/) anyway.

danrabbit:
Going to close this PR since there's been no activity. Feel free to propose again when the branch is ready

PR #54, opened 2015-01-11
Added basic Stripe functionality

khonquer:
JS is inline. If a zero amount is specified the download link will be opened directly. If a non-zero amount is specified the Swipe dialog will open, and upon completion the token callback will direct the user to the download page. Regexp are used in order to remove non-digits as well as test for a non-zero amount.

The CSS for the donation box is copied from .form-control from bootstrap. I didn't want to dictate the style, but can change if you guys have any suggestions.


danrabbit:
I would prefer that we have text input styles separated from what is actually specific to that particular input. I've just proposed #55 which adds styles for inputs.

"elementary OS" is always written as two words.

"/square-image.png" is referenced, but I don't see the image added in this branch

I'm sure @fabianthoma will have some input about the server side code required for us to process payments.

Thanks for your work on this :)

khonquer:
Alright will do. 

It is not complete yet, I will be adding some form stuff from the docs to make it easy to finish whenever you guys actually go forward with it.

khonquer:
All the above was done and I have cleaned it up a little bit.

Popup blocking will only allow opening a tab as a result of user actions, which is not the case since after donating. So there are a few options and I could implement any one of them.

1) Separate Donate and Download buttons would be the cleanest option.
2) Change the current page to the download link upon donation.
3) Open the download page before the user finishes the transaction.
4) If you want to separate the donate and download buttons, the download button could be disabled until a zero amount is entered or the transaction is completed. If you want one button, the first click could trigger the donation, upon completion the user would land back on the page, a thank-you message would be displayed, and the next click on the download button would result in opening a new tab with the download link.

The 1st option would allow the user to finish donating without distracting them with the download link. The 2nd option would take the user away from your site. The 3rd just feels wrong since it would probably be best to ensure donation before download is available. The 4th just seems too complex, but with a little bit of tweaking I think the Thank You could work.

Let me know what you would prefer @danrabbit and I will submit it asap.

fabianthoma:
Hi, coincidentially we both worked on this implementation.
I have a few issues I'd like to point out that make this unmergable to me, however most of the work looks good.
1. You added jquery to the site, we wanted to avoid that, at least for now.
2. You added a lot of inline javascript, that should really go into a file and be more modular.

So now on to how I'd do popup blocking.
Since we had a lot of issues with this on the current site as well, I'd go with opening another overlay window looking like the stripe one and let the user choose his download options in there, for example 32bit vs 64bit and torrent vs direct download. Then when the user clicks one of those buttons the action will be user triggered and will work perectly fine.

I've pushed my own work to here: d4c01b6e8a0cd86fbb9d24e1c418730c01897255

khonquer:
Alright, I think that might make you happy.

khonquer:
@fabianthoma I saw your issue for implementing a torrent download with https://github.com/feross/webtorrent 

Is it officially what you want? I can do this for you if you want.

cassidyjames:
@khonquer I don't believe we're going to do WebTorrent. We can offer up a .torrent file for those who prefer that download method and just link everyone else to the download via SourceForge.

cassidyjames:
I believe this got thrown off by us implementing our templating system. Please rebase off of master.

khonquer:
Alright, how does that look?

I also made a change that someone else (who is decent with CSS) should do correctly.

In order to have the overlay-content hide/appear I would toggle between display: none/inline-block. I used an inline-block to set the width and height dynamically. I switched it to be set to a block instead of inline-block because I do not know if this is correct.

danrabbit:
According to GitHub it still can't be merged, unfortunately.

khonquer:
Sorry for the absence. I fixed the CSS merge issues.

danrabbit:
Hey Khonquer,

We've started working together in this branch: https://github.com/elementary/mvp/tree/stripe-payment

Can you propose any changes into this branch now?

danrabbit:
I'm closing this PR since we've just merged basic stripe stuff into master.

PR #55, opened 2015-01-11
Input style

danrabbit:
This branch adds a style for text inputs. This is a pre-requisite to adding a payment amount text input


fabianthoma:
Looking good, love that slight transition between states

PR #58, opened 2015-01-11
fix-52: added 404.html

### LINKED ISSUE #52: Create 404 page
What it says on the tin. Need a 404 page with a way to get back to home

PhillipSz:
use most things from old page. Fix #52

### DIFF
--- a/404.html
+++ b/404.html
@@ -13,7 +13,7 @@
     </head>
     <body>
         <div class="grid-margin">
-            <p class="big-label">404</p>
+            <p class="error-label">404</p>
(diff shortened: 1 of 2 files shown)


danrabbit:
Can you revert the change to h1 font size and weight and instead add "opacity: 0.8;"? This would be more like what we see on desktop.

Also let's change to the page title to "Page Not Found | elementary OS" instead of 404.

Make sure you're using "row" and not "grid-margin" as this has changed in master

cassidyjames:
Port to the templating system, please. :)

PR #85, opened 2015-01-12
Legacy warning

cassidyjames:


cassidyjames:
@lewisgoddard here you are

PR #90, opened 2015-01-13
move button row inside alert row

danrabbit:
This branch reduces the amount of padding between the button and alert body on the 404 page


KeitIG:
Seems right to me.

How about add a margin top of 30~35vh to the content container for this page ?

```html
<div id="content-container" class="page-not-found">...</div>
```
```css
.page-not-found {
    margin-top: 30vh;
}
```

PR #96, opened 2015-01-13
Add Alternative Downloads page (Fixes #89)

### LINKED ISSUE #89: Alternative downloads page
We're going to want an "alternative downloads" page to offer 32-bit and torrents.

lewisgoddard:


danrabbit:
Looks like you've added a class "p.error-label" that you're not using

lewisgoddard:
Ah, `.error-label` got removed after I branched.

danrabbit:
Instead of .button-alternative can we use .button.flat ? This would be inline with what we do on desktop

lewisgoddard:
Oh, wait, I just realised I didn't put the links in.

danrabbit:
Cool and then my next question is how to people get to this page? :p

lewisgoddard:
@danrabbit For now I've just added a link in a `p.small-label`

We can style links globally later, the color seems a little sharp.

danrabbit:
I'd like to preserve the blue highlight for the button focus state please :) Maybe we can add "border-color: #3d9bda;" to that focus state to make sure the highlight looks okay for regular buttons.

I'm -1 on making the download size less accurate.

I'm also -1 on adding styles that you're not using in this branch. Specifically I don't see where you're using the text-left/center/right

Thoughts on using a linked button with the magnet icon a la isohunt? http://fortawesome.github.io/Font-Awesome/icon/magnet/ http://isohunt.to/torrent_details/12438824/ubuntu-14-04-1-desktop-amd64-iso

lewisgoddard:
You want a blue outline on a grey button when it's focused? That seems odd, but I do agree maybe the grey doesn't stand out quite enough.

I can trim out those styles, I hadn't noticed I'd stopped using them (I was at one point).

I originally had that Icon in there, but it looked very heavy next to the slim button font. I can push it if you like and see what you think.

danrabbit:
Yes I want to keep the blue focus highlight. I don't think it's particularly odd. Check the focus highlight for the text boxes on GitHub :p 

Yeah I was just thinking as a linked button it would cut down on the number of buttons here.

To be honest though, I can't help but think this doesn't really need to be a whole page for only 4 buttons. I think a menu/popover/dialog would probably do the trick.

lewisgoddard:
I was thinking that it could also serve as an archive of old releases, but I now realize that you probably wouldn't be so silly as to offer up old releases.

danrabbit:
Yeah I think we solved that in slack :)

We can refine the design as we go. But if you want to fix those other few issues from the review (unused classes, revert changing the file size, etc) we can merge

PR #103, opened 2015-01-18
Stripe w input

danrabbit:
Add some (stolen) code for variable input


fabianthoma:
Awesome, I'll do a bit of tweaking to it, but this is great for now

PR #110, opened 2015-01-19
Payment buttons

danrabbit:
This branch is the beginning of the Humble-style Pay-what-you-want bar thingy. The styling is mostly taken care of here. The logic just needs to be hooked up. We probably want to also add validation. But here it is :)


lewisgoddard:
Use `.target-amount` for targeting and move the `.checked` class for selection.

Only move `.checked` to "Custom" if the user inputs a value.

You can then just fetch the value of `.target-amount.checked` and parse it.

The logic already strips non-number characters, but needs better float handling (fun).

PR #111, opened 2015-01-19
Add logos to quotes. Fixes #106

### LINKED ISSUE #106: Use logos for press quotes
Would be super cool and trendy to use their logos instead of plain text

TuurDutoit:


danrabbit:
I think we want to use monochrome logos and preferably SVG so that they display correctly for HiDPI displays.

danrabbit:
Also, I think it might make more sense to select ".column.third img" than to create a new class. Presumably we'll have other 3rd-width columns with images up top

emersion:
Just found a SVG for Wired: http://commons.wikimedia.org/wiki/File:Wired_logo.svg

For Lifehacker and Mac Life, seems that we'll need to make our own SVGs.

PR #113, opened 2015-01-19
"64 bit" becomes "64-bit"

maliktunga:
Seldom have I seen that term without a hyphen.


danrabbit:
Google trends disagrees http://www.google.com/trends/explore#q=64%20bit%2C%2064-bit&cmpt=q&tz=

Sorry, "Won't Fix"

maliktunga:
@danrabbit I don't want to sound pedantic. However, I'd like to see at least a single respectable instance of "64 bit" being used instead of "64-bit". See [this Wikipedia article](https://en.wikipedia.org/wiki/64-bit_computing) for example.

maliktunga:
@danrabbit Oh, and look at the news headlines in your Google Trends example. Trends does not care about whether a hyphen is present or not.

lewisgoddard:
Shouldn't the alternative downloads page also have been changed?

maliktunga:
@lewisgoddard I'm working on that right now ;-)

PR #119, opened 2015-01-20
Flat notebook hero

### LINKED ISSUE #23: Add a screenshot
Need at least one sexy screenshot of the default Freya desktop.

Might be interesting to do a shot with a few different sexy computers running Freya.

Would be very warm and indie of us to have kind of a set up "natural" shot like the new photos on system76

danrabbit:
This is basically intended to be replaced at some point. But it gives us something to work around.

fixes #23


cassidyjames:
Any way we can be less Macbook-y? I like it but we're not going to hear the end of "Hurr durr OS X ripoff" if we use a Macbook render, as sexy as it is, on our home page.

PR #120, opened 2015-01-20
Update alternative-downloads.php

muriloventuroso:


danrabbit:
This PR no longer applies since we switched the page out for a dialog

Issue #9, opened 2015-01-07
Installation Instructions for Ubuntu


danrabbit:
How to burn & install from the latest LTS of Ubuntu (LTS for convenience of not having to update screenshots every 6 months)

emersion:
I take this one :-)

Issue #11, opened 2015-01-07
Installation Instructions for Mac OS


danrabbit:
Likely we'll want to use Sam's installer? https://github.com/sdaitzman/elementary-os-install-utility

Issue #10, opened 2015-01-07
Installation Instructions for Windows


danrabbit:
How to burn and install from Windows

PR #121, opened 2015-01-20
Installation pages

emersion:
Adds installation manual.

- [x] Copied download instructions to `installation.php`
- [x] Copied installation instructions for Ubuntu, Windows and Mac OS X
- [x] Current OS auto-detection for installation instructions
- [ ] ~~Update screenshots~~
- [x] Add toggleable buttons for CD/USB and PC/Mac
- [x] Remove "continue to Booting the Install Disk" links
- [x] Nice transitions?


emersion:
Do we need to make multiple pages, as on the current website?

danrabbit:
Not necessarily. What's most important is that we get the relevant information to people in the easiest to digest format possible.

Something I was thinking about would be detecting the OS and showing the relevant directions. Of course we'd still need a way to manually choose which directions you want in case you're installing on a different device than what you're viewing from.

danrabbit:
* The link "download the elementary OS disk image from our home page." should probably open in a new tab/window

* We should only mention Install methods that we have directions for. Currently we mention both CD and USB, but only have directions for CD

* CD/DVD drive should not be listed as a requirement. I don't have one on my current computer ;)

emersion:
> CD/DVD drive should not be listed as a requirement. I don't have one on my current computer ;)

So should I remove it from the list, or add _CD/DVD Drive or USB port_ ? Do I remove _CD/DVD Drive or USB port_ from _Recommended System Requirements_?

danrabbit:
The only reason I could think to put that is to say that your computer should be able to boot from external media. I think in most cases just about every computer should be able to boot from external media. So I don't think it's necessary

emersion:
That's right, just removed both.

Do you think it's a good idea to add some fast-and-discrete animations in CSS when toggling OSes?

danrabbit:
That probably wouldn't be an issue :)

I'm wondering if we should use 3 .linked buttons to toggle between the different OS's? That way the choice is a little more prominent and we're not moving them around when we change sets of directions.

emersion:
One more thing: should I add the [_Booting the Install Disk_](http://elementaryos.org/docs/user-guide/booting-the-install-disk) on the same page? Keeping the current layout, this will be like a long and boring install guide, but if using buttons I think it's reasonnable to add this last part.

Also, should I add some `.linked` buttons to choose between CD and USB install?

danrabbit:
Yeah that would be smart I think. We can use linked buttons to keep the guide shorter by having the choices be "horizontal" :)

Scrolling is fine, I think

danrabbit:
If we can't use a.button:active directly we should probably add a "a.button.active" and use that as the :checked state instead of using .suggested-action

danrabbit:
Btw, I don't really expect this branch to have perfectly accurate information. We can make adjustments after we merge. But in case you want to dig into that, Avi has a guide to installing Freya on a Mac https://github.com/aroman/freya-on-a-mac

emersion:
All right, I just need to toggle paragraphs when clicking on buttons for CD/USB and PC/Mac.

So, we could merge Avi's guide into this installation guide? Sounds quite technical, but it would be useful. And what about UEFIs? (should we merge https://github.com/skylerseverns/elementary-os-freya-uefi)

danrabbit:
Yeah, maybe we should wait until after we merge this and then we can decide about guides for EFI (PC or Mac)

emersion:
Okay! Just added toggles for other install options (CD/USB + PC/Mac) :-)

The only remaining thing is a screenshot update for Ubuntu (screenshots there are very ugly). I have no Ubuntu installation for the moment (just Arch with GNOME and Elementary), so I'll do it tomorrow (something nicer on a white background with window decorations).

danrabbit:
* The linked buttons are too wide for mobile. You might want to end the row between sections and have the linked buttons exist outside of that so you don't get a double margin.

* The images need to have a max width set so that they fit on mobile. We can probably just add "img { max-width: 100% }" since I don't think we'll ever want an image to be wider than the viewport.

@cassidyjames is around Ubuntu machines all day. We can probably get him to do some updated screenshots after this branch hits master :)

emersion:
Just fixed issues for small screens, and added some transitions. ;-)

danrabbit:
Noting that this closes #9 #10 and #11 

* Can we remove the references to Answers and the live chat since we're not sure these will exist on the new website?
* What do you think about putting the two system requirements in columns so that they are side-by-side?
* Can we move all CSS into classes? I'm not sure we should set so much css directly in js.
* Can we make the animation time shorter? I think we use like 200ms for these swipes on desktop.
* Let's change the text for usb to just "Create a USB". I think it's safe to assume both forms of media should be bootable ;)
* The "Burn" instructions don't have a default on mobile. Since most of our downloads come from Windows, I think that would be a safe default.

Also, please be aware that GitHub can no longer automatically merge your branch. It will need to have master manually merged in to remain compatible.

emersion:
Just fixed all remaining issues and merged with upstream work. You should be able to automatically merge this branch now. ;-)

danrabbit:
Awesome! One more small change: On the home page, in the download dialog, let's add a link to your guide. So the dialog would be something like:

"We recommend 64-bit for most modern computers. Read the installation guide here" Where "here" is a link to your guide that opens in a new tab/window.

Thanks again!

danrabbit:
I'm going to go ahead and merge and then link after :) Thanks!

PR #125, opened 2015-01-21
Add Modal with Download options.

lewisgoddard:


danrabbit:
I've submitted a PR to adjust the styles.

I do think we need to add a bit of JS to block scrolling behind the dialog. Something like the following:
```
$('.js-target-jquery-leanmodal-overlay').on('scroll touchmove mousewheel', function (e) {
    e.preventDefault();
    e.stopPropagation();
    return false;
});
```

lewisgoddard:
I also think we need to stop modals being bigger than the viewport.
```
.modal {
    max-height: 100vh;
    overflow: auto;
}
```

danrabbit:
I agree about modals needing to have a max size. However keep in mind that vh is currently broken on Safari iOS7 and I'm not sure height is the more concerning attribute. Might be better to do max-width 100%?

We talking in Slack about the code. I think we need to do both since overflow: hidden doesn't solve mobile but my code doesn't solve desktop.

danrabbit:
Can we add margin: 16px; to .modal? Just realized it's bumped up against the edge on mobile

lewisgoddard:
jQl catches jQuery functions and executes them when jQuery is loaded (something else it does). Otherwise, you either have to load jQuery in a load-blocking manner, or any jquery scripts that run before it is loaded will simply error.

PR #130, opened 2015-01-22
adjust modal styles

danrabbit:
This branch makes some adjustments to modals to make them appear like dialogs on desktop.


lewisgoddard:
What do you think about adding this?

```
button.suggested-action,
.button.suggested-action {
    font-weight: bold;
}
```

danrabbit:
I think that is unrelated to this branch :p 

But yes there is a font-weight problem with buttons we should address somewhere else

lewisgoddard:
Fair enough, it is the button that launches the modal, and in the modal though. :stuck_out_tongue_closed_eyes:

PR #131, opened 2015-01-23
Modal fix

danrabbit:
* Target the action area of modals more accurately with .action class
* Move the modal out of the regular layout. It doesn't exist inside the downloads row; It doesn't make sense for it to be there in HTML.
* Fix a regression from trunk where the hero image is squished inside a row
* Remove an unnecessary row from the modal.
* Remove unnecessary options from leanModal
* Add margin and max-height


lewisgoddard:
If you remove the top option, it will set it to 100px every time. That will look silly on mobile.
Don't we want a close button? It seems odd not to have one?
And the padding seems to small on the modal.

PR #140, opened 2015-01-23
A custom input disguised as a button, among others.

lewisgoddard:


fabianthoma:
Looking good, let's do seperate requests for seperate things in the future. :wink:

PR #151, opened 2015-01-23
"amd64" becomes "AMD64" [See comment]

maliktunga:
It's the correct spelling. However, the term "x86-64" is also commonly used (https://en.wikipedia.org/wiki/X86-64).

### DIFF
--- a/installation.php
+++ b/installation.php
@@ -12,7 +12,7 @@
                 <div class="column half">
                     <p>Minimum System Requirements:</p>
                     <ul>
-                        <li>1 GHz x86 or amd64 processor</li>
+                        <li>1 GHz x86 or AMD64 processor</li>
                         <li>512MB of system memory (RAM)</li>
                         <li>5GB of disk space</li>
                     </ul>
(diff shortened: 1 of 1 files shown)


lewisgoddard:
It should either be i386 and amd64, x86 and x86-64, or 32-bit and 64-bit.
This is an issue of mixing referencing styles, not incorrect spelling.

maliktunga:
@lewisgoddard Do you have any reference for that? The only correct capitalization I see is with capitals.

cassidyjames:
Hm, there might be a bigger issue here. We're suggesting 64-bit by default. We might just say 64-bit is required and add a note about 32-bit and the alternative download?

lewisgoddard:
AMD64 was back when AMD (the company) had the total monopoly of AMD64 (the architecture). Since i386 is rarely referenced as Intel-386, I see no reason why amd64 should be capitalised.

Regardless, we're still mixing how we're referencing architectures.

danrabbit:
Yeah for simplicity sake we should just always use "32-bit" and "64-bit"

maliktunga:
I agree. It should be 32-bit and 64-bit. "x86" and "amd64" are too technical for how simple and straightforward elementary OS strives to be. 

Can I change the pull request to changing the terms to "32-bit" and "64-bit" then?

maliktunga:
Here's the new pull request: https://github.com/elementary/mvp/pull/154

I can close the present request now.

PR #153, opened 2015-01-23
Don't prompt twice for payment, unless they didn't pay. Fixes #68

### LINKED ISSUE #68: Users shouldn't be prompted to pay twice.
After paying, set a release-specific cookie (ike `has_paid_freya`) and check it in future to avoid prompting them to pay again. Use a special url in receipts to set the cookie then too.

lewisgoddard:
Fixes #68


cassidyjames:
Getting Stripe error when testing. 
screenshot from 2015-01-24 05 53 12

lewisgoddard:
You'll have to clone it and add a Stripe config file if you want that to work. Branches don't get a copy of it by default because of `.gitignore`. Or you could just run the cookie creation and refresh `$.cookie('has_paid_freya', amount, { expires: 3650, path: '/' });`

danrabbit:
I can confirm that setting the cookie manually works as expected.

I'm curious about why the cookie needs to know how much they've paid?

jacobp100:
Use localStorage over cookies. Then you don't need jQuery-cookie.

lewisgoddard:
@danrabbit We could just set it to one, but I thought it might be useful later.

@jacobp100 localStorage isn't sent server-side. To avoid jumps, buttons are never even loaded in PHP.

scripts/homepage.js
```diff
@@ -68,6 +68,11 @@ function process_payment (amount, token) {
     payment_http.send('amount=' + amount + '&token=' + token.id);
 }
 
+function payment_complete (amount) {
```
fabianthoma: Would it not be better to set the cookie from the stripe payment hook we have serverside? We have that http request anyway, and setting the cookie there seems more logical to me.

lewisgoddard: I don't suppose it makes much difference, and it would remove the jquery.cookie dependency. Just be sure to set a far-future date (currently 10 years).

fabianthoma:
Looking good with the new commits, can we get an update to make it mergable to master?

fabianthoma:
Alright, merged.

PR #160, opened 2015-01-23
"Install Disk" becomes "Installation Media"

maliktunga:


danrabbit:
I'm not sure about "an Installation Media". Maybe "Creating Installation Media" makes more sense there? It might be worth some discussion if we really want "Media" instead of "Disk"

In either case, I think the ID should be updated to match.

maliktunga:
@danrabbit Oh woops, I'm really sorry the commit got attached to the "1GB" pull request. That was unintended.

I'm pretty sure you don't want "Disk", since a USB stick is not a disk. I agree "a Media" in English must sound wrong. How about "an Installation Medium"?

danrabbit:
Well a "disk" with a "k" I feel like can be a USB or just any sort of storage media really. "Disc" with a "c" no. I'm not sure "an installation medium" is better than "an install disk" :p But maybe @cassidyjames has some thoughts on that.

maliktunga:
@danrabbit I don't think so: https://en.wikipedia.org/wiki/Disk_storage

"[D]ata are recorded by various electronic, magnetic, optical, or mechanical changes to a surface layer of one or more rotating disks." That excludes USB flash drives.

lewisgoddard:
This needs a merge or rebase from master, I can't merge it automatically.

maliktunga:
@lewisgoddard Right, working on it.

maliktunga:
@lewisgoddard Meh, I couldn't automatically rebase it from master either. I created another pull request in another branch: https://github.com/elementary/mvp/pull/180

PR #167, opened 2015-01-24
Rename scripts/installation.js to js/installation.js

emersion:
There are two folders with JS files at the moment.


lewisgoddard:
My pull request #153 actually moves homepage.js from `js/` to `scripts/`, to match the way we stored css, which sits in `styles/`, not `css/`.

Issue #132, opened 2015-01-23
Block scrolling while modal is open


danrabbit:
We need to block scrolling of the body while the modal is open. This includes:

* Two finger touchpad
* 'touchmove'
* arrow keys
* 'mousewheel'
* Clicking and dragging the scroll bar

I have a bit of code from the ChiFi site, but for some reason it doesn't seem to work with leanModal

```
$('.dialog-background').on('scroll touchmove mousewheel', function (e) {
    e.preventDefault();
    e.stopPropagation();
    return false;
});
```

lewisgoddard:
This works for me to stop scrolling with my touchpad, but not the scrollbar or arrowkeys because it doesn't try to catch those.

```
$('.js-target-jquery-leanmodal-overlay').on('scroll touchmove mousewheel', function (e) {
    console.log('Scroll Ignored');
    e.preventDefault();
    e.stopPropagation();
    return false;
});
```

PR #168, opened 2015-01-24
Begin removing jQuery

jacobp100:


lewisgoddard:
You should be able to set a zero amount in custom to get elementary OS for free. That seems to be broken in your branch at the moment. Also, because you're fetching the value in vanilla instead of jQuery, you're going to get rounding errors. Try putting `36.66` into the custom box. You'll get `36.65` out.

I'm going to be honest here, "don't re-invent the wheel". I don't seem the point in coding a lot of things to replace work already done elsewhere. You're already making the code grow in size, and modals are generally more hassle than they're worth.

However, should you want it, here's [the uncompressed source for leanModal2](https://github.com/eustasy/jQuery.leanModal2/blob/master/jQuery.leanModal2.js).

lewisgoddard:
Also, why are the button types set to submit? They don't submit data. In fact, the only button you haven't done that to is the one that does!

jacobp100:
Oh, my bad—completely absent minded that it wasn't a form. In which case, the index.php changes definitely need to make it, as button defaults to type=submit.

jacobp100:
Cannot reproduce 36.66 rounding.

jacobp100:
I've got a branch to remove the strict dependence on JS and CSS in the payments form. https://github.com/elementary/mvp/tree/Use-forms
However, I need to add JS to it. Can I know if this accepted/rejected asap, so I can continue the work on that branch?

lewisgoddard:
I'll personally accept this so long as it all works, but we still want to have all the functionality, and I think jQuery will just end up being added back at some point for something or other, like ajax requests to send emails, they were coming.

Keep in mind we want to add a scroll-blocker when the modal is open (#132), and I'm unsure about vanilla bindings for `scroll touchmove mousewheel`.

The rounding occurs in Chrome 40 on Luna.

Currently classList support is IE 10+
You can fix IE 9 by loading https://cdn.jsdelivr.net/g/classlist for < IE 10

jacobp100:
If you're set on using jQuery, by all means reject this. I think if we're adding a lot of polyfills, it becomes less worth the removal. Maybe we could do a custom lite build of jQuery when the time comes.

emersion:
As stated in the `README`, we do not support IE<10 ;-)

lewisgoddard:
Well, we've just added a classList polyfill for installation.js, which can be removed if we stick with jQuery and convert the code there. I also think capturing scroll does not work well cross-browser, and in the past Ajax requests have been horrific, so I'm against.

That said, I did introduce jQuery, so I guess we should get a second opinion.

emersion:
I worked on the install guide's JavaScript, and I've to say that it is not very pleasant to use directly the DOM api. This makes the code less maintainable. In another hand, using jQuery slows down page loading and it's a quite heavy library (we will never use all jquery's features on this website).

One thing to do is to use jQuery 2.*. It is the same as 1.*, but they dropped support for IE<9.

lewisgoddard:
Using conditional comments we can load jQuery versions. Note that IE 10+ ignores conditional comments and works like a regular browser.

```
<!--[if lt IE 9]>
    <script>jQl.loadjQ('//cdn.jsdelivr.net/g/jquery@1.11.2,jquery.cookie);</script>
<![endif]-->
<!-- Else, just play it cool and load stuff -->
<!--[if IE 9]><!-->
    <script>jQl.loadjQ('//cdn.jsdelivr.net/g/jquery,jquery.cookie');</script>
<!--<![endif]-->
```

emersion:
Since we're not supporting IE<9, I don't think loading jQuery 1.x is necessary.

lewisgoddard:
Fair enough, in that case we are already using jQuery 2, as the following URL combines and loads all latest versions of the projects automatically. http://cdn.jsdelivr.net/g/jquery,jquery.leanmodal2,jquery.cookie

PR #175, opened 2015-01-24
Img h1

danrabbit:
This branch more closely styles the logo images like h1 (what they were as text).
* Add a img.h1 style
* Move img styles into one location in the stylesheet
* Get rid of new .column.third img style
* Get rid of br tags

### DIFF
--- a/index.php
+++ b/index.php
@@ -28,15 +28,15 @@
             </div>
             <div class="row">
                 <div class="column third">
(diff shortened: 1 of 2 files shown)


danrabbit:
Before: screenshot 24-01-2015 13 27 11

After: screenshot 24-01-2015 13 27 24

PR #176, opened 2015-01-24
Adds classList support in IE<10 on install guide

emersion:
`classList` is not supported in IE < 10. This adds support for IE 8 and 9.

### DIFF
--- a/installation.php
+++ b/installation.php
@@ -162,6 +162,7 @@
                 </div>
             </div>
 
+            <!--[if lt IE 10]><script type="text/javascript" src="https://cdn.jsdelivr.net/g/classlist"></script><![endif]-->
             <script type="text/javascript" src="scripts/installation.js"></script>
 <?php
     include '_templates/footer.html';


danrabbit:
We have a big warning on the homepage that we don't support IE<10...

emersion:
Ow, sorry about that. @lewisgoddard talked about that in another issue, so I thought we were supporting it :-/

emersion:
> IE 9+ Compatability

?

danrabbit:
You're right! I misread it xD

emersion:
Okay, I test in a VM and merge if it works well (never sure that it will with IE :-P ).

lewisgoddard:
I can test this in about 12 hours. I have a lovely test environment set up just for this, but it's at work. However, based on the diff, this looks good.

emersion:
I've the same _lovely test environment set up just for this_! :-P

Just tested, it works well.

PR #189, opened 2015-01-25
Local modal js

### LINKED ISSUE #132: Block scrolling while modal is open

danrabbit:
This branch pulls in leanModal locally for some fixes
* Center modals
* Don't hardcode a bunch of styles in js
* Set overlay opacity to 0.6 to match stripe
* Remove now unnecessary code from homepage.js

Fixes #132


scripts/jQuery.leanModal2.js
```diff
@@ -0,0 +1,91 @@
...
+                    if ( $(this).attr('href') ) {
+                        var modal_id = $(this).attr('href');
+                    // IFHREF Fall back to if no href data-open-modal.
+                    } else if ( $(this).attr('data-modal-id') ) {
+                        var modal_id = $(this).attr('data-modal-id');
+                    } else {
+                        return false;
+                    } // IFHREF
+
+                    // Set the function to close the overlay if you click it.
```
lewisgoddard: I'm not sure if we want clicking on the overlay to close the modal. If not, we want this gone.

danrabbit:
Fixed :) So no accidental closes.

I'm hesitant to add a rule that centers the dialog higher just because I don't want modals to go off screen on mobile. Perhaps that can be done in another branch if it's really important.

lewisgoddard:
I still think it's important, but I agree mobile could be an issue.

Issue #44, opened 2015-01-10
Get Involved


danrabbit:
Either a new page or section needs to be created to explain how to contribute.

This section should include links to Launchpad, GitHub, and Bountysource at a minimum.

danrabbit:
Mockup from Harvey: https://github.com/harveycabaguio/elementary-Site-Mockups/blob/master/Get%20Involved/Desktop.svg

emersion:
I'll be happy to start to work on this tomorrow, if nobody starts before :-)

danrabbit:
I think today I will try to get a simple navigation set up. That seems to be kind of a pre-requisite to me ;)

One thing I think we have to be careful with here is that many of the pages illustrated here may not exist. So we should be careful to make sure that we're only building things that are real.

PR #196, opened 2015-01-26
Get involved page

emersion:
Implements the _Get Involved_ page. See #44.

Result here: http://beta.elementaryos.org/branch/get-involved/get-involved

Currently WIP. Styles are for now in a separate CSS file to avoid conflicts with `main.css`.

Not included: big chart (code available here: https://github.com/KeitIG/elementary-stats)


lewisgoddard:
I do not think the buttons below "Contribute by doing these things" are readable. Those colors on mid-grey do not have a whole lot of contrast.

danrabbit:
* Can we change the link in the nav so it goes to this new page ;p
* The link under "Web Development" should go to this github repo
* "section > .row .column" is too thin on mobile. It would make more sense, I think, to set "max-width: 576" since a row cannot be larger than 1200px - 24px padding / 2 for 50%
* For the flat buttons on this page, we should add a selector ".button.flat"
* Code for the stats piece is here: https://github.com/KeitIG/elementary-stats

danrabbit:
* Open links in new tab/window. At least a couple of the sections have multiple links. I think it would be good to be able to open multiple pages that you're interested in
* I don't think we're currently using MySQL on the beta site so we should drop that from the web section
* Can we make the Google+ community link a "read-more" class and remove the "."?
* Let's make sure all the button labels are Title Case. For example "View Our Translations Page" and "See Our Open Bug Reports"
* According to Harvey, the quotes should use Satisfy: http://www.google.com/fonts/specimen/Satisfy

danrabbit:
* The web browser in the web section doesn't fit on mobile. Maybe we can cut it in half and just repeat the middle so it stretches?
* The web section button should be changed to something like "Fork Us on GitHub"

danrabbit:
* "elementary is created..." should be "elementary OS is created".
* Can we convert the quotes to inline tweets? So the tweet would be like:
“A different language is a different vision of life.” —Federico Fellini elementaryos.org/get-involved

danrabbit:
-[ ] Hmm it looks like we need to change the color for "inline-tweet" to "inherit"
-[ ] "section .heading h1" is a little big on mobile. Maybe we should take it down to like 24px there. Or maybe not and it's fine. I guess second opinion on that one?
-[ ]  on mobile, can we either center align the buttons or have them fill the width?

danrabbit:
- [ ] Can we add a link on the design section to this with a label like "Read About Our Workflow"? http://blog.elementaryos.org/post/107662321291/so-you-fancy-yourself-a-designer

It should probably be converted into a legit part of the docs at some point, cleaned up, expanded upon etc. But I think it's good reference material.

danrabbit:
In Slack we decided to pull out the chart for now so we can land this sexiness right away. We'll bring the chart back in another PR :)

danrabbit:
It looks like there's just regular left and right margins. So maybe we say at like 414px (That's iPhone 6 Plus size) we reduce it to 24px?

emersion:
Yes, seems good to me!

danrabbit:
Do we need font-family: "Open Sans"; in get-involved.css? I think that should already be taken care of in main.css

Looks like we need to add alt text for the images. We can probably keep it simple like "World map" "Application wire frame" "Scratch text editor"

danrabbit:
- [ ] As discussed in slack, we want to change the color of the hero row action buttons up top to #c8d0d0 just to make sure they're entirely legible.

aroman:
This is looking really, really good. Huge props.

One suggestion: maybe make the quote authors into links. Dieter Rams should go to his Wikipedia page?

And a few grammar/wording things I've noticed:

- [ ] "Read the HIG" -> "Read the Interface Guidelines". I know we all call it the HIG because that's what it is, but a prospective contributor might not know what HIG means and why they should click it (we don't define the acronym anywhere on the page).

- [ ] In the Desktop Development section, change "We're always looking for contributors of all skill." to "... of all skill _levels_."

emersion:
@aroman: maybe link to the quote authors Twitter accounts?

Issue #225, opened 2015-01-27
Mobile icons


cassidyjames:
We should provide icons for when people add the site to the homescreen on mobile. Specifically iOS and Android.

cassidyjames:
Reopening for Android icon

PR #226, opened 2015-01-27
Add manifest for Android homescreen title; fixes #102

### LINKED ISSUE #102: Set smarter mobile web app title
I don't think this is a good page title. Especially if we consider that this site will eventually host documentation and other links and not just the Freya download.

Mainly, where I'm experiencing this problem is with bookmarking this site. Especially using the "add to homescreen" function on my mobile. Having a homescreen icon with "elementary OS Freya" is weird.

I think simply "elementary" would be the most encompassing

cassidyjames:
per https://developer.chrome.com/multidevice/android/installtohomescreen


danrabbit:
I think this should go in here? I'm not 100% sure it's right. Lemme know if it looks odd.

cassidyjames:
@danrabbit gonna check here shortly. :)

danrabbit:
Made some alterations based on our discussion and here's sizes

danrabbit:
Let's just scaled the 192 one because the material design guidelines are BS and the phone is grabbing the wrong size

cassidyjames:
Ha aight, updating...

cassidyjames:
@danrabbit looks aight to me.

cassidyjames:
This branch will also fix #225

PR #232, opened 2015-01-27
Promote bounties on the "get involved" page

carbn:
I think it would be nice to promote bounties in the "get involved" page. This branch implements the basic idea, but the graphic should probably be re-done in vector format.


danrabbit:
I do like the concept. I agree it needs a different image :p Maybe we can get @harveycabaguio to mock us up something.

I'm not sure that quote directly relates here. Maybe something like Marsha Sinetar's "Do what you love and the money will follow."?

There should be line breaks after the paragraph titles and they should be statements instead of questions.

I think it's worth taking a couple sentences to explain just what a bounty is."speed up development" and "make developers happy" makes it sound like funding is something extra instead of something essential. Our message should be that bounties create entry-level open source jobs, especially for students. 

It might be worth considering changing the section title to a term more people would be familiar with like "Funding" or "Sponsorship"

danrabbit:
Also, the footer color should probably be changed so it matches that last "light" section

lewisgoddard:
I would love to see bounties included more in our materials, but this needs rebasing before I can pull it into an internal branch for testing.

carbn:
Rebased with the new master.

danrabbit:
Closing this since we fixed it in another branch

Issue #241, opened 2015-01-28
Get javascript variables out of the global scope


eddiemoore:
Very bad to have all the javascript variable in the global scope because it means that users can just open the dev tools and start changing variables which can screw things up.

lewisgoddard:
I think originally some javascript was being called on the page, but it has since all been moved into the separate files.

PR #242, opened 2015-01-28
Getting JS out of global scope. #241

eddiemoore:


lewisgoddard:
This needs to be re-done by the looks of things, probably because so much has happened since it was proposed. I should be available over the weekend to review.

PR #247, opened 2015-01-28
Adds quotes authors links

### LINKED ISSUE #208: Link quote authors on get involved
either wikipedia or maybe if there exists novelty twitter accounts?

emersion:
Fixes #208


lewisgoddard:
I'll merge, but I'm not sure if these are the links we want. We can always change individuals later if a more appropriate one becomes available I suppose.

PR #248, opened 2015-01-28
Adds chart in "Get Involved"

### LINKED ISSUE #206: Add chart to "Get Involved" page
Chart mockup: https://github.com/elementary/mvp/blob/2fd27b3a1b7ad1daa3aa5408c56f5d633f504787/images/get-involved/stats.svg

Source code: https://github.com/KeitIG/elementary-stats

Notes:
* ~~#196 needs to be merged in master before~~
* We should save data as JSON instead of using a CSV parsing library
* Maybe switch from Google Charts to another library? (something like http://bonsaijs.org/ ? Being Google-dependent seems not to be a good thing, and Google Charts seems like a quite heavy lib)
* Clean up `update.php`, remove ugly things like `exec("grep ...");` :-P
* Use Launchpad HTTP API instead of parsing raw HTML (https://launchpad.net/+apidoc/beta.html#milestone-searchTasks)

emersion:
Fixes #206.

Uses https://github.com/nnnick/Chart.js and https://github.com/Regaddi/Chart.StackedBar.js


danrabbit:
For the "fixed" label, to provide good contrast we should be closer to something real bright like  #95A5AC

Can we remove the large h1? I feel like it takes up a lot of space and doesn't add a terrible amount of value.

PR #262, opened 2015-01-28
merge styles from get involved into main

### LINKED ISSUE #255: Merge text styles from get-involved into main.css
We have some little variances here. We should bring this into alignment so we're using the same hn elements for the same sizes and weights.

<bountysource-plugin>

---
Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/8112219-merge-text-styles-from-get-involved-into-main-css?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
This branch tries to reduce the size of get-involved.css by merging styles into main.css and removing obsolete and redundant styles. It also tries to fix some code-style blunders like spaces instead of tabs, new lines after selectors, etc.

Watch out for regressions in the alignment/resizing of columns and their contents on desktop and mobile. Also check to make sure regressions weren't introduced to header styles on the homepage and installation page.

fixes #255

### DIFF
--- a/get-involved.php
+++ b/get-involved.php
@@ -7,7 +7,7 @@
     include '_templates/header.php';
 ?>
             <section class="row hero">
-                <h2>Check out our progress for Freya Beta 2</h2>
+                <h1>Check out our progress for Freya Beta 2</h1>
 
                 <div class="charts">
                     <div class="barchart-ctn">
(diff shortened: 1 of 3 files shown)


lewisgoddard:
There seems to be some serious changes in  header styles on get involved. Were these intentional?

danrabbit:
Yes. I think if we feel like we need to change the header styles that should happen in main.css so we have consistent typography across the site. This branch should more or less keep the same sizes, but it most likely introduces changes in weight and color

PR #273, opened 2015-01-29
Added a simple Square store page

maffity:
@cassidyjames helped me set up git and create a branch. :smile:


maffity:
Fixes #274

danrabbit:
- [ ] That's pretty legit. I wonder if we should remove the border?
- [ ] I feel like this needs a link in header.php
- [ ] Clicking checkout takes me to square's website. Is there any way to do this in a new page so we don't navigate away with no way to get back?

maffity:
The border is added automatically by Square, but we might be able to style it with CSS.

maffity:
Oh, it was a setting in the embed code. Removed!

Issue #263, opened 2015-01-28
Port the Developer landing page over [$25]


danrabbit:
Unless Harvey has a better design he's working on, we should port the Developer landing over.

PR #281, opened 2015-01-29
Create developer page (#263)

KeitIG:
First try to port http://elementaryos.org/developer


danrabbit:
- [ ] We'll need to link this page in the header template. Probably next to "Get Involved"?
- [ ] Can we set the background color/image/etc at html, body in developer.css? That way we'll get the "transparent" navbar. CSS changes in master should fix the footer color if you'd like to merge master in ;)
- [ ] Can we set the links to the same as their currently set on the main/old site? (We can just reference main/old site for now for the sake of making things work).
- [ ] The left dashed border looks a little funky on mobile. Can we add a media query that only sets that left border width when the links are actually to the right? (Looks like that happens at about 1200px)

Otherwise, great port!

danrabbit:
The styles in your media query are backwards ;p

danrabbit:
- [ ] there should be a line break between html, body in your css.
- [ ] Since you're only using 100, probably don't need to load other weights of Raleway

Issue #136, opened 2015-01-23
Real boot instructions for Mac [$15]


danrabbit:
I dunno who wrote the nonsense that's currently there but it's just as easy to boot on a Mac as a PC. Pop in your USB, hold down option during boot, and select your boot disk. nbd.

caketable:
I wrote a copy for Booting elementary OS on Macs during the (failed) documentation sprint a while back, but it got rejected. Here's the document. I tried to make it close to the Luna Boot Instructions.   https://docs.google.com/document/d/1iILdxwABYV3FqjFKrxERRU-qSsMJU8oDmULv-0uIUSk/edit?usp=sharing .

PR #286, opened 2015-01-29
Adds boot instructions for Mac

emersion:
See #136


danrabbit:
I don't think most Apple keyboards have the option symbol printed on them anymore. See http://www.apple.com/keyboard/ . It might be better to just say " After you here the chime, press and hold <kbd>option</kbd>..."

It might also be worth noting that the bootloader may incorrectly identify the disk as "Windows" and that this is normal

PR #287, opened 2015-01-29
Allow user to zoom in on touchscreen devices.

### LINKED ISSUE #280: User should be able to zoom in
You should not forbid users from zooming in. Just set the initial scale to 1.0 and let the user zoom in. It's generally a bad practice and should not be done.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/8125326-user-should-be-able-to-zoom-in?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github).
</bountysource-plugin>

iparr:
As per suggestion from H5BP project and recommentation from @cybre

fixes #280

### DIFF
--- a/_templates/header.php
+++ b/_templates/header.php
@@ -3,7 +3,7 @@
     <head>
 
         <meta charset="UTF-8">
-        <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1, user-scalable=0">
+        <meta name="viewport" content="width=device-width, initial-scale=1">
 
         <meta name="description" content="<?php echo !empty($page['description']) ? $page['description'] : $sitewide['description']; ?>">
         <meta name="author" content="<?php echo !empty($page['author']) ? $page['author'] : $sitewide['author']; ?>">


cassidyjames:
@iparr a requirement of merging this is setting input text size to 16px, as mobile Safari will automatically zoom in on inputs otherwise.

iparr:
I hear your @cassidyjames 

Do you think implementing something like this would be a solution?:
https://gist.github.com/zachleat/2008932

I know it feels a bit messy, but I believe it'd be an improvement.

cassidyjames:
@iparr I think I'd rather just set the input size than use that code. I get it (and it'd likely work), but 16px text looks just fine. And actually, removing the text size from the input should make it inherit the default of 16px.

lewisgoddard:
This issue has already been closed via pull request #473

PR #288, opened 2015-01-29
l18n

### LINKED ISSUE #37: Translate the website [$50]
We should probably get this set up sooner than later

<bountysource-plugin>

---
There is a **[<total-bounty>$50</total-bounty> open bounty](https://www.bountysource.com/issues/7626065-translate-the-website?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github).
</bountysource-plugin>

emersion:
Fixes #37 without deeply modifying HTML pages.

Result: 
* Auto-detect user language: http://beta.elementaryos.org/branch/i18n-html/
* In French: http://beta.elementaryos.org/branch/i18n-html/?lang=fr

This solution buffers all HTML and translates everything that can be translated. It works a little bit like Javascript solutions, but it is server-side. [Transifex](http://transifex.com) is used for collaborative online translations.

Translation guide: https://github.com/elementary/mvp/blob/i18n-html/TRANSLATE.md


danrabbit:
How would translators submit translations with this system? I think we need a solution that will work with a user-friendly translation tool (like Transifex or Loco)

emersion:
Ow, as I'm able to replace untranslated strings by translated ones, I'll be able to produce empty translation files for translators. ;-)

On the one hand, we'll be able to continue to write code without thinking about translations, on the other hand it's a little bit more complicated than a simple `<?php echo _("my string"); ?>`. But I think it will be quicker to use this method than to convert all existing strings.

emersion:
Just added strings extraction in backend. For instance, to extract strings from the index page: http://beta.elementaryos.org/branch/i18n-html/backend/extract-i18n.php?page=index

This can be useful to update automatically translation files, services like Transifex provide a "Automatic update URL" for each file. This outputs JSON because it was the simplest, but I can output any format (like `gettext`, `.properties` and so on).

KeitIG:
If I understand well, the page is loaded server-side, then the server review all the code and replace translatable strings.

If the concept is good, I'm not sure It would be really efficient. I would prefer a real translation rather than re-parse the whole page sent to the client.

emersion:
> the page is loaded server-side

The page is always loaded server-side. ;-)

Here, I'm just reading the response to search for translatable strings. The complexity of this algorithm is linear, so it should not impact too much on performance.

I just performed some benchmarks:
```
Running 100 tests.
Time without i18n: 0.09564995765686
Time with i18n: 0.17601299285889
Factor: 1.8401784712788
```

So it is approximatively 1.8 times slower. I think this is acceptable, what do you think @KeitIG @danrabbit @lewisgoddard ?

lewisgoddard:
I don't think speed is the main issue here, I think it's more about potential contribution issues. This isn't terribly easy to add new languages and translations to. Plus, I foresee encoding issues.

emersion:
Just added JSON i18n files support and automatic user language detection. 

* Autodetection: http://beta.elementaryos.org/branch/i18n-html/
* French: http://beta.elementaryos.org/branch/i18n-html/?lang=fr

Only the index page is translated ATM.

And here is a little Transifex test page: https://www.transifex.com/projects/p/mvp-testing/

emersion:
I translated `index`, `get-involved` and `developer` pages in French, and it's working pretty well. See my first comment for more info about the workflow.

I updated the router to support `/<lang>/<page>` URL scheme too.

emersion:
All right, I've finished now. We just need to update the server config (@fabianthoma) and we'll be able to merge!

Issue #244, opened 2015-01-28
404 page not working


eddiemoore:
Currently if i go to **/asdasd** it just says ```File not found.```
Looks like the router.php file will need to updated for this.

fabianthoma:
Actually this is the fault of nginx, which does routing on the server, the router.php is just for local testing on developers machines, I'll fix it next time I log on to the server

lewisgoddard:
We don't have a "nice" not found error page at all. This needs to be designed, added, and then configured in both Nginx _and_ the router file.

cassidyjames:
Umm guys? We have http://beta.elementaryos.org/404. It just needs to be set up server-side.

lewisgoddard:
"Go to Home Page" ?

That is the most un-elementary phrasing I have ever seen.

Also, when the hell was that added!?

cassidyjames:
Alright, developer-side is taken care of with the router.php. Just waiting for @fabianthoma to update the nginx config server-side and we'll be all set.

fabianthoma:
Done, 404 now works

PR #289, opened 2015-01-29
Adds 404 page support in router.php

emersion:
See #244


cassidyjames:
I like the idea, but it doesn't work if you try to go directly to a non-existent *.php page. Probably okay, as it just emulates the server config, but it probably wouldn't be too hard to add. ;)

emersion:
Fixed!

cassidyjames:
Works great. :smile: :shipit:

PR #297, opened 2015-01-29
Add android chrome favicon

RoryJ-:
Reference: http://realfavicongenerator.net/blog/android-chrome-and-its-favicon/


cassidyjames:
@RoryJ- from what I'm seeing, Chrome on Android uses the same icon as the add-to-homescreen icon that we already defined in the manifest. I'm okay with explicitly tagging it in the head, but I don't think we need another image besides it.

Issue #298, opened 2015-01-29
Set up subdomains for "developer", "developers", "dev", and "develop"


cassidyjames:
Right now our language isn't super clear and we use a mix of "Developer" and "Develop". "Developer" doesn't really make sense as a page title/URL to me as much as either "Develop" or "Developers".

I vote we change the link wording to "Develop" and the URL to `/develop`, and leave the heading that's on the page.

danrabbit:
Developer.foo.bar is pretty much defacto. Check basically any website. Apple, Facebook, Paypal, Microsoft (it's a redirect, but still)

cassidyjames:
Eh, Facebook and Google both redirect to developers and Microsoft redirects all three (develop, developer, and developers) to their actual site. I just feel like "Developer" is weird because it's the only thing that references who the user is instead of what they're doing or where they're going.

cassidyjames:
I guess another reason "Develop" felt right in the right nav is that the left nav is nouns and the other item in the right nav is a verb.

aroman:
Developer is definitely the standard from what I can see.

Like Dan said, everybody does it.

Check these search results out (esp. page 2): https://www.google.com/search?client=opera&q=develop+or+developer&sourceid=opera&ie=UTF-8&oe=UTF-8

My vote is for developer.elementaryos.org.

aroman:
Another good argument in favor of developer vs. develop is that "develop" seems more like "develop elementary OS itself", rather than "third party developer(s)"

cassidyjames:
Hang on Avi, we're not done bikeshedding this! Hop into #web on Slack :wink:

cassidyjames:
Okay, I concede. I have opened pandora's box. I vote that we make the URL whatever @danrabbit wants to do.

danrabbit:
Okay so it sounds like in Slack we decided to:

- [ ] Make the URL "developer"
- [ ] Set up redirect URLs for "developers" and "develop"
- [ ] Set up subdomains for "developer", "developers", and "develop"
- [ ] Call it whatever Cassidy wants in the UI as long as that name is clearly separated from 1st party developers. ie someone should not be confused thinking that to hack on elementary itself they should visit the "developer" page.

danrabbit:
mentioning @fabianthoma for subdomains

danrabbit:
Changing the description to just the subdomain bit.

Let's open separate reports for any other changes we need.

fabianthoma:
[...]

PR #301, opened 2015-01-29
Redirect dev pages

cassidyjames:
As per #298


cassidyjames:
@danrabbit remember to delete these branches after merging. :)

PR #303, opened 2015-01-30
Title vars

### LINKED ISSUE #302: Include "elementary" in title variables
I like the idea of not forcing all page titles to end with ` | elementary OS`, but we should probably include some natural reference to elementary in the page titles. Both for SEO and for clarity if someone sees just a page title (in their browser, on social media, etc.)

For example: **Developer** becomes **Develop apps for elementary OS**, **Store** becomes **elementary Store**, **Get Involved** becomes **Get Involved with elementary**, etc.

<bountysource-plugin>

---
Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/8143533-include-elementary-in-title-variables?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github).
</bountysource-plugin>

cassidyjames:
Add references to "elementary" in page titles to fix #302


danrabbit:
I think maybe "elementary" should always go at the end. If the tabs are not very wide you could end up with several tabs that just say "elementary" instead of "code of conduct" or other useful titles.

cassidyjames:
Alright, doing `Title &sdot; elementary` when it doesn't naturally come at the end, as we discussed in Slack.

cassidyjames:
Updated @danrabbit

PR #304, opened 2015-01-30
Animated hero image

devTristan:
A few things to note:

- This is not perfect. The image (images/hero-poster.png) and the video (currently hosted by a third party) have the wrong aspect ratio. It must be 1.6:1, since that is the aspect ratio of the notebook's screen. This results in overlap as seen in the screenshot below.
- The video is about 24 megabytes. It won't affect the page load, and shouldn't be perceptible to the user (even if it never loads) but methods of getting smaller file sizes should be explored.
- While the video is loading, or if the video never loads due to browser incompatibility, hero-poster.png is shown on the screen.
- If hero-poster.png does not load, the screen is black.
- This scales with screen size, works on phones.
- The image and video are rendered slightly larger than the frame they fit into. This is in order to prevent edge bleed issues.
- When this is set up properly, the effect is really cool.

Thanks @Tajnymag for the video!


devTristan:
Ideas to make the video load quickly: Make it short. Avoid actions that will hamper vp8 compression, such as animated background changes. Do simple things like opening noise, closing noise, opening midori, closing midori. Wait several seconds between actions, so that it doesn't feel rushed or look distracting. Don't move windows. Don't move the mouse unless necessary, don't type.

cassidyjames:
@devTristan the general idea is nice, but before we think about merging we'll want to 1. get the right aspect ratio (whether that's recording on the right resolution or resizing our laptop frame is up for discussion), and 2. record a cleaner video.

I had actually already started working on this locally this evening without seeing your branch. :stuck_out_tongue:

devTristan:
I totally agree with all of that. Thinking of this PR as a good start.

16:9 is the most common laptop aspect ratio, so I'm inclined to suggest that. On the other hand, the aesthetics of the notebook itself are important, so if 16:10 looks better it might be worth the hassle of getting video at that ratio.

lewisgoddard:
I'm going to go ahead and close this. Since the last commit, we've massively changed how we deliver the image across sizes, as well as the model used in photos.

PR #306, opened 2015-01-30
remove unnecessary styles from developer.css

danrabbit:
This branch basically just tries to make developer.css smaller

Also fixes an error where focus styles were disabled for some reason.


cassidyjames:
@danrabbit hm, the footer link is blue again. Also, the `:focus` styles are blue for header and footer links.

PR #311, opened 2015-01-30
make fonts awesomer

### LINKED ISSUE #291: Heading sizes are tiny
Right now main.css has the following:

```
h1 {
    font-size: 24px;
    opacity: 0.8;
    font-weight: 600;
}

h2 {
    font-size: 18px;
    font-weight: 300;
}

h3 {
    font-size: 11px;
}
```

These sizes are pretty crazy. We should lay out more sensible sizes for h1 through h6. We should also probably use rems as the unit instead of pixels so it is relative to the base text size.

<bountysource-plugin>

---
Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/8136456-heading-sizes-are-tiny?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
Fixes #291 

This branch makes fonts awesome by moving styles from developer into main and setting up h1-h5 then using those on various pages.

### DIFF
--- a/developer.php
+++ b/developer.php
@@ -13,25 +13,25 @@
                 <div class="column half developer-sections">
                     <div id="dev-section-1" class="dev-section">
                         <a href="http://elementaryos.org/docs/human-interface-guidelines" title="Design" target="_blank">
-                            <h3>Design</h3>
+                            <h5>Design</h5>
                             <p>Read the Human Interface Guidelines (HIG) and learn how to make your app a first-class experience.</p>
                         </a>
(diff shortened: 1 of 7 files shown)


cassidyjames:
Looking good.

PR #312, opened 2015-01-30
headings for installation

danrabbit:
forgot to fix installation and 404 too


cassidyjames:
Whoops :D Those pages look better now.

Issue #310, opened 2015-01-30
No focus style for home/logo


cassidyjames:
The home link (our "e" logo) has no focus style.

PR #313, opened 2015-01-30
Inline SVG and set its style with CSS

cassidyjames:
Works pretty dang well. Just remember, any time you're setting a color, set the fill to the same thing and the logomark will inherit it.


cassidyjames:
Fixes #310

PR #316, opened 2015-01-30
Add high-res png favicon

RoryJ-:
On android chrome, the apple-touch icon is preferred over favicon.ico,
so this png favicon is needed.


lewisgoddard:
I'm not entirely clear on what this is for. Do we not want touchscreens to use the apple-touch icon? Why do we need a 256 icon anyway, favicon.ico is already 128? Where's the documentation on this Chrome behaviour?

RoryJ-:
So for whatever reason, android chrome prefers the apple-touch icon over the favicon - so in the multitasking view, the apple-touch icon is shown. We do not want this because the apple-touch icon has a blue background and is designed for iOS homescreen. To solve this, we link a high-res png file that chrome prefers over the apple-touch icon. The launcher-icon/4x.png will still be show on the home screen. It's 256x256 because of hi-dpi screens.

KeitIG:
as .ico is a bit more "democratized", Lewis decided to not use a .png.

Otherwise, I confirm the bug. 

screenshot_2015-01-30-13-48-49

RoryJ-:
I agree, although IE10 is the only semi-modern browser that doesn't support it. With this, if the browser doesn't support png, it will fallback to the ico.

lewisgoddard:
@KeitIG Can you confirm this fixes the bug? http://beta.elementaryos.org/branch/png-favicon/

KeitIG:
@lewisgoddard This does not fix anything, because in c83855a6a325a1ae10c5bfd6cab0683b2d74e843, line 28, images/favicon.png does not exists :)

lewisgoddard:
@KeitIG `images/favicon.png` is included in that commit, as far as I can see.

KeitIG:
Oops, 

even with a cache cleared, it was not working this morning, anyway, it's good now.

screenshot_2015-02-02-15-29-56

PR #333, opened 2015-02-01
Fixes 404 page with URLs like /does/not/exist

emersion:
It was trying to load something like `/does/not/exist/styles/main.css`.


lewisgoddard:
This is going to break a lot of our `/branch/` pages I should think. They'll be forever loading the root stylesheet instead of their own.

emersion:
Yes, you're right!

PR #335, opened 2015-02-01
Solved svg rendering on firefox mobile in get involbed page

choyan:
Refers to the issue https://github.com/elementary/mvp/issues/323

### DIFF
--- a/get-involved.php
+++ b/get-involved.php
@@ -68,7 +68,7 @@
                         </div>
                     </div>
                     <div class="column half">
-                        <img src="images/get-involved/translations.svg" alt="World map">
+                        <img src="images/get-involved/translations.svg" class="svg" alt="World map">
                     </div>
                 </div>
             </section>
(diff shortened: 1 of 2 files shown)


emersion:
I don't think adding a `.svg` class is the right solution to solve this. Instead, adding this works for me:
```css
.column.half {
    max-width: 100%;
}
```

lewisgoddard:
Shouldn't this be fixed by `img { max-width: 100%; }` ?

emersion:
It should, but it doesn't work because of this:
```css
.column {
    display: inline-block;
}
```

lewisgoddard:
Okay, so is  `img { max-width: 100%; }` already in the CSS then ?

emersion:
Yes ;-)

https://github.com/elementary/mvp/blob/master/styles/main.css#L118

choyan:
I thought its a firefox only problem. So, i didn't think working with the column width is a good idea. :)

danrabbit:
I can't see a reason why a column should ever be more than 100% width, so I think it's probably safe to fix it there. Would be the smallest diff.

PR #337, opened 2015-02-02
Add browsers icons to privacy-policy (fix #307)

### LINKED ISSUE #307: Add browser icons in privacy policy [$10]
It'd be a nice touch to have browsers listed horizontally at the bottom there with their icons.

<bountysource-plugin>

---
There is a **[<total-bounty>$10</total-bounty> open bounty](https://www.bountysource.com/issues/8147699-add-browser-icons-in-privacy-policy?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github).
</bountysource-plugin>

KeitIG:
Branch not finished, waiting for your comments for responsive.

Otherwise, I think the icons-as-bullets version was better, display browsers icons in columns makes (imo) cookies section the only usefull section of the page, and that's not.


danrabbit:
Hm I'm not sure we should be adding a non-reusable class here. What's wrong with using row? It seems like it'd be better to do that and then use img.browsers-list so we're only adding one new class that is indeed a class.

But what would be even better would be if we can come up with a more generic selector for app icons above columns since I think we'll probably want to do that again and again

KeitIG:
for your last comment @danrabbit, I think we did not make it right.

A nice way would have been to do a .icon class, something like:

```css
.icon {
/* generic style for all icons */
}
...


.icon-ie:before {
/* do something here to display IE icon */
}
```

and then use (a bit like font-awesome)
```html
<span class="icon icon-ie">Internet Explorer</span>
```

Tell me if I have to do this, and/or if I should do this in another branch

danrabbit:
Hey Pierre, this looks pretty solid to me. I think we can probably make something more generic at a later time. But I'm good to merge this :)

Issue #339, opened 2015-02-02
Payment not remembered until reloading the page


cassidyjames:
If I:

1. I pay for the download using Stripe
2. Click a button to download the ISO
3. Go back to the elementary tab in my browser
4. Click to download again

...I expect to get the download dialog again. Instead, I'm prompted to pay again. If I reload the page then click the download button again, I'm properly remembered.

lewisgoddard:
I will be able to grab this on Wednesday.

lewisgoddard:
Done by devTristan.

PR #343, opened 2015-02-03
Remove amount selection after payment, fixing #339

devTristan:
This should fix it. Could use a review. Hope the style is okay.

All I did in index.php was add `<div class="amounts"> ... </div>` plus an indentation, hence the unnecessarily long diff.

I'm fairly sure it'll work in theory, but I don't have a spare stripe API key sitting around.

Relevant: Would it be possible/safe to put a stripe test key in this repo?

### DIFF
--- a/index.php
+++ b/index.php
@@ -15,25 +15,27 @@
             </div>
             <img class="hero" src="images/notebook.png">
             <div class="row">
-                <?php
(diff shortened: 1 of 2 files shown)


lewisgoddard:
This needs master merged or rebased before I can test it.

devTristan:
you got it, looking at this now

devTristan:
Okay, that should do it, I think.

devTristan:
Thanks @lewisgoddard :)

PR #355, opened 2015-02-06
Add Midori browser icon in privacy-policy (fix #353)

### LINKED ISSUE #353: Add Midori to privacy policy
So, after https://github.com/elementary/mvp/pull/337 got merged (and fixed #307 issue), I think that it is a perfect time to discuss adding Midori to that list. I think that this is an important issue because Midori _is_ our default browser. Not mentioning it seems like we're not supporting our own product.

KeitIG:


lewisgoddard:
I think we're better off linking to the official Midori website, rather than the Wiki that runs the backend.

KeitIG:
Done

PR #360, opened 2015-02-07
Re template

lewisgoddard:


emersion:
I agree with this approach, but there is still sitewide hardcoded, even if I don't think we can find another solution.

lewisgoddard:
The idea is you don't move sitewide, it's always there, providing any sitewide variables. But if you want to move anything else, this should make it easier. Mostly, it sets a standard for it we want to move the navigation out of the header.

As you said, I don't think we can dynamically load sitewide easily.

Issue #334, opened 2015-02-01
put a wikipedia link next to the social icons


ferst9:
New users might be curious about elementary's history.
There should be a wikipedia link about elementary next to the social icons (reddit,...).
BTW the german version (http://de.wikipedia.org/wiki/Elementary_OS) is more complete than the english one.
PS : It would be great to have detailed information about elementary on wikipedia. A lot of information is missing.

danrabbit:
There doesn't appear to be a FontAwesome icon for Wikipedia, so that would mean we'd need our own SVG there.

cassidyjames:
I've commented on the FontAwesome project. A ton of people have requested the icon, so I'd imagine it'll be added in a future release. I'm okay with adding our own SVG until then.

ferst9:
Thanks guys, now the wikipedia article should include much more details about elementary. How it all began until today.

aleksandar-todorovic:
Maybe we don't need a separate SVG until then? After all, Wikipedia logo is nothing more than a capital W. Here is a [quick fix](http://jsfiddle.net/fLktadyn/1/) as suggested in the Font Awesome issue mentioned above.

danrabbit:
That doesn't look like the wikipedia "W" here. See the actual logo: http://en.wikipedia.org/wiki/Logo_of_Wikipedia

snwh:
I question the need to link to wikipedia. Shouldn't our own site explain the project?

ferst9:
Maybe the website should only contain the essential part of elementary. The whole history can be found on wikipedia.

danrabbit:
I'm going to close this as our wikipedia page contains inaccurate information and I don't really trust linking to an external source like this.

PR #361, opened 2015-02-08
Added link to Wikipedia page #334

nblackburn:
See #334

### DIFF
--- a/_templates/footer.php
+++ b/_templates/footer.php
@@ -6,6 +6,7 @@
                 <li><a href="http://reddit.com/r/elementaryos" target="_blank" title="Reddit"><i class="fa fa-reddit"></i></a></li>
                 <li><a href="https://plus.google.com/114635553671833442612" target="_blank" title="Google+"><i class="fa fa-google-plus"></i></a></li>
                 <li><a href="https://www.facebook.com/elementaryos" target="_blank" title="Facebook"><i class="fa fa-facebook"></i></a></li>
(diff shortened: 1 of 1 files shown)


cassidyjames:
Oh the book icon is an interesting idea. In case anyone is wondering, here's what it looks like in context.

screenshot from 2015-02-07 20 29 58

danrabbit:
Hrm I dunno about the book. At first glance it looks like "documentation" or something. I think we should still use the wiki "w"

snwh:
I agree with Dan, the logo would be more recognizable. 

However, is there any need to link to the wikipedia page?

fabianthoma:
First I also think the iconic W of Wikipedia should be used, but I have to agree with @snwh here, does one need/should link to wikipedia? It's a place for knowledge, and if anything that knowledge should also be represented on our own website.

nblackburn:
The "W" icon would be more relevant but FontAwesome doesn't contain that icon so made do with the resources available.

KeitIG:
as Cassidy posted here, this icon may be on the road https://github.com/FortAwesome/Font-Awesome/issues/388

shoully:
Yeap, it's on Font-Awesome 5.0.0 milestone.

emersion:
See #334.

nblackburn:
The [Wikipedia](http://fontawesome.io/icon/wikipedia-w) icon has arrived.

PR #378, opened 2015-02-10
"Change language" button

### LINKED ISSUE #347: "Change language" button
Add a "Change language" button, and maybe a little message "This page is available in your language" when accessing `/de/installation` from a Spanish browser?

emersion:
Fixes #347.


KeitIG:
Click on developer and -> http://beta.elementaryos.org/branch/change-lang/fr/developer

danrabbit:
On the popover can you add "background-clip: content-box;" and also change the border color to rgba(0,0,0,0.2); This allows the border to blend more nicely on dark backgrounds like the get-involved page :)

It doesn't seem to work on mobile. Tapping the link scrolls you to the top of the page. The popover also doesn't re-center into the page.

cassidyjames:
In Firefox, I can't click any links in the popover. Moving away from the "Language" link hides the popover before I get the mouse in it.

danrabbit:
I agree with Cassidy, it's probably better to use a click (both to show and dismiss) than a hover

lewisgoddard:
How are these languages organised? They aren't in alphabetical.

And why are they all in English? Shouldn't language names be presented in their own language?

emersion:
@lewisgoddard They are alphabetical, but English is kept at the top of the list. Better to order it with others?

lewisgoddard:
It doesn't look very separate, maybe throw in a `hr` to show it's not in order.

emersion:
- [ ] max-height on popover
- [ ] Cookie to save user language
- [ ] http://fvsch.com/code/transition-fade/test5.html#test3

emersion:
All right, fixed all ;-)

lewisgoddard:
I get a style that sets `.popover-content { left: 0.00019...px` It means the popover isn't centered over the languages button.

emersion:
@lewisgoddard Which browser? Tested in Chromium and Firefox

lewisgoddard:
Google Chrome on elementary OS Luna.

emersion:
Hum, which version? And, is the popover not centered _at all_?

lewisgoddard:
Chrome 40

`left: 0.001953125px;`

screenshot from 2015-02-11 18 03 42

emersion:
Strange, I've the same version. I dunno, maybe try to clear the cache?

lewisgoddard:
Still does it. It's that small left parameter. The `left: -70px` works perfectly. I think it's how the javascript is figuring out the offset. Maybe _you_ should hard-refresh?

lewisgoddard:
Okay, so looking at your current JavaScript, it just shouldn't work with box models. The current calculation doesn't give anything reliable. The link offset is changed by loading, which is then figured out again on a second click, and the whole thing returns a very small number for me. These are my proposed solutions.

```
var $popover = $('.popover');
var $content = $('.popover-content');
$popover.addClass('active');

// with Fixed Positioning
var popoverPos = ( $popover.offset().left + ( $popover.outerWidth() / 2 ) ) - ( $content.outerWidth() / 2 );
$content.css({ left: popoverPos });

// with Absolute Positioning (Current)
var popoverPos = ( $popover.outerWidth() / 2 ) - ( $content.outerWidth() / 2 );
$content.css({ left: popoverPos });
```

emersion:
Better now?

lewisgoddard:
Absolutely.

PR #384, opened 2015-02-12
Fixes some bugs with translations

emersion:
Disables translations on localized languages names, fixes a bug when `<` and `>` are present in script tags.


lewisgoddard:
Wouldn't it be cleaner to put `$l10nDisabled = false;` at the beginning of the while loop, so it isn't set twice?

emersion:
It is at the beginning, too. What we need here is to reset it each time we find a new tag, otherwise all tags will be disabled after one data-l10n-off attribute.

lewisgoddard:
I mean you can set it off by default for each loop rather than setting it in two places.

Old code, set once before beginning, and at the end of each loop.
```
$l10nDisabled = false;
    while ($i < strlen($input)) {
        ...
        if ($name == 'data-l10n-off') { // Disable translation for this tag
            $l10nDisabled= true;
        }
        ...
        $l10nDisabled = false;
    }
```

New code, set once at the beginning of each loop.
```
    while ($i < strlen($input)) {
        $l10nDisabled = false;
        ...
        if ($name == 'data-l10n-off') { // Disable translation for this tag
            $l10nDisabled= true;
        }
        ...
    }
```

Don't even default, just set conditionally.
```
    while ($i < strlen($input)) {
        ...
        if ($name == 'data-l10n-off') { // Disable translation for this tag
            $l10nDisabled = true;
        } else {
            $l10nDisabled = false;
        }
        ...
    }
```

emersion:
> Don't even default, just set conditionally.
```php
while ($i < strlen($input)) {
    ...
    if ($name == 'data-l10n-off') { // Disable translation for this tag
        $l10nDisabled= true;
    } else {
        $l10nDisabled = false;
    }
    ...
}
```

Won't work: if there is no attribute, `$l10nDisabled` will not be set to false.

> ```php
while ($i < strlen($input)) {
    $l10nDisabled = false;
    ...
    if ($name == 'data-l10n-off') { // Disable translation for this tag
        $l10nDisabled= true;
    }
    ...
}
```

Won't work, will be set to false at the beginning of each loop, so will always be false when  entering [the else statement](https://github.com/elementary/mvp/blob/l10n-off/_templates/l10n.php#L219)

lewisgoddard:
Oh, okay, this is only setting it to off at the end of the tag, which isn't every while loop.

PR #381, opened 2015-02-11
Add receipts emails to stripe query. (fixes #340)

lewisgoddard:
Add receipts emails to stripe query.
Fixes #340.

PR #389, opened 2015-02-13
Some fixes for Stripe receipts [fixes #381]

bradaric:
Not sure it will completely resolve the issue, but this fixes the problem with payment.php and its call to Stripe_Charge::create(). It also adds another setcookie() call as the current one was not enough when testing on localhost.


lewisgoddard:
This isn't currently mergeable, could you rebase it off my branch? I'm curious as to whether this fixes things.

PR #393, opened 2015-02-14
Revert "Update index.json"

lewisgoddard:
Reverts elementary/mvp#392


emersion:
Will revert when pulling new translations from Transifex.

PR #394, opened 2015-02-14
Switch to elementary.io

lewisgoddard:


fabianthoma:
It's actually elementary.io, without OS.

PR #405, opened 2015-02-15
App carousel

danrabbit:
This branch sets up initial work for the app carousel on the home page. known issues:

- [x] not fully responsive. Needs help especially at tablet sizes
- [ ] lame linear animation
- [x] slider code is duplicated from installation.js. We should do something such that it's shared

Preview: http://beta.elementary.io/branch/app-carousel/


JoshStrobl:
Commenting as per recommendation by @danrabbit (*thanks for the quick response*) with a known issue (but figured I'd c&p by issue content here):

"When changing between the photos and music slide, the mountains background "shifts" substantially. I went ahead and [recorded a short clip to show the issue](https://www.youtube.com/watch?v=o7wTyUpWdwk).

Browser (as you can see in the video): `Chrome 40.0.2214.111`
System (not like it really affects anything, but I'll provide it anyways): `Fedora 21`"

emersion:
@JoshStrobl Just fixed the "shift" issue.

danrabbit:
Okay so we agreed that we just want *something* and we can come back and make it nice for mobile later.

PR #410, opened 2015-02-17
Updated Serbian translation

tomicakorac:
As discussed in the bug reported to elementary OS repo, Serbian language uses Cyrillic script, and translations into Serbian must use Serbian Cyrillic script, as per the current Orthography of Serbian language, and constitution and law of Serbia.


emersion:
Please update translations on Transifex, because otherwise they'll be overridden when pulling translations from Transifex. See TRANSLATE.md

You can directly import JSON files in Transifex so you won't have to copy each string.

tomicakorac:
I already did that earlier today. Do you have any idea when can we expect the changes to take effect?

emersion:
Okay, I'll merge that, and anyway it will be overridden with new translations next time we'll pull those from Transifex.

PR #425, opened 2015-02-21
Translation in german

Anducar:


emersion:
Please don't update translations on Github, do it on Transifex. See https://github.com/elementary/mvp/blob/master/TRANSLATE.md

PR #431, opened 2015-02-22
Added colored third party logos on hover

user501254:
Note: 
* more optimizations can be made to the SVGs
* "WIRED" logo shows no apparent change on hover due to the base colors in the logo (black and white).


styles/main.css
```diff
@@ -365,6 +365,16 @@ footer ul li a:focus {
     margin-left: auto;
     margin-right: auto;
     display: block;
+    filter: url("data:image/svg+xml;utf8,<svg version='1.1' xmlns='http://www.w3.org/2000/svg' height='0'><filter id='greyscale'><feColorMatrix type='matrix' values='0.3333 0.3333 0.3333 0 0 0.3333 0.3333 0.3333 0 0 0.3333 0.3333 0.3333 0 0 0 0 0 1 0' /></filter></svg>#greyscale");
+    -webkit-filter: grayscale(100%);
+    filter: grayscale(100%);
+    filter: gray;
+}
+
+.column:hover img {
```
Scribblex: Why don't to use class on img tag? 
Advantages:
* no cascades
* more evident code
* one class for one appereance

user501254: I tried to make as few changes as possible.

danrabbit:
Hrm, I'm not sure this is really better than what we already have. Especially the lifehacker logos becomes very difficult to see

lewisgoddard:
I think this would be much better if you used the existing black and white logos and switched them on hover.

user501254:
@lewisgoddard, I used 2 images. One gray-scale and the other colored which did make the "lifehacker" logo equally visible as the others; also the "WIRED" logo shows changes on hover.

lewisgoddard:
Hm, any way we can fade between the two? We have jQuery anyway, so we should be able to.

user501254:
Yeah, we can make the change more gradual.

user501254:
I gave this a fair amount of try. However I couldn't fade between the two images using JS, therefore I replicated the same using CSS. Should have done this the first time itself.

lewisgoddard:
I'm going to close this because the discussion seems to be that the logos themselves are not the main focus, and adding colors oh hover does make lifehacker hard to read. It also breaks some of the more minor discoverability rules and makes the difference in tone below the recommended accessibility standard (not that we've followed that much anyway).

I'm not sure this is something that we really want implementing, but thank you for your contribution.

PR #445, opened 2015-03-10
Better Mac Documentation

linusbobcat:
To note:
- I tried to keep these instructions as simple and user-friendly as possible. I know that DVD booting is very unsexy, but for pre-2010 Macs it's the only way to boot elementary OS, which is why I added instructions for it.
- For the sake of keeping it simple, I avoided adding instructions on installing rEFInd.
- The "elementary thumbnail creator" is referred to for creating USB drives. This is just a GUI wrapper for dd, which tends to reliable. It uses Automator, meaning that it should run on any older versions of OS X.
- So far, I'm unable to create a new tab for "Creating an Install Drive on OS X"
- The layout is completely broken on desktop.
- This is my first time on Github, so apologies if I'm doing something wrong.


installation.php
```diff
@@ -43,7 +43,8 @@
                     <div class="text-center">
                         <div id="creating-a-usb-choices" class="column linked">
                             <a class="button creating-a-usb-on-windows" href="#creating-a-usb-on-windows">Windows</a>
-                            <a class="button creating-a-usb-on-others" href="#creating-a-usb-on-others">OS X & Ubuntu</a>
+                            <a class="button creating-a-usb-on-osx" href="#creating-a-usb-on-others">OS X</a>
```
emersion: Why does the `href` property is set to `#creating-a-usb-on-others` ?

lewisgoddard: I think emersion means why have you changed the class and not the href for OS X, and why is it still `*-others` for the Ubuntu section.

lewisgoddard:
Additional to the class/href issue documented above, you also need to take a look at `/scripts/installation.js` It needs adding both to the options, defined at the beginning of the file, and the OS detection, figured out at the end. I can help you with this if you need it.

linusbobcat:
I fixed the above two issues. The only remaining issue is fixing how the images wrap around the text, although I don't know what CSS to change.

linusbobcat:
Apologies for the late response, but I don't know what's the best and preferred way to make text align right from the images rather than fully wrap around.

lewisgoddard:
Unfortunately the document you're working on has changed format and location, so this branch will never be mergeable. It is probably best to branch anew and add any content there. `/installation.php` is now located at `/docs/installation.md`

For future reference there is a class for that, `<img class="float-left" ..` will make a margin in the appropriate places.

PR #449, opened 2015-03-21
Serbian Ijekavian

### LINKED ISSUE #444: Serbian Ijekavian localization returns 404 error
I'm not sure what is the best way to go about fixing this problem.

a) Transifex's code for Ijekavian dialect of Serbian language is sr@Ijekavian.
b) eOS's beta site is configured to have localized URL in this case as http://beta.elementaryos.io/sr@Ijekavian/

I'm guessing that the '@' in the URL is causing the URL to fail. If I am right, these are the questions:

- Is it possible to to have the /mvp/lang/sr@Ijekavian directory renamed, but keep its synchronization with Transifex's translations?
- What are the chances to ask Transifex to change the existing code from 'sr@Ijekavian' to something else, like 'sr_Ijekavian'?

If I am not right (if the '@' in the URL isn't the cause of this problem), does anyone else have an idea why this URL is failing?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/9307853-serbian-ijekavian-localization-returns-404-error?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github).
</bountysource-plugin>

emersion:
Fixes #444. The Nginx server config needs to be updated @fabianthoma 

To test, go to `/sr_Ijekavian`.


lewisgoddard:
I am unsure if this has already been partially implemented (current links) but remains broken. Seeing as it's not mergeable, maybe re-base and I'll ~~harass~~ ping fabian to fix it.

emersion:
This branch is now mergeable.

PR #452, opened 2015-03-22
Updated local development instructions.

agats:
Small update, added the required PHP version to develop locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ Simple Local Development
 
 For a simple local development environment, you will need:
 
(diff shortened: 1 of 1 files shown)


lewisgoddard:
This could actually be simplified to just be any [supported version](http://php.net/supported-versions.php), so it never needs updating, but this will do for now.

agats:
That's good idea. I had to have at least 5.4 to get the built-in web server. The packages on my machine installed an older version.

PR #460, opened 2015-04-02
Update download links to RC1

lewisgoddard:
This update changes:
- the download title.
- the download size.
- the direct download links.
- the magnet download links.

This does **not** change:
- the get involved graph towards completion.
- any other reference outside the index page about this being a "beta".


danrabbit:
I would like to not do this since a final announcement will follow very shortly

PR #475, opened 2015-04-10
Add Code Documentation

lewisgoddard:
* Add Code Documentation
* Add `docs` and `docs-index` classes
* Add `index.php` usage in `/docs`
* Redirect to starting point in `/docs/code`


fabianthoma:
Awesome, but there is no styling to code blocks, so they look all weird. Other than that and the links not working on branches this is OK to merge from my side.

PR #486, opened 2015-04-11
Publish proper config.

lewisgoddard:


emersion:
Not mergeable anymore.

lewisgoddard:
@emersion Could you give this a look? Helpful for those of us that don't just use `router.php`.

nginx.conf
```diff
@@ -1,9 +1,23 @@
...
 location / {
     rewrite "^/([a-z]{2}(?:_(?:[A-Z]{2}|[A-Z][a-z]+))?)/(.*)$" /$2?lang=$1 last;
     try_files $uri $uri.html $uri/ @extensionless-php;
 }
 
+location @extensionless-php {
+    rewrite ^(.*)$ $1.php last;
+}
+
+location ~ \.php$ {
```
emersion: This is specific to the platform, is it needed? (e.g. not the same address on Debian and Arch)

emersion:
I still get a 403 error when accessing to `/docs/`

lewisgoddard:
I should be able to fix that by allowing index browsing.

nginx.conf
lewisgoddard: `location ~ \.php$ {` ?

I would have thought that applies to all Nginx installations.

emersion: See https://wiki.archlinux.org/index.php/Nginx#nginx_configuration

PR #488, opened 2015-04-11
Added file to fetch tweets favorited by user

voldyman:
just need to add the twitter keys to a config file and run fav-tweets.php as a cron job.

we will get a json file tweets.json with all the required info.

How-to:
* Create a Twitter app from http://dev.twitter.com
* Edit `backend/config.php`, write consumer key & secret
* See https://dev.twitter.com/oauth/overview/application-owner-access-tokens to get access token & secret

### DIFF
--- a/backend/fav-tweets.php
+++ b/backend/fav-tweets.php
@@ -133,11 +133,14 @@ class TwApi {
         return $result;
     }
 
-    function get_favorites($user_id, $count = 10) {
+    function get_favorites($user_id = '', $count = 10) {
         $request = 'favorites/list.json';
-        $params = array('user_id' => $user_id,
-                        'count' => $count);
+        $params = array();
+        $params['count'] = $count;
 
+        if ($user_id != '') {
+          $params['user_id'] = $user_id;
+        }
         $result = $this->send_request($request, $params);
 
         return $result;
@@ -149,7 +152,7 @@ $api = new TwApi($config['twitter_access_token'],
                  $config['twitter_access_secret']);
 
 log_info('Fetching tweets.');
-$favs = $api->get_favorites('elementary');
+$favs = $api->get_favorites();
 log_info('Fetched tweets');
 
 $tweets = array();
--- a/backend/fav-tweets.php
+++ b/backend/fav-tweets.php
@@ -5,25 +5,31 @@
  * access tokens should be stored in twitter-tokens.php
  */
 
-// this file defines the keys required by twitter in the following way
-// define('TWITTER_CONSUMER_KEY', 'val');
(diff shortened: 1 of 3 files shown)


gamerlv:
Instead of requiring a new config file could you please use the existing `backend/config.php`

It looks like this file is never to be called from the user end, if that is so some method of restricting it to a few IP's (eg 127.0.0.1) or just CLI would be usefull

emersion:
How-to:
* Create a Twitter app from http://dev.twitter.com
* Edit `backenbd/config.php`, write consumer key & secret
* See https://dev.twitter.com/oauth/overview/application-owner-access-tokens to get access token & secret

- [ ] PHP notice when running from CLI: `PHP Notice:  Undefined index: REMOTE_ADDR in /home/simon/public_html/mvp/backend/fav-tweets.php on line 10`
- [ ] Could not test from localhost - `$_SERVER['REMOTE_ADDR']` was `::1`
- [ ] 4 spaces indentation - see https://github.com/elementary/mvp#code-style
- [ ] I executed the script from the root (`php -f backend/fav-tweets.json`), so the `tweets.json` file was written to the root instead of `backend/`
- [ ] Try not to mix `'` and `"` - use only `'` and `'string'.$var`
- [ ] After executing the script, `tweets.json` only contained `[]` - maybe because I'm not logged in with the elementary account?
- [ ] It's generally a good practice to use protected/private variables in classes. Use `protected` instead of `var`.

Thanks! :-)

voldyman:
For the 6th point, you need to add keys in the config file

emersion:
I've added keys in the config file. Do you execute the script with your own Twitter account?

voldyman:
yes, i have tested that

backend/fav-tweets.php
```diff
@@ -0,0 +1,175 @@
...
+        $this->consumer_secret = $config['twitter_consumer_secret'];
+
+        $this->oauth = array(
+            'oauth_consumer_key' => $this->consumer_key,
+            'oauth_token' => $this->access_token,
+            'oauth_signature_method' => 'HMAC-SHA1',
+            'oauth_timestamp' => time(),
+            'oauth_nonce' => time(),
+            'oauth_version' => '1.0'
+            );
```
emersion: Indentation problem here ;-)

backend/fav-tweets.php
```diff
@@ -0,0 +1,175 @@
...
+        $request = 'users/lookup.json';
+        $params = array('user_id'=>$list);
+
+        $result = $this->send_request($request, $params);
+        return $result;
+    }
+
+    function get_favorites($user_id, $count = 10) {
+        $request = 'favorites/list.json';
+        $params = array('user_id' => $user_id,
```
emersion: Change this to `screen_name`.

See https://dev.twitter.com/rest/reference/get/favorites/list

emersion:
LGTM, thanks a lot! :-D

PR #491, opened 2015-04-12
Make HIG and Code links open in same tab

### LINKED ISSUE #490: HIG and Code links open in new tab
The HIG and code links at http://elementary.io/developer open in a new tab, but since they're on the same site now we can remove the `target=_blank`.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/10962333-hig-and-code-links-open-in-new-tab?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github).
</bountysource-plugin>

cassidyjames:
fixes #490

### DIFF
--- a/developer.php
+++ b/developer.php
@@ -13,13 +13,13 @@
                 </div>
                 <div class="column half developer-sections">
                     <div id="dev-section-1" class="dev-section">
-                        <a href="http://elementary.io/docs/human-interface-guidelines" title="Design" target="_blank">
+                        <a href="/docs/human-interface-guidelines" title="Design">
                             <h5>Design</h5>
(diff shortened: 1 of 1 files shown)


emersion:
It would be better to make links relative (`docs/[...]`) and not absolute (`/docs/[...]`), in order to make them work in `/branch/<whatever>` too.

lewisgoddard:
@emersion We're looking at using sub-domains instead of branches, and all the established docs use absolute paths. I don't think that's a blocker, but a separate bug.

PR #492, opened 2015-04-12
Fix Chinese languages' names

alphachung:
Changed from vague meanings of "Mandarin" and "Taiwan" to "Simplified Chinese" and "Traditional Chinese". Both mainland China & Taiwan speak mandarin, and also "mandarin" was written in traditional chinese while taking you to the simplified chinese page, and "taiwan" was written in simplified chinese but takes you to the traditional chinese page. By changing them to "Simplified Chinese" written in simplified chinese and "Traditional Chinese" written in traditional chinese, it fixes the problem and makes it clearer which variety of chinese it is.


emersion:
Thanks!

PR #495, opened 2015-04-12
Fixes the 404 page bug

emersion:
How to reproduce:

1. Go to the main page
2. Change language
3. Append random chars to the URL
4. White 404 page

### DIFF
--- a/404.php
+++ b/404.php
@@ -1,4 +1,5 @@
 <?php
+echo '404 included<br>';
     include '_templates/sitewide.php';
     $page['title'] = 'Page Not Found &sdot; elementary';
     include $template['header'];
--- a/_templates/header.php
+++ b/_templates/header.php
@@ -1,4 +1,5 @@
 <?php
+echo 'header included<br>';
 include_once __DIR__.'/l10n.php';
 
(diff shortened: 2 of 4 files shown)


emersion:
Seems a problem with nginx config, the 404 page isn't even executed. See http://beta.elementary.io/branch/fix-404/?lang=fr (we can see that `_templates/` files are included) and http://beta.elementary.io/branch/fix-404/fdsafdsa?lang=fr (still white, should print at least _404 included_).

emersion:
One possibility: it loads the 404 page from the root and not from my branch so I'm not able to debug.

lewisgoddard:
Yeah, it will be loading the 404 page from root. You'll need to get @fabianthoma to update the config just for your branch.

emersion:
Okay, will setup nginx locally too. :-P

lewisgoddard:
@emersion Is this good to review then? Labels suggest not.

emersion:
No, I have to test locally the nginx config and see why it's not working, since there is no problem with the php dev server.

lewisgoddard:
The test URLs seem to have been completed by subsequent work, probably in #683

PR #499, opened 2015-04-12
Refactoring l10n

emersion:
Created a new class `Translator`.

Thanks @vimishor for the feedback!


vimishor:
Like I said in the chat, more descriptive commit messages would be nice and a bug fix is not a "_minor fix_" from my point of view.

You should tag it accordingly and send a separate PR, because in order to merge that bug fix, you force other people to also accept other changes that are not related to that bug.

emersion:
All right. Just refactored all code in `l10n.php`.

_templates/l10n.php
```diff
@@ -35,334 +37,368 @@ function list_langs() {
...
+        if (empty($lang)) {
+            $lang = $this->get_page_lang();
+        }
+        $this->lang = $lang;
     }
-    if ($lang == 'en') {
-        return true;
+
+    public function init() {
+        global $sitewide, $page; // Global site variables
```
vimishor: With or without a class, globals are still bad. Use dependency injection instead of globals.

emersion: Is decided to keep these global variables because they are global site variables, ie. _safe_ (used on every page, breaks the whole website if corrupted). Do you think it's still important to pass them as arguments?

_templates/l10n.php
```diff
@@ -35,334 +37,368 @@ function list_langs() {
...
- * (Useful to extract strings from a file for instance.)
- */
-function translate_html($input, $translate = 'translate') {
-    global $l10nDomain;
+    /**
+     * Translate a HTML string. This includes a minimalist XML parser.
+     * Can be provided a $translate callback to override default behaviour.
+     * (Useful to extract strings from a file for instance.)
+     */
+    public function translate_html($input, $translate = null) {
```
vimishor: This method is doing more than the name suggests. 

It should just translate a string, not parse XML, doing security, etc. You could split it in separate methods and use a proper [XML parser](http://php.net/manual/en/book.xml.php) instead of inventing your own.

_templates/l10n.php
```diff
@@ -35,334 +37,368 @@ function list_langs() {
...
 
-/**
- * Translate a HTML string. This includes a minimalist XML parser.
- * Can be provided a $translate callback to override default behaviour.
- * (Useful to extract strings from a file for instance.)
- */
-function translate_html($input, $translate = 'translate') {
-    global $l10nDomain;
+    /**
+     * Translate a HTML string. This includes a minimalist XML parser.
```
vimishor: Why ? Use a proper [XML parser](http://php.net/manual/en/book.xml.php) instead of inventing your own.

_templates/l10n.php
```diff
@@ -35,334 +37,368 @@ function list_langs() {
...
+        return self::LANG_DIR.'/'.$lang;
     }
 
-    if (empty($_SERVER['HTTP_ACCEPT_LANGUAGE'])) {
-        return null;
+    public function is_lang($lang) {
+        if (!is_string($lang)) {
+            return false;
+        }
+        if (!preg_match('#^[a-z]{2}(_([A-Z]{2}|[A-Z][a-z]+))?$#', $lang)) {
```
vimishor: If all supported langs are available in `available_langs` array, why not use that for this check ?

_templates/l10n.php
vimishor: You started to create this class after our discussion about the current procedural code which abuses of globals all over the place which makes it hard to test.

Unit testing should be done in isolation and when your unit depends on globals and super globals is hard to test it. Its not impossible, but you will end up with a pretty long bootstrap for that test.

Your unit should know that it needs param one, two and three in order to work and that's it. It should not care from where those parameters came from.

_templates/l10n.php
emersion: Available languages are a list displayed to the user in the footer. But it's better not to rely on it to be able to test incomplete translations.

_templates/l10n.php
emersion: First of all, I think this is not the purpose of this PR. This PR is refactoring the translation code in order to make it cleaner.
Then, I don't agree : parsing the whole XML tree is far slower than just analyzing the HTML code to find strings to translate. If we were using a parser, we should first process the whole HTML page, find strings to translate, and then format the parsed HTML to send it to the client.
Anyone else (@lewisgoddard for instance, if you have two minutes!) could give his advice?

_templates/l10n.php
vimishor: > But it's better not to rely on it to be able to test incomplete translations.

Nope. If its not complete it should not be in production.

danrabbit: *ALL* of the translations are incomplete. But that doesn't mean they aren't useful. The home page, get involved page, and support pages are only like 30% of the strings

_templates/l10n.php
vimishor: > This PR is refactoring the translation code in order to make it cleaner.

Dropping some code inside a class doesn't make it cleaner in any way.

> Then, I don't agree : parsing the whole XML tree is far slower than just analyzing the HTML code to find strings to translate.

Did you measured that or is just an assumption ?

Since you want speed, you are using the wrong approach.

Parsing the output in order to find something that you could translate is the wrong approach and my guess one of the slowest. You should translate a simple string, not the entire output.

_templates/l10n.php
vimishor: I'm skipping the fact that incomplete translations are in production and even so there is no reason to make all that verification just to see if a translation exists.

If the translation exists in `available_langs` array load it and if it's incomplete it should fall back to English for what is missing. 

But wait, we can't do that, because we are parsing the entire output and we can't be sure what gets translated from that.

emersion: It's the current behavior. It's working like that. You can check on beta.elementary.io if you want to.

_templates/l10n.php
emersion: As said in the chat, I benchmarked this algorithm.

In terms of complexity, it is in O(n), which is far better compared to a whole parsing, translation and formatting.

_templates/l10n.php
emersion: As said in the chat, we probably won't make unit tests. The aim of this PR is to solve bad practices such as the $lang variable.

lewisgoddard:
Not currently mergeable.

emersion:
Done ;-)

lewisgoddard:
Not mergeable again.

emersion:
@lewisgoddard Conflicts resolved!

lewisgoddard:
Well, this appears to work now.

PR #507, opened 2015-04-14
Enable SSL when available using an asynchronous HSTS header. Fixes #258

### LINKED ISSUE #258: HTTPS should be enforced when available [$50]
Although we do have HTTPS support enabled, for some reason, it's not currently on by default. I think that it is time we enable the visitors to browse the site securely by default.

<bountysource-plugin>
---
There is a **[$50 open bounty](https://www.bountysource.com/issues/8113221-https-should-be-enforced-when-available-50?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github).
</bountysource-plugin>

lewisgoddard:
Loading http://beta.elementary.io/branch/hsts-conditional/ will attempt to load a fixed, secure resource. This will enable HTTPS on all subsequent requests if available, or fail if not.

### DIFF
--- a/_templates/header.php
+++ b/_templates/header.php
@@ -80,7 +80,7 @@ begin_html_l10n();
         
         <script>
             (function(d,s,f){g=d.createElement(s),u=d.getElementsByTagName(s)[0],g.async=1,g.src=f,u.parentNode.insertBefore(g,u)})
-        	(document,'script','https://beta.elementary.io/branch/hsts-conditional/backend/hsts.php')
+            (document,'script','https://beta.elementary.io/branch/hsts-conditional/backend/hsts.php')
         </script>
 
         <script>
--- a/_templates/header.php
+++ b/_templates/header.php
@@ -77,6 +77,11 @@ begin_html_l10n();
         <link rel="stylesheet" type="text/css" media="all" href="https://fonts.googleapis.com/css?family=Raleway:100|Open+Sans:300,400,600|Droid+Sans+Mono">
         <link rel="stylesheet" type="text/css" media="all" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.2.0/css/font-awesome.min.css">
         <link rel="stylesheet" type="text/css" media="all" href="styles/main.css">
+        
+        <script>
+            (function(d,s,f){g=d.createElement(s),u=d.getElementsByTagName(s)[0],g.async=1,g.src=f,u.parentNode.insertBefore(g,u)})
+        	(document,'script','https://beta.elementary.io/branch/hsts-conditional/backend/hsts.php')
+        </script>
 
         <script>
             <?php include __DIR__.'/../scripts/jql.min.js'; ?>
--- a/_templates/header.php
+++ b/_templates/header.php
@@ -80,7 +80,7 @@ begin_html_l10n();
         
         <script>
             (function(d,s,f){g=d.createElement(s),u=d.getElementsByTagName(s)[0],g.async=1,g.src=f,u.parentNode.insertBefore(g,u)})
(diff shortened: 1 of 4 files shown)


lewisgoddard:
Fixes #258

_templates/header.php
```diff
@@ -77,6 +77,11 @@
         <link rel="stylesheet" type="text/css" media="all" href="https://fonts.googleapis.com/css?family=Raleway:100|Open+Sans:300,400,600|Droid+Sans+Mono">
         <link rel="stylesheet" type="text/css" media="all" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.2.0/css/font-awesome.min.css">
         <link rel="stylesheet" type="text/css" media="all" href="styles/main.css">
+        
+        <script>
+            (function(d,s,f){g=d.createElement(s),u=d.getElementsByTagName(s)[0],g.async=1,g.src=f,u.parentNode.insertBefore(g,u)})
+        	(document,'script','https://beta.elementary.io/branch/hsts-conditional/backend/hsts.php')
```
gamerlv: This url needs to be dynamic, otherwise the script will fail after the old branches are deleted

backend/hsts.php
```diff
@@ -0,0 +1,17 @@
...
+
+// "preload" would not do anything as we don't meet the requirements
+// https://hstspreload.appspot.com/
+
+// Compatibility
+// http://caniuse.com/#feat=stricttransportsecurity
+
+// We're going to pretend this is JavaScript, so we can log something.
+header('Content-Type: application/javascript');
+?>console.log('HSTS Loaded');
```
gamerlv: A console.log will cause a javascript error on any internet explorer version if the console isn't opened. I'm not sure what other browsers do (esp, mobile)

_templates/header.php
lewisgoddard: I'm not sure if it's possible to have a relative path with a separate protocol, I would change it to be fixed to root.

backend/hsts.php
lewisgoddard: What would be better? JSON `{"success":true}` ? A 204 No Content header?

lewisgoddard: Changed to a 204 header, no content.

_templates/header.php
gamerlv: Yes you can't, but I said dynamic. This is a php file, it's possible to determine the needed path. If only there is a check for `$_SERVER['HTTP_HOST']` Maybe also some looking in `$_SERVER['REQUEST_URI']` to get the branch path.

This is probably beter served as a function if it not already exists

backend/hsts.php
gamerlv: Any of those would do, just not a `console.*`

I would recommend returning a json code and check against that. To double check it's actually going though and not being cached by CF or something.

lewisgoddard:
Ready for merge with approval.

lewisgoddard:
Correct behaviours has been exhibited in a new Firefox installation, where the initial page load is on HTTP, but all future requests happen over HTTPS.

danrabbit:
I can confirm that it works appropriately in Chrome and Midori

emersion:
Anyone tried on Midori on Luna? I can do it right now if you want.

lewisgoddard:
@emersion Sure, hopefully it has no effect.

emersion:
@lewisgoddard Just checked, it has no effect ;-)

PhillipSz:
So if I use chrome or Firefox on android I don't get redirected to https. If I type it and the try to load http, hsts does work however.

PR #508, opened 2015-04-14
Fixes buttons heights in get-involved

### LINKED ISSUE #489: Get Involved: variable height on header buttons in some languages [$15]
As seen here:
screenshot from 2015-04-12 00 10 04

<sub>BTW these translations are terrible, but I'm not going to create another account to fix them. And when I tried improving translations on Launchpad people kept reverting to the terrible, incomprehensible, unnatural word-for-word translations arguing that they "respect the Gnome translation guidelines".</sub>

<bountysource-plugin>
---
There is a **[$15 open bounty](https://www.bountysource.com/issues/10950398-get-involved-variable-height-on-header-buttons-in-some-languages?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github).
</bountysource-plugin>

emersion:
Fixes #489.

screenshot from 2015-04-14 18-29-28

### DIFF
--- a/styles/get-involved.css
+++ b/styles/get-involved.css
@@ -198,6 +198,7 @@ section:nth-child(5) .button.flat.suggested-action:focus {
     border: 2px solid #c8d0d0;
     border-radius: 4px;
     min-width: 200px;
+    vertical-align: middle;
 }
 
 .row.hero .actions .button.flat:nth-child(1) {


danrabbit:
If possible, I'd prefer they all have the same height instead of just centering

emersion:
@danrabbit Seems that it's not possible with CSS http://stackoverflow.com/questions/6378524/float-variable-height-containers

lewisgoddard:
Hm, I think it _is_ technically possible, with a little bit of magic, but its certainly not as simple as it should be.

emersion:
Some suggest to use a `<table>`.

lewisgoddard:
Urgh, unnecessary. Nothing but tabular data should be in a table. Semantics.

emersion:
Ow, seems that we could use `display: flex;` :-D

http://caniuse.com/#feat=flexbox

danrabbit:
Meh, nbd. If someone opens a duplicate ticket, we can re-address it then.

PR #513, opened 2015-04-15
Update /get-involved to Freya Update 1

lewisgoddard:
Add chart back. Testable at https://beta.elementary.io/branch/chart-back/get-involved


danrabbit:
Unsure if we want to do this right now. The update1 milestone is really empty. Maybe we want to do it if we add more items there?

lewisgoddard:
I did ping cassidy, he said there would be more later. Maybe have a little push of some bugs to it and see what you think.

lewisgoddard:
screenshot from 2015-06-20 18 26 08

PR #516, opened 2015-04-15
link to old hig for now

danrabbit:
Because the new hig is not finished importing


emersion:
Is there a link to HIG in get-involved?

danrabbit:
There is! Good catch. Fixed

lewisgoddard:
Do we want to redirect with the index.php that's currently there?

PR #518, opened 2015-04-15
point to old hig instead of coming soon

danrabbit:


lewisgoddard:
Should we delete the `coming-soon.md` then?

PR #530, opened 2015-04-16
Update compressed size of home images

mateuspv:
Hi, I compress the size some main images of home, this change 2.6MB to 909.4 kb, with is greeat IMHO
Note: 
 - Some images have a small difference, better check on `swipe` mode on github.
 - PNG loses loses editable layers.


lewisgoddard:
What did you use to compress these images? I tend to recommend tinypng.com

mateuspv:
Is tinypng.com, I think is the best for size/quality.

PR #540, opened 2015-04-18
Prevent scrolling behind language popover

### LINKED ISSUE #533: Prevent scrolling behind language popover [$10]
When you open it up, if you scroll up the page will start to scroll once you reach the top. This is pretty annoying. We should block scroll events from passing through the popover.

<bountysource-plugin>
---
There is a **[$10 open bounty](https://www.bountysource.com/issues/11317834-prevent-scrolling-behind-language-popover?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github).
</bountysource-plugin>

alexgleason:
Fixes #533

This implementation uses pointer events. See browser compatibility here: http://caniuse.com/#feat=pointer-events

When the popover is moused over, it disables scrolling for the body and enables it only for the popover. When the pop over is moused out, the original settings are restored.

### DIFF
--- a/scripts/popover.js
+++ b/scripts/popover.js
@@ -23,4 +23,15 @@ $(function () {
 			$popover.removeClass('active');
 		});
 	});
+
+	// Prevent scrolling behind popover
+	$('.popover').hover(
+		function() {
+			$('body').css('pointer-events', 'none');
+			$(this).css('pointer-events', 'all');
+		}, function() {
+			$('body').css('pointer-events', '');
+			$(this).css('pointer-events', '');
+		}
+	);
 });


emersion:
This does not solve the issue: when scrolling the window, the mouse is out of the popover so it is still possible.

You cannot solve this with `pointer-events` because we need to detect when the user clicks out of the popover to close it. Try instead to listen to the `scroll` event. ;-)

Thanks for contributing!

danrabbit:
Closing since we fixed it in another branch

PR #545, opened 2015-04-18
Makes footer translatable

### LINKED ISSUE #510: Footer not translated
The footer doesn't get translated when choosing a language. It always says "Copyright" and "Language" in English.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/11155869-footer-not-translated?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github).
</bountysource-plugin>

emersion:
Fixes #510.

### DIFF
--- a/_templates/footer.php
+++ b/_templates/footer.php
@@ -4,7 +4,7 @@ set_l10n_domain('layout');
         </div>
         <footer>
             <p>
-                Copyright &copy; <?php echo date('Y'); ?> elementary LLC. <a href="<?php echo $page['lang-root'].'privacy-policy'; ?>">Privacy Policy</a>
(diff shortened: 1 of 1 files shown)


lewisgoddard:
I am confused as to how the code here actually fixes the problem in the title.

emersion:
`$page['lang-root']` contains `/fr/` if the current language is _fr_, but `$sitewide['root']` contains `/`. If we use the first one, `/fr/` will appear in the translation key so there will be one key per language.

lewisgoddard:
Okay, so this was breaking it, rather than the footer not being translated.

PR #546, opened 2015-04-18
Update prefer css class selector over ids and tags

mateuspv:
Hi, some small changes:
- Ids: are unique, can be only used once on page, also, it create javascript variables in `window`, like (```'amount-custom' in window```)

- tags: so much generic, stylize footer by example, we can never have other footers in the same page, it will give conflict

### DIFF
--- a/_templates/footer.php
+++ b/_templates/footer.php
@@ -2,7 +2,7 @@
 set_l10n_domain('layout');
 ?>
         </div>
-        <footer>
+        <footer class="footer">
             <p>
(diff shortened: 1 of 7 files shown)


lewisgoddard:
I think the use of IDs is generally for precisely that reason: unique objects.

`#amount-custom` appears only once, and you have broken much of the javascript by changing it.

Also, the developer ids, while non-descriptive, are useful for scrolling to. Perhaps replacing them with nicer names as well as the additional classes would be better.

mateuspv:
Oks...could you close this at moment?
* I prefer data attributes for javascript, easy to change style without change scripts.

lewisgoddard:
I too like data attributes, but usually only for data. I use classes like `.js-target-amount-custom` usually, but much of this was started before me. Semantically, we _should_ never have two footers on the same page though.

PR #556, opened 2015-04-19
block scrolling behind an open popover

danrabbit:
* Change tabs to spaces (as god intended!)
* set overflow hidden and block scroll events while popover is open


danrabbit:
One thing I was trying to get working was allowing you to scroll the page and closing the popover when you do. I added 'scroll touchmove mousewheel' to the document function at the bottom, but the problem is that the popover anchor was remaining active so you couldn't reopen it without first clicking the body

emersion:
Can you use the `.blur()` method on the anchor?

Issue #478, opened 2015-04-11
Add Patreon and Bountysource [$30]


fabianthoma:
I guess this needs no further description, other than the question as to where.

danrabbit:
Someone had a branch at one point that added monetary stuff to "get involved". As in get involved by funding us. Not sure what happened to the branch, but could be a place for it.

danrabbit:
Found the PR with funding stuff https://github.com/elementary/mvp/pull/232

lewisgoddard:
Fixed by #562

PR #562, opened 2015-04-20
Get funding

danrabbit:
* Adds "Funding" section to Get Involved with links for Paypal, BountySource, and Patreon
* Solves #478 

- [x] Needs Copy
- [x] Footer bg color is not correct
- [x] Suggested action buttons need a focus color


lewisgoddard:
Code-wise, you are approved up to here.

PR #585, opened 2015-04-24
Carousel Slider fix

jurosh:
Carousel Slider fix - removed second scroll-bar while switching items.


emersion:
LGTM, thanks! :-)

Issue #580, opened 2015-04-23
Add link buttons for headers in docs


danrabbit:
BootStrap has this neato feature where hovering over a header in the docs shows a handy link button. I like that shit http://getbootstrap.com/components/#glyphicons-glyphs

seanballais:
I think those are called 'anchors.'

PR #589, opened 2015-04-24
Add anchor links to h1s through h6s

cassidyjames:
Thoughts?


cassidyjames:
Ah, need to pull in master. One sec.

lewisgoddard:
_Meniacal laughter._

danrabbit:
fixes #580

cassidyjames:
Thank god for Meld.

scripts/docs/main.js
```diff
@@ -19,6 +19,11 @@ $(document).ready(function() {
         hljs.highlightBlock(block);
     });
 
+	// Anchor headings
```
cassidyjames: Whoopsie, meld put in a tab instead of 4 spaces. Fixing...

lewisgoddard:
I can't hover over the link on `h2`s

cassidyjames:
@lewisgoddard ah, the icon isn't actually hoverable. The whole heading is. Lemme poke at it.

lewisgoddard:
I'd go with what I said on Slack, prepend an `<i class="fa fa-link"></i>` on the `h#` and just hide it with CSS when not hovered.

danrabbit:
- [ ] Sidebar is busted. No subheadings

lewisgoddard:
You've broken the next page links a little too. Maybe we should add a class to those that can be targeted with `:not('.next-page')`

cassidyjames:
@lewisgoddard hm, imho the next page links should not be `<h4>`s. They're not headings.

lewisgoddard:
I think that's what they were previously, and I can't add classes or ids to paragraphs.

danrabbit:
Seems to work as expected for me. Link can be copied. Open in new tab works. Smooth scrolling still works.

PR #605, opened 2015-04-25
Adds filesize script

### LINKED ISSUE #60: Set file size programmatically
So that we don't have to modify this page every time the iso changes (creating potential for human error), we should be setting the file size of the ISO programmatically.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/7658599-set-file-size-programmatically?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github).
</bountysource-plugin>

emersion:
Fixes #60.

Probably needs a cronjob or something.


lewisgoddard:
What file does this get the size for? The latest upload? That's not necessarily accurate.

Maybe it would be better to have the sourceforge and magnet links as variables in PHP, then you can dissect the 64-bit one into a more specific API call.

emersion:
@lewisgoddard Better like this?

lewisgoddard:
That's better, I approve your changes. Will you approve mine?

emersion:
Yes, thanks!

PR #612, opened 2015-04-27
Docs hig merge

danrabbit:
Merges the HIG into a single massive (but now searchable) document. A good review of this branch would make sure:
- [x] All the links work correctly
- [x] We're not missing sections
- [x] Things are in a sane order
- [x] Heading levels make sense


cassidyjames:
- [ ] Duplicate IDs (i.e. `#usage`)

lewisgoddard:
The link at _"Portions of this page are derived from work created and shared by the [Android Open Source Project](https://code.google.com/policies.html) and used according to..."_ is broken. Do we want to use http://developer.android.com/license.html as the current license page for Android documentation.

PR #614, opened 2015-04-28
Fixes anchors

emersion:


scripts/smooth-scrolling.js
```diff
@@ -8,33 +8,44 @@ $(function () {
...
+            }
+            scrollTop = $target.offset().top;
         }
 
         // Smooth scrolling
         $('html, body').stop().animate({
             scrollTop: scrollTop
         }, 'normal', function () {
-            window.location.hash = href;
+            window.location.hash = '#'+href;
```
lewisgoddard: You shouldn't need to prefix the href with an actual hash, it does that automatically.

scripts/smooth-scrolling.js
```diff
@@ -8,33 +8,44 @@ $(function () {
...
     });
+
+    // Fix anchors
+    // Let other scripts add links to DOM before fixing them
+    setTimeout(function () {
+        $('a[href^="#"]').each(function () {
+            var href = $(this).attr('href');
+            $(this).attr('href', window.location.pathname+href);
+        });
+    }, 0);
```
lewisgoddard: I'm unclear on how this helps if the timeout is zero.

scripts/smooth-scrolling.js
```diff
@@ -8,33 +8,44 @@ $(function () {
         // Get link href
         var $anchor = $(this);
         var href = $anchor.attr('href');
-        var domain = new RegExp('/' + window.location.host + '/');
         if (
             href[0] !== '#' &&
-            domain.test(encodeURIComponent(href))
+            href.indexOf(window.location.host) == -1
```
lewisgoddard: This is indeed better.

scripts/smooth-scrolling.js
emersion: >  `setTimeout(function () {}, 0)` re-queues the new JavaScript at the end of the execution queue.

See http://stackoverflow.com/questions/779379/why-is-settimeoutfn-0-sometimes-useful

It's the intended behavior since scripts executed after this one are adding other links to the DOM.

lewisgoddard:
This will need to be merged again, it seems to break the download dialog.

PR #618, opened 2015-04-29
update deps to mention php curl

danrabbit:


lewisgoddard:
Maybe mention it expects `php5-json` as well as `php5-curl`?

PR #630, opened 2015-05-05
Added some german translation

c143:


KeitIG:
We don't accept PR for translations, we use Transifed instead, see there: https://github.com/elementary/mvp/blob/master/TRANSLATE.md :)

PR #635, opened 2015-05-06
Actually make the blog link internal, now it has internal navigation.

lewisgoddard:
On https://elementary.io/ the "Blog" link opens in a new tab.

On https://beta.elementary.io/branch/blog-internal/ the "Blog" link does not, as intended.

PR #638, opened 2015-05-07
Stop subdomain links opening in new tab

roryjs:
Fix regex matching. Stops blog.elementary.io opening in new tab.


lewisgoddard:
I think we prefer #635 since that eliminates the regex entirely.

Issue #634, opened 2015-05-06
Donate button for website besides the one for download


Thegoru:
Please add a visible donate button besides the one for when you download it (I would actually put the option of $0 available for the download button). Here is why, I did not donate at first (used the Custom = 0), since I wanted to try it first but after people see how awesome it is they will want to donate. You don't want to make people feeling like you are trying to trick them (especially us nerds).So it would be great if you add an option of $0 for the download and give people to opportunity to come back and donate by adding a visible donate option. Also, when you click on the donate button for the download to enter your credit card currently it says "Pay", I suggest changing that to "Donate". It just sounds better for public relations.

snwh:
But it's not a donate model. It's a pay-what-you-want model.

Thegoru:
You guys already have the funding option on "get involved - funding" for Patreon, I would just suggest at-least bringing that option the the main page for returning folks like me. For the download, the option of 0 should be there and visible, in my opinion since it is open source. For example in my case, I would not pay for a distro until I try it and see that I like it, which is what I did in this case and had to cancel the second download to pay for the awesome OS. I think there would be other people who would do that, too.  To summarize, there should be a visible option to come back and contribute (pay/donate) to the project without any hassle.

danrabbit:
Sorry OP, we're going to stick with making people type 0. We really on funding to keep elementary afloat. 

#624 is about adding Stripe (Credit and bitcoin) to the get involved/funding page.

We purposely don't use the word "Donate" because we're not a non-profit organization and we don't want to trick people into thinking we are. Your payments are not tax-deductible.

Thegoru:
I understand. I guess just adding that option for returning folks who did not pay initially would be nice without having to click the download button. 

Just for kicks and beyond, below is a suggestion on how to make people pay you (I just read your blog on the whole controversy with the pay-what-you-want model). I do believe you guys should get paid for the great work, hopefully without pissing off to many people. 

Remembering learning this in a class, make people feel like they are being watched to act just. Plato wrote about it, too. I think it could be applied to the open source software to increase contributions. Put some eyes on the download screen and add a 0 option or not? Could get creative with it. There is a lot of research that says it might work.

Here is an article about it:
http://www.scientificamerican.com/article/how-the-illusion-of-being-observed-can-make-you-better-person/

Thanks for the response!

PR #655, opened 2015-05-13
Add new 0$ button

maxrimue:
This has bothered me a bit when I downloaded elementary myself, I'm aware that you can click custom and then just type "0", but it might seem confusing for some other people. I think the goal of elementary is to be attractive to all kinds of users, and some might think that you have to donate for downloading elementary.
Also, I don't think this change would have an impact on how many people donate, as they should do it only if they want anyway.

### DIFF
--- a/index.php
+++ b/index.php
@@ -39,6 +39,7 @@
                             <?php
                         } else {
                             ?>
+                    <button id="amount-zero"        value="0"  class="small-button payment-button target-amount">0</button>
                     <button id="amount-five"        value="5"  class="small-button payment-button target-amount">5</button>
                     <button id="amount-ten"         value="10" class="small-button payment-button target-amount checked">10</button>
                     <button id="amount-twenty-five" value="25" class="small-button payment-button target-amount">25</button>


lewisgoddard:
Seeing as less than half of our downloads come through the site anyway, and more than 99% are currently not paying on the site, I don't think this is as confusing as you think. If more than 99% of people understand how to do it, then that's actually some pretty damn good UI. On the other hand, this change of actively _not_ contributing, rather than having the default of nothing, has nearly doubled the amount of people paying for their download, which has further has enabled elementary to place over $15,000 on bug bounties, a vast majority of which has been claimed already.

The design decisions behinds this have been previously explained in the blog post [Payments (Or, Why we make you type “$0”)](http://blog.elementary.io/post/110645528530/payments), with a follow-up post on our spending detailed in a [thank you](http://blog.elementary.io/post/118806943356/thank-you-everyone) post.

lewisgoddard:
Functionally, the code is sound. It will add a `$ 0` button to the left of existing buttons, keeping the left-to-right increment, with correct styles and logic applied. The `$ 10` button will remain the default selection.

I am going to mark this as "Needs Design" since that is not my area to make a final decision in.

maxrimue:
@lewisgoddard 99% of the people downloading it are not paying, this doesn't mean that everyone who wants to download it really does it after all.
Having read the post you linked, I can only agree that probably more people donate when you make it easier to donate, or highlight the option of donating. However, only those people who want to donate will donate anyway, others will type "0" or leave, and since this PR doesn't remove these buttons with pre-defined values, I don't think this has any impact on the donations.

lewisgoddard:
As I said, final decision will go to the design team. Here's a screenshot for reference on how the page looks on load.
screenshot from 2015-05-13 12 32 45

maxrimue:
@lewisgoddard Thanks

KeitIG:
Well, I agree if the 10$ button is selected by default.

But of course, all the debate is about "should the user write '0' to download Freya for free ?". There was long discussions about that.

maxrimue:
@KeitIG Well, I wouldn't remove the other buttons and such as well, I get the reason why it is how it is now, but adding the 0$ button, while not removing another button nor setting another one as the new default, I guess it wouldn't change the behaviour of the users that want to download elementary. Instead, seeing it's okay if you click "0$" might get more people to download it, as I said, there are maybe some that aren't downloading it because of the missing option of 0$.

I'd actually be interested in some stats on how many people click the download button with the 10$ selected, but then leave the site without downloading.

lewisgoddard:
@maxrimue Unfortunately, we're not currently collecting stats on how many people click the download button, nor whether or not they exit.

cassidyjames:
We've had lengthy discussions about this topic. In the post Lewis linked, we explicitly explain why we moved to a model of typing $0, and that still applies.

lewisgoddard:
#634 is an issue about this with some discussion from our design lead, @danrabbit, in addition to the following posts:
http://blog.elementary.io/post/110645528530/payments
http://blog.elementary.io/post/118806943356/thank-you-everyone

PR #667, opened 2015-05-15
removed the lame linear transition to slider

btkostner:
last 'to do' on pull request #405


lewisgoddard:
I don't think we want to be doubling the transition time, and I would have thought `ease-in-out` would suffice, but we can check your animation too.

danrabbit:
Yeah I'm gonna close this. It's already fixed since we use ease-in-out

btkostner:
the css code in master branch is linear right now

Issue #646, opened 2015-05-12
Redirect old answers and journal folders.


lewisgoddard:
In the Nginx config, we need to redirect `/journal*` to blog.elementary.io and `/answers*` to `/support`, so that urls like `/journal/luna-beta-1-released` and `/answers/missing-dropbox-indicator` get shown a more informative page. That's almost 90% of all 404 errors.

Note: Trailing slashes are important, they stop `/support` ending in  a redirection loop, but still fix things like `/support/answers/3366`.

### Nginx Redirects (for catch-alls)
- [ ] `/journal*` to `http://blog.elementary.io/` (replaces file)
- [ ] `/answers*` to `/support` (replaces file)
- [ ] `/user-guide*` to `/docs/learning-the-basics`
- [ ] `/docs/human-interface-guidelines/*` to `/docs/human-interface-guidelines`
- [ ] `/download*` to `/`
- [ ] `/payment*` to `/`
- [ ] `/support/*` to `/support`

### File Redirects (for individual pages)
Only existing ones.

### To Sort
- [ ] `/search*`

lewisgoddard:
Additionally, `/docs/user-guide*` should be redirected to `/docs/learning-the-basics`, so that `/docs/user-guide`, `/docs/user-guide/`, and `/docs/user-guide/thing` all work.

lewisgoddard:
I think this just stopped being bitsize.

lewisgoddard:
```
server {

    listen 80;
    listen 443 ssl spdy;
    server_name elementary.io www.elementary.io;

    root /var/www/io.elementary;
    error_log /var/www/io.elementary/logs/error.log info;
    access_log off;

    index index.html index.htm index.php;
    error_page 404 /404.php;

    ssl_certificate     /etc/nginx/ssl/cert-dotio.pem;
    ssl_certificate_key /etc/nginx/ssl/elementary-nopwd.key;

    location / {
        # for Translations
        rewrite "^/([a-z]{2}(?:_[A-Z]{2})?)/(.*)$" /$2?lang=$1 last;
        try_files $uri $uri.html $uri/ @extensionless-php;
    }

    location @extensionless-php {
        rewrite ^(.*)$ $1.php last;
    }

    location ~ \.php$ {
        include /etc/nginx/fastcgi_params;
        fastcgi_pass unix:/var/run/php5-fpm.sock;
        fastcgi_index index.php;
[...]

PR #672, opened 2015-05-16
Create generate_204.php

lewisgoddard:
http://beta.elementary.io/branch/generate_204/backend/generate_204
http://elementary.io/backend/generate_204

### DIFF
--- a/backend/generate_204.php
+++ b/backend/generate_204.php
@@ -1,0 +1,2 @@
+<?php
+header($_SERVER['SERVER_PROTOCOL'].' 204 No Content');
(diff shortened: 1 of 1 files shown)


gamerlv:
It would be more efficient to do this in the nginx config. That would save a php thread for the actual site.

lewisgoddard:
It would, but I don't currently have that access. This would be quicker to implement, short term. I suppose this could be tied in with #646

gamerlv:
Yeah, it's possible to add this when I/we update the config for #646 
All that is needed for nginx to do this is:

```
location /capnet/204 {
               keepalive_timeout 0;
                return 204;
 }
```

PR #682, opened 2015-05-26
Fix #681

### LINKED ISSUE #681: "Meet our apps" awkward on big displays
screenshot from 2015-05-26 11 51 20

We should find a way to keep it more centered.

lewisgoddard:
Fix #681 https://beta.elementary.io/branch/fix-681/


lewisgoddard:
Breaks small screens.

lewisgoddard:
I am happy that this works on both small and large screens, with the only issue being pre-existing (background doesn't modify height to fit changes in window dimensions).

Issue #595, opened 2015-04-24
Update hash/fragment when scrolling in docs


cassidyjames:
Right now in the docs if you click to a section, then scroll on a bit, the URL still includes the hash of the earlier section.

Would it be useful to update the hash to link to the current section?

lewisgoddard:
I assume this should only apply to docs, and not every id we scroll past.

Check out `/scripts/smooth-scrolling.js` for how to get an id quickly from an absolute path in javascript.

cassidyjames:
@lewisgoddard yes, only in docs.

PR #684, opened 2015-06-03
Hash update

btkostner:
added hash updating url and sidebar navigation to the docs

### DIFF
--- a/scripts/docs/main.js
+++ b/scripts/docs/main.js
@@ -61,8 +61,8 @@ $(document).ready(function() {
         // Saves and removes id from element
         var id = currentSection.id;
         currentSection.id = '';
-        // Changes browser hash
-        location.hash = id;
+        // Changes browser hash without adding to history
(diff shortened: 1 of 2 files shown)


btkostner:
fixes bug #595

lewisgoddard:
This is going to require some serious testing, and I think emersion wrote most of that code.

emersion:
Okay, will review.

lewisgoddard:
I've only got a couple of minor problems, mostly the issues it exposes with our consistency, or rather lack of consistency. `/docs/installation` only has 1 header that is detected, since the rest were not properly formatted.

lewisgoddard:
Since I can't edit your repository, I've proposed a branch at #706 that fixes the issues with `/docs/installation`

If you are happy with those modifications and merge them here, I will merge this branch (unless someone else merges mine to master first, although we seem to be slacking with that).

PR #697, opened 2015-06-14
Add download classification for new servers.

lewisgoddard:
### [beta.elementary.io/branch/sorry-sourceforge](http://beta.elementary.io/branch/sorry-sourceforge/)
Will render links with the new format, direct download servers and webseed-less Magnet links.

### [/backend/classify.get.php](http://beta.elementary.io/branch/sorry-sourceforge/backend/classify.get.php)
Renders a debug version of your location lookup, giving the hostname (typically an IP), followed by a Continent, Country, and if not already allocated, a Hash and Remainder. This will be followed by your allocated server.  
`109.149.93.184 Continent: "EU" Country: "GB" Longitude: "-0.13" lon1`


lewisgoddard:
I've now got 7 confirmations of reasonable results, from just two continents though, as far as I can tell.

lewisgoddard:
Hang on, I'm going to add tracking.

lewisgoddard:
Okay, relative-reference, my bad. The Cloudflare caching the actual _page_ didn't help though.

lewisgoddard:
_Links will not attempt to open in a new tab once in `master`, the `scripts/external-links.js` is interfering.

backend/classify.get.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+include_once 'classify.functions.php';
+
+if (!empty($_GET['IP'])) {
```
fabianthoma: This looks like debug code, if it is, can we remove it?

backend/classify.functions.php
```diff
@@ -0,0 +1,124 @@
...
+		return 'sgp1';
+
+	// Other
+	} else {
+		return array('nyc3', 'ams3');
+	}
+
+}
+
+function ipHash($hostname, $debug = false) {
```
fabianthoma: What does this function do? Please add comments and explain it a bit, because I have no clue frm reading this code

backend/classify.get.php
lewisgoddard: This entire _file_ is a debug code, but my concern is that someone will complain they are downloading from somewhere weird and I will need a place to test to send them so I know what Continent/Country/Timezone they are getting.

fabianthoma: ok, we can leave that part in, might be a good idea to mark it as that then though

PR #702, opened 2015-06-20
Adding vagrant environment

javaguirre:
Hi!

I've just created a simple [Vagrant](https://www.vagrantup.com/) environment to isolate the development of the project. I selected the `ubuntu\trusty64` box for It, but we should change It to the correct production server OS (I don't know which is It).

Thank you!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 # Get rid of silly temp files
 *~
 backend/config.php
+.vagrant/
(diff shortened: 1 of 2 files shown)


lewisgoddard:
What version of PHP does this run with?

javaguirre:
The same as in the Elementary production server.

screenshot from 2015-06-21 16 52 35

In the vagrant machine:

screenshot from 2015-06-21 16 54 27

lewisgoddard:
I guess that's for the best. Ideally, I'd like both upgraded to 5.6, but I think that's a separate bug.

javaguirre:
I agree with you.

PR #707, opened 2015-06-23
Update code style

danrabbit:
Some adjustments based on Corentin's recommendation.

* Add rule for case statement
* update GPL header
* Change rule for braces on single lines


tintou:
Looks nice, the changes are a big step forward in code readability!

PR #715, opened 2015-06-25
Traslations checker

lewisgoddard:
https://github.com/elementary/mvp/blob/traslations-checker/backend/translations-checker.php


lewisgoddard:
I appear to have accidentally branched from #714 instead of master, so long as that is merged first it shouldn't be an issue. https://github.com/elementary/mvp/blob/traslations-checker/backend/translations-checker.php

PR #720, opened 2015-06-26
Fix 404 link to HIG

quassy:
Link to `/docs/human-interface-guidelines` instead of `/docs/human-interface de guidelines`


quassy:
Reading a bit more about translation this probably has to be fixed in Transifex. As I don't get Transifex someone else should do that ;)

Issue #732, opened 2015-07-10
Geary support link points to Ubuntu package


embik:
As [this question on Launchpad](https://answers.launchpad.net/elementaryos/+question/269074) points out, the Geary link on [elementary.io/en/support](https://elementary.io/en/support) forwards users to the [Ubuntu package questions](https://answers.launchpad.net/ubuntu/+source/geary). The Ubuntu community won't support these questions though (for good reason, we're a different distro after all), so we either need to pull this link (for now) or find a better link for Geary support (shame we don't have support forums :stuck_out_tongue: )

danrabbit:
As stackexchange is currently still private, we'll probably just need to pull this link for now since there isn't a geary-specific link for support other than Launchpad Answers

danrabbit:
StackExchange is now in public beta. We could change the links to there now? http://elementaryos.stackexchange.com/questions/tagged/geary

PR #736, opened 2015-07-17
Changed links in support page to StackExchange

shockry:
This was intended as a fix for the bitesize issue #732

### DIFF
--- a/support.php
+++ b/support.php
@@ -10,52 +10,52 @@
 </div>
 
 <div class="row apps">
-    <a class="app" href="https://answers.launchpad.net/maya" target="_blank">
+    <a class="app" href="http://elementaryos.stackexchange.com/questions/tagged/maya" target="_blank">
         <img width="64" height="64" src="images/icons/office-calendar.svg" />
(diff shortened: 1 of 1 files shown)


danrabbit:
Some of these are incorrect. The names of the tags on StackExchange aren't exactly what the LP project names are.

snap-elementary -> snap
pantheon-files -> files
pantheon-photos -> photos
switchboard -> settings
audience -> videos

It's worth noting that currently some of these tags have no associated questions

fabianthoma:
Something I also noticed was that you are linking to `http`, how about linking to `https` by default?

shockry:
My bad, it's my first contribution ever, now for me to be clear, the tags with no questions are ok to the change, right?

lewisgoddard:
I think we'd be okay to linking to empty space for now, seeing as they will likely be filled later.

fabianthoma:
@shockry Looking good to me, this is a great first pull request. Also no worries, we double check everyones work, and most of the time there are improvements to be made :wink:

PR #737, opened 2015-07-18
UPDATE: code-of-conduct lang:de

njchw:
Bitte diese Aktualisierung in die Website aufnehmen. Wer mag kann sie noch fertig stellen
Please upload this update. Its not done, but the original translation was full of mistakes...
----------------------------------
===ENGLISH===
Translation errors corrected. I'll begin with translation of the parts they're still complete in English later. Got the original translated with Google Translator? 

===GERMAN===
Übersetzungsfehler ausgebessert. An die Übersetzung der Stellen, die noch komplett in Englisch sind mache ich mich später. Wurde das mit Google übersetzer übersetzt??


emersion:
Thanks for fixing these translations issues! Sorry, but we don't accept translation updates via pull requests, because we're managing translations with Transifex. You can instead check if translations are correct on Transifex and submit new ones if needed (you can even import your file `lang/de/code-of-conduct.json`).

Please read [`TRANSLATE.md`](https://github.com/elementary/mvp/blob/master/TRANSLATE.md) for more information.

njchw:
Yes, i've fixed the issues on Transifex, too!

emersion:
Then translations will be merged in #739 ;-)

We try not to duplicate languages: here German has already been added and we don't want to have two German translations of the website which are basically the same. If there are major differences between German and German (Austria), let me know and I'll accept it.

njchw:
There are actually no differents. In Austria (Where I live) the people are speaking an other dialect, but the "offical" language is the same.

njchw:
Could you tell me why https://elementary.io/de/code-of-conduct is not complete translated? at Transifex exists a complete Translation, why is it not online?

emersion:
It is.

njchw:
Wow! Tanks a lot!

fabianthoma:
@njchw yes, we update the translations from transifex periodically

njchw:
@fabianthoma The translations with the most votes or the newest?

lewisgoddard:
Whichever it defaults to, most likely.

lewisgoddard:
The issue is those translations get over-written when we push new code, which leaves us with the problem that we are periodically rolling back translations unintentionally.

I wonder if we could get a bot to file a pull request with new translations weekly?

njchw:
Ok... I think, now is elementary.io complete translated in german, isn't it?

PR #740, opened 2015-07-18
Geoip2 is no longer a dependency

### LINKED ISSUE #716: GeoIP triggers fatal error
Cannot run website when GeoIP2 is not installed.

```
$ php backend/extract-l10n.php index 
PHP Notice:  Undefined offset: 1 in /home/simon/public_html/mvp/backend/function.branchname.php on line 6
PHP Notice:  Undefined index: REQUEST_URI in /home/simon/public_html/mvp/_templates/sitewide.php on line 23
PHP Notice:  Undefined index: REMOTE_ADDR in /home/simon/public_html/mvp/backend/classify.current.php on line 12
PHP Fatal error:  Class 'GeoIp2\Database\Reader' not found in /home/simon/public_html/mvp/backend/classify.functions.php on line 28
```

emersion:
Fixes #716.

### DIFF
--- a/backend/classify.functions.php
+++ b/backend/classify.functions.php
@@ -9,7 +9,7 @@
 // OC   Oceania          SGP1
 // AN   Antarctica       SGP1
 
-include 'geoip2.phar';
+require 'geoip2.phar';
 use GeoIp2\Database\Reader;
 
 ////    ipCheck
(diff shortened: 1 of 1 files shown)


lewisgoddard:
Maybe we need to catch that exception in https://github.com/elementary/mvp/blob/geoip-fix/backend/classify.current.php, but I still think we should just list `php5-geoip` as a requirement.

lewisgoddard:
I've had to revert this due to a weird error. You need to make a new pull request with the full reference. `GeoIp2\Database\Reader`, not just `Reader`. (Also, your indenting is in tabs.)

> If you are using aliasing to import namespaced classes, take care that class_exists will not work using the short, aliased class name - apparently whenever a class name is used as string, only the full-namespace version can be used
```php
use a\namespaced\classname as coolclass;
class_exists( 'coolclass' ) => false
```

PR #743, opened 2015-07-18
Fixes pull request #740 - geoip fix

emersion:


lewisgoddard:
Oddly, I can't get this to fail even without GeoIP installed. I'm not entirely sure that it does produce a fatal error when `php5-geoip` is not installed, rather, it may have been caused by my original branch not containing the `phar`.

emersion:
Just tested, and it works well:

screenshot from 2015-07-18 21-26-05

Issue #647, opened 2015-05-12
Resize the Notebook Hero image based on media


lewisgoddard:
Google complains about (82% reduction) notebook.png, maybe have it so it scales from different files?

cassidyjames:
Are you sure it's the notebook hero and not the wallpaper?

cassidyjames:
I'm getting tiny compression on either one, only a 4% reduction.

cassidyjames:
(and after the 4% reduction, it looks like poo)

cassidyjames:
@lewisgoddard is drunk, Google said resizing it could reduce filesize. They're already compressed as much as possible without losing quality. Editing the title. :)

PR #744, opened 2015-07-18
Fix for issue #647

sailormansam:
Used CSS and background images to serve up image sizes that would best
fit different screen sizes. Created three new images of varying sizes.


styles/main.css
```diff
@@ -146,11 +146,33 @@ img.h1 {
...
+* Responsive Hero *
+******************/
+.hero {
+    padding-top: 0;
+    height: 176px;
+    background: url(../images/responsive-images/notebook-small.png) 50% /contain no-repeat;
+}
+
+@media all and (min-width: 530px) {
+    .hero {
```
lewisgoddard: How did you select these breakpoints? Are they used elsewhere?

sailormansam: I looked at the smallest size the current notebook.png scales to when you go down to a screen width of 320px. I doubled that image size approximately, and created notebook-small.png to accomodate 2x pixel density screens. I picked that first break point (min-width: 530px) where notebook-small.png would no longer fit the needs of a 2x screen. From there I made two more breakpoints and images that I thought would be a good fit for different screen sizes. I didn't check to see if those breakpoints were used elsewhere.

lewisgoddard: I think things are so inconsistent with breakpoints right now that it's a separate but related issue.

sailormansam: Agreed. I'd be interested in cleaning them up and consolidating them.

sailormansam:
Another option we have instead of CSS is to use a small js plugin https://github.com/kvendrik/responsive-images.js. This would allow for more semantic markup by displaying the responsive images in the html rather than css. This would be more maintainable for future implementations of responsive images while we wait for the browser support of `srcset` to rise.

lewisgoddard:
I don't think we want extra scripts right now here they can be avoided, not if we can do this with CSS. I will attempt to test this tonight.

lewisgoddard:
Fixes #647

sailormansam:
elementary.io It looks like the new css isn't on production.

lewisgoddard:
Whoops, we're working on clearing the cache on CloudFlare every time.

PR #745, opened 2015-07-20
Update jQuery.leanModal2.js

fabianthoma:
No idea why the download modal broke, but updating the jsquery lib to the version @lewisgoddard has in his github repo works.


lewisgoddard:
I did say at the time that moving away from an official release could cause problems.

PR #753, opened 2015-07-23
Responsive scaling of images in the "Meet Our Apps" slider in Firefox

shockry:
I've found that this style rule makes the images in the slider scale down (and up) to the viewport in Firefox, but I couldn't come up with a better selector


lewisgoddard:
I'm unclear on exactly what this branch is trying to accomplish. Are you trying to force images in the carousel to go larger than their natural size to fill the display? I'm not sure that's desirable, or that the display ever actually goes that large. The carousel itself already has a maximum width, even though the background is always full screen.

shockry:
It's just that images inside the carousel take their width (that would be 100% width of the "column" div it's inside) when the viewport changes size in Chrome but not in Firefox, it may be that it ignores the max-width rule set to the "img" tags so images are bigger than the viewport when the width is relatively small, so the 100% width rule was intended to make the image fit its intended space.

lewisgoddard:
I think a better option would be the following, as this pull request puts it, images will become blurry at larger sizes, and it will interfere with other browsers. For some reason, this `max-width` is set to `initial` at the moment.

```
#carousel .slide.row {
    max-width: 100%;
}
```

shockry:
Yes I think this would work better, I thought that if the image and its container have a max-width that would avoid the problem of the image being too large, that's filling what has a max-width already, what did I get wrong? if you could explain to me as I'm fairly a beginner in this

lewisgoddard:
The problem is that setting `width: 100%` on an image will stretch it when the container gets too large, that's why we use a global `max-width` instead. I will check this for side-effects this evening, then I think it's okay to merge.

PR #759, opened 2015-07-25
Adds nice 403 template

### LINKED ISSUE #724: There is no nice 403 page.
http://elementary.io/_templates

isdampe:
Adds a nicer 403 page, fixes #724.

### DIFF
--- a/403.php
+++ b/403.php
@@ -14,7 +14,7 @@
         <i class="warning fa fa-warning"></i>
     </div>
     <div class="column alert">
-        <h3>Access Denied.</h3>
+        <h3>Sorry, access is denied.</h3>
         <p>You don't have permission to view the requested resource.</p>
     </div>
     <div class="row">
--- a/403.php
+++ b/403.php
@@ -1,0 +1,27 @@
+<?php
+    include '_templates/sitewide.php';
+    $page['name'] = '403';
+    $page['title'] = 'Forbidden &sdot; elementary';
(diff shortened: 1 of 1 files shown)


403.php
```diff
@@ -0,0 +1,27 @@
+<?php
+    include '_templates/sitewide.php';
+    $page['name'] = '403';
+    $page['title'] = 'Forbidden &sdot; elementary';
```
lewisgoddard: I think the page title should probably be "Access Denied", to match how the others are formatted (in a short but readable way).

403.php
```diff
@@ -0,0 +1,27 @@
...
+    ga('send', 'event', '403: Forbidden', window.location.host);
+</script>
+
+<div class="row">
+    <div class="column alert">
+        <i class="warning fa fa-warning"></i>
+    </div>
+    <div class="column alert">
+        <h3>Access denied.</h3>
+        <p>You don't have permission to view the requested resource.</p>
```
lewisgoddard: Most of our in-page titles are longer and more friendly than this. Maybe change the paragraph to have a "Sorry, ..." at the beginning, since it's probably not that users fault.

403.php
isdampe: Agreed. Will fix.

403.php
isdampe: Should the small ```<p>``` also be nicer?

PR #762, opened 2015-07-25
Add Hashes and Instructions to Documentation (Fix #502)

### LINKED ISSUE #502: Add hash sums to Installation page
Several people have asked for verified hash sums of the ISO, which is a good idea. Perhaps the Installation page would be a good place for this?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/11008793-add-hash-sums-to-installation-page?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github).
</bountysource-plugin>

lewisgoddard:
- Add hash table to /docs/installation


lewisgoddard:
I think we need to pick which kind of checksum we want to provide instructions for, likely whichever is easiest across all 3 platforms, and just give instructions for that, otherwise we'll have 18 sets of instructions (2 architectures x 3 systems x 3 hashes).

lewisgoddard:
### Linux
`md5sum filename.iso`
`sha1sum filename.iso`
`sha256sum filename.iso`

aleksandar-todorovic:
Checking both MD5 and SHA1 checksums seems straightforward on Mac.

Now Windows is a problem. They seem to offer one tool (https://support.microsoft.com/en-us/kb/841290) which can check both MD5 and SHA1, but I'm not sure does this tool work with Windows 7 and older systems

My suggestion is to ditch the SHA256 and keep MD5 and SHA1.

lewisgoddard:
That is a _really_ complex way to check checksums.

I think ideally we'll keep all the checksums on the page, but only provide instructions for the easiest ones in each OS. Surely Windows has better systems that that?

aleksandar-todorovic:
I found an easier solution.

It's a PowerShell command and it basically goes like this:

`Get-FileHash <filepath> -Algorithm MD5`

I don't think there's a way to compute them using `cmd` without installing anything.

lewisgoddard:
I can take a look at Windows again tomorrow, but OS X is going to be an issue.

Shnatsel:
SHA256 is preferred, SHA1 as a last ditch. No MD5 please.

In an ideal world we'd just plug WebTorrent in there that would handle all the hashes behind the scenes and also show a nice progressbar and estimated completion time.

lewisgoddard:
screenshot from 2015-08-21 18 36 32

lewisgoddard:
All I need now are some instructions for sha256 on Windows. Hint: It's not native.

lewisgoddard:
screenshot from 2015-08-21 18 46 12

lewisgoddard:
My only issue is, these vertical scrollbars don't seem to work, but I'm assuming that's an outstanding issue. I also don't think most people will verify more than the first few digits.

screenshot from 2015-08-21 19 25 13

danrabbit:
Cassidy has banned "Unfortunately" last I checked :p

"All you have to do is" seems kind of unnecessary. Might be easier to digest as steps:
1. Open DivHasher
2. Select "Choose.." and find the elementary OS ISO file
3. Verify that the text box next to "SHA-256" matches one of the following hashes

danrabbit:
Yay now we just need to fix the padding on the ol. It looks like it thinks it's just below the image and not beside it, so the padding isn't being applied to indent it.

danrabbit:
also it's broken on mobile. The image is wider than the display

danrabbit:
Looks like mobile is already borked due to other reasons. So I don't think we should hold this up over it. We'll need to address that in another branch.

PR #767, opened 2015-07-26
Fix horizontal scrolling on homepage.

SimplyCorey:
Disables horizontal scrolling on mobile.


lewisgoddard:
I'm fairly certain this isn't how we want to fix this. We had a long discussion before about making things zoomable for people who have vision problems.

PR #769, opened 2015-07-27
Changing "label.label" to "button.label"

aleksandar-todorovic:
This seemed suspicious to me, so I tried it and it looks like I was right.

`label.label` doesn't get the desired result, `button.label` does.


lewisgoddard:
ricotz approves

PR #775, opened 2015-07-30
Fix #768

### LINKED ISSUE #768: Add sizing info to store page
This should be in smallish type on the bottom of the store page.
> All apparel is in US sizes. American Apparel sizes run a bit snug; for sizing details, see [this chart](http://www.americanapparel.net/sizing/default.asp?chart=mu.shirts).

lewisgoddard:
Test at http://beta.elementary.io/branch/origin/fix-768/store

### DIFF
--- a/store.php
+++ b/store.php
@@ -12,7 +12,7 @@
                 <script src="https://cdn.sq-api.com/market/embed.js" charset="utf-8"></script>
             </div>
             <div class="row">
-                <p>All apparel is in US sizes. American Apparel sizes run a bit snug; for sizing details, see <a href="http://www.americanapparel.net/sizing/default.asp?chart=mu.shirts"></a>this chart</a>.</p>
+                <p>All apparel is in US sizes. Tees run a bit snug; for sizing details, see American Apparel&rsquo;s <a href="http://www.americanapparel.net/sizing/default.asp?chart=mu.shirts">sizing chart</a>.</p>
             </div>
 
 <?php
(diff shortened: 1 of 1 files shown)


cassidyjames:
Good call. How about

```html
All apparel is in US sizes. Tees run a bit snug; for sizing details, see American Apparel&rsquo;s <a href="http://www.americanapparel.net/sizing/default.asp?chart=mu.shirts">sizing chart</a>.
```

lewisgoddard:
I'm okay with that, but I've removed the `target="_blank"` since we have a script that handles that for us.

lewisgoddard:
Do you think we should make it a `h4`, to match the top tagline. It seems to go with the "Our store currently only supports US orders.", they're similar things. 
screenshot from 2015-07-30 22 40 36

cassidyjames:
> I've removed the target="_blank" since we have a script that handles that for us.

Eh, I'm indifferent about that. I still think it's good practice to keep the target if we know we want it popping into its own tab (and it seems weird to _rely_ on js to do that rather than keeping js as a safeguard against accidentally opening links in the same tab). I am also unsure if keeping the target in the tag affects spiders/crawlers since they won't be hitting the js and will just see the markup as-is.

cassidyjames:
> Do you think we should make it a h4

Sure, looks good. As long as it's still below the store frame. :)

lewisgoddard:
Actually, most spiders run JS now, so we shouldn't have any problems there. As for relying on JavaScript, I think most of our site breaks without it. You certainly can't download elementary.

cassidyjames:
Looks good. :+1:

PR #776, opened 2015-07-30
Fix #773: Remove SourceForge reference.

### LINKED ISSUE #773: We're still getting the ISO size from SourceForge.
That's not going to work come the next update, I think we will have to go back to hard-coding it.

It _isn't_ updated that regularly.

lewisgoddard:
Fixes #773: We're still getting the ISO size from SourceForge.
Test at https://beta.elementary.io/branch/origin/fix-773


lewisgoddard:
Seems like the branch is failing on something Stripe, but I didn't touch that.

ping @fabianthoma

index.php
```diff
@@ -54,7 +53,7 @@
                     ?>
                 </div>
                 <button type="submit" id="download" class="suggested-action">Download Freya</button>
-                <p class="small-label"><?php echo round($filesize / 1000 / 1000); ?> MB (for PC or Mac)</p>
+                <p class="small-label">893 MB (for PC or Mac)</p>
```
cassidyjames: This is a workaround that works for now, but if we're going to want updated ISOs, we'll have to manually keep updating this page. Are we cool with that?

lewisgoddard: @cassidyjames We have to manually update the links anyway, so I don't see this as a major problem.

PR #777, opened 2015-07-31
371 fix custom amount

### LINKED ISSUE #371: Custom amount input looks different (and wrong) with different browsers
Generally the rendering seems different from browser to browser: The dollar signs are different in size as (for Firefox and Internet Explorer) is the surrounding box. Only in Chrome the text is displayed in grey.

Only for the German translation: The text overflows the custom amount textbox. "Benutzerdefiniert" translates to "Custom" (literally "User defined"). Unfortunately I cannot think of a better, shorter translation, so maybe the box should just be made bigger for some locales.

Firefox 36
Chrome 40
Internet Explorer 11

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/8415000-custom-amount-input-looks-different-and-wrong-with-different-browsers?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Andertech:
This should fix #371 for now. I think css could get more tidy if we introduced BEM or at least re-style some of the components. Also this is my first PR for elementary, I hope its useful!

### DIFF
--- a/index.php
+++ b/index.php
@@ -44,8 +44,8 @@
                     <button id="amount-ten"         value="10" class="small-button payment-button target-amount checked">10</button>
                     <button id="amount-twenty-five" value="25" class="small-button payment-button target-amount">25</button>
                     <div class="column">
-                        <sup class="pre-amount">$</sup>
                         <input type="number" step="0.01" min="0" max="999999.99" id="amount-custom" class="button small-button target-amount" placeholder="Custom">
+                        <sup class="pre-amount">$</sup>
                         <p class="small-label focus-reveal text-center">Enter any dollar amount.</p>
                     </div>
                     <div style="clear:both;"></div>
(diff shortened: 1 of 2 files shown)


danrabbit:
Sorry for the long wait time on the review!

I like your solution of only showing the $ while we're inputting and having the button just have text. But there's a problem still in that now the $ will move to try to match text that's not there. We should probably just put the $ always in the corner now so it never overlaps the text

Andertech:
No problem, I'll see if I can fix it and ping you back then :+1:

danrabbit:
Hey sorry, I'm going to go ahead and close this. Lewis went ahead and fixed it already :)

PR #781, opened 2015-08-02
Improved mac docs

linusbobcat:
This includes more accurate instructions regarding making install disks
for Macs. It includes instructions on how to burn DVDs as well because
older Macs are incapable of booting elementary from USB. The only issue
is that the long Mac documentation causes the Windows and Linux tabs to
have lots of whitespace. I don't have enough knowledge of Javascript or PHP to fix that. I apologize if I'm not following the contributing guidelines --- I don't really know how Github works.


docs/installation.md
```diff
@@ -59,17 +60,79 @@ Click "Start", then just wait for the process to finish.
...
+ - iMac 13,1 (Late 2012) or newer
+ - Mac Pro5,1 (Mid 2010) or newer
+
+For these Macs, proceed to the instructions on creating an install USB drive. If your Mac is older than the ones shown above, proceed to the instructions on creating an install DVD.
+
+#### Creating an Install USB Drive
+
+Begin with a spare USB flash drive with at least 1 GB of free space.
+
+You'll need a small application called "elementary thumbdrive creator" Download it [here](https://github.com/sdaitzman/elementary-thumbdrive-creator/archive/master.zip), uncompress it, and open the app to launch it. You'll see a window looking like the one below.
```
aleksandar-todorovic: I don't think that this tool should be a recommended way to create a bootable drive.

lewisgoddard: Do you have any specifically against it? We recommend lots of third party projects for this.

linusbobcat: I chose that project, https://github.com/sdaitzman/elementary-thumbdrive-creator/ for the following reasons:

- It uses Automator, which bypasses the OS X Gatekeeper and should be able to run on older versions of OS X
- It uses dd, which, in my experience, is the most reliable way to create a Linux install disk on OS X. Unetbootin never worked reliably for me. 
- It gives clear instructions to the user.

If it's the copywriting (UI text) or name that's an issue, I can fork it and change it. Why do you disagree with recommending it and what do you suggest instead?

danrabbit:
I'm onboard with adding Mac-specific instructions for a USB, but this is way too much. I don't think we need to worry about people installing elementary OS on a 10+ year old Mac (yep it's been that long since they switched to Intel) and Apple themselves don't support creating an install DVD anymore. I think we can/should assume USB across the board and not speak to legacy devices.

linusbobcat:
I can remove some of the super excessive information about PPC Macs and such, looking back, it is a little ridiculous.

At the same time though, pretty much only Macs released from 2012 onwards can boot elementary OS from a USB; those are macs from just the past *three* years ago. It gets more confusing that the different Mac models gained proper EFI USB booting support at different times. I addressed the lather part in my copy. I did a lot of research into this, and there's no way around that.

I don't see why we should remove support for every Mac released before 2012 just because the text is a little too long on the website.

danrabbit:
I definitely think calling out specific models and the like is excessive. Opening about this Mac and everything. It's overly complicated and frankly kind of scary. It makes installation sound really hard. From this documentation I would assume it's not worth it to try. It's too much.

We don't provided DVD instructions for PC either. USB is our recommendation.

linusbobcat:
I'll come around to remove the information and make the changes once I have the time to do so. 

I understand your reasoning on user-friendliness regarding documentation, but I hope you will consider accessibility as well. 

Would be alright to suggest, in one sentence or so, that older Macs can be booted from a DVD? USB might be the recommendation, but it will *not* work on pre-2010 or pre-2012 Macs, point blank. I don't think we should leave out an enormous chunk of Mac users.

Are you also happy with recommending the [elementary thumbdrive creator](https://github.com/sdaitzman/elementary-thumbdrive-creator/) for creating USB drives on OS X. If you think the UI text of that app is off, I can fork it and fix it.

linusbobcat:
Could I recommend my own solution for creating elementary OS USBs? 
https://github.com/linusbobcat/create-elementary-os-installer

I'm still new to Github and might have messed up my branch/commits. It's suppose to contain the new text I wrote in the new documentation format.

I removed redundant text but added two lines for suggesting DVDs for older Macs. I hope this is alright.

If my branch is too messed up, I can create a new one.

lewisgoddard:
You need to do `git fetch origin` and `git rebase master` on your branch, that should shorten the diff down to be a little more manageable, but you might need to manually accept some changes yourself. Always use the one from `master` unless it's something you are trying to change.

lewisgoddard:
@linusbobcat If you don't rebase this soon then I may close and steal this, your GitHub hasn't seen any activity since the last post and this potentially useful information is languishing here.

linusbobcat:
I've been extremely busy the past few months, and I apologize for not putting any work into this. Feel free to steal my content. The latest commit I did might not have the latest changes, so I'll put it up on a GitHub gist sometime in the next few days. 

To clarify, these are the additions I made, the reasons I added them, and some background info:

1. I *suggested* that elementary can be booted using a DVD on older Macs. While Daniel said not to include full instructions on how to make a DVD, I replaced it with one sentence or so. I don't see why we should cut support for millions of Macs which are perfectly capable of running elementary. The addition of adding this line is of course up to you guys.

FYI, Macs' USB EFI boot (for platforms other that OS X) support comes around 2010-2012 depending on the type of Mac. (Apparently MacBook Pros from 2010 can do USB EFI boot whereas iMacs only support it from 2012 up.)

2. Instead of referring to Unetbootin, which has been largely unreliable in my experience, I refer to an app made from myself which has been largely based on one suggested by aroman in his elementary on Mac guide. It's build using AppleScript, which makes it compatible with many older versions of OS X, and uses DD as a backend, which is super reliable.  

Feel free to steal this and make it officially part of elementary. The license is a bit awkward, it's under a CC, but that was the license used by the original author whom I  based the project on. I think it's possible to transfer the CC to a GPL license now, but I'm not sure. I'll check if it is and transfer the license if I can. The project also contains some screenshots for you to use, although those might also be a bit out of date. I'll update the screenshots as soon as possible along with sending that Github gist. There might also be a need for additional instructions on how to bypass gatekeeper which I'll add to the gist.

3. I recommended installing firmware updates before installing elementary. OS X updates (of OS X itself) are also recommended as Apple sometimes bundles firmware updates along with them.

linusbobcat:
[Here's the Github gist](https://gist.github.com/linusbobcat/bdd4b7b0d9ad2502e24d) with all the content. I hope this can be useful. I apologize for this inconvenient way of contributing.

PR #784, opened 2015-08-04
Adds StackExhange to footer.php

aleksandar-todorovic:
I think that the current StackExchange site is mature enough to be included in the footer since it's already in the `/support` page.

### DIFF
--- a/_templates/footer.php
+++ b/_templates/footer.php
@@ -32,6 +32,7 @@ $l10n->set_domain('layout');
             <ul>
                 <li><a href="https://twitter.com/elementary" target="_blank" data-l10n-off title="Twitter"><i class="fa fa-twitter"></i></a></li>
                 <li><a href="http://reddit.com/r/elementaryos" target="_blank" data-l10n-off title="Reddit"><i class="fa fa-reddit"></i></a></li>
+                <li><a href="http://elementaryos.stackexchange.com" target="_blank" data-l10n-off title="StackExhange"><i class="fa fa-stack-exchange"</i></a></li>
                 <li><a href="https://plus.google.com/114635553671833442612" target="_blank" data-l10n-off title="Google+"><i class="fa fa-google-plus"></i></a></li>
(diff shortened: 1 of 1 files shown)


lewisgoddard:
I'm not sure if there's any existing order, but we probably want this to be `https`.

aleksandar-todorovic:
Yeah, I wasn't sure about the order neither so I just kind of put it in a random place.

I'll fix the link to `https`, but I'll wait a while to see if I need to fix something else too as I don't want to create 50 commits on one line of code.

lewisgoddard:
Fair enough, maybe we could sneak in `https` for Reddit as well.
I've pinged the designers regarding ordering.

danrabbit:
Order is fine. I think it's pretty much random. But I agree with Lewis, let's go HTTPS :)

aleksandar-todorovic:
I switched both StackExchange and Reddit links to https. :)

aleksandar-todorovic:
Well, this happened.

screenshot from 2015-08-05 20 51 47

No idea why, it looks perfectly fine on my local machine.

_templates/footer.php
```diff
@@ -31,7 +31,8 @@
             </div>
             <ul>
                 <li><a href="https://twitter.com/elementary" target="_blank" data-l10n-off title="Twitter"><i class="fa fa-twitter"></i></a></li>
-                <li><a href="http://reddit.com/r/elementaryos" target="_blank" data-l10n-off title="Reddit"><i class="fa fa-reddit"></i></a></li>
+                <li><a href="https://reddit.com/r/elementaryos" target="_blank" data-l10n-off title="Reddit"><i class="fa fa-reddit"></i></a></li>
+                <li><a href="https://elementaryos.stackexchange.com" target="_blank" data-l10n-off title="StackExhange"><i class="fa fa-stack-exchange"</i></a></li>
```
lewisgoddard: `<i class="fa fa-stack-exchange"</i>`
`<i class="fa fa-stack-exchange"></i>`

lewisgoddard:
@aleksandar-todorovic Fixed it, can I get a review on https://github.com/elementary/mvp/pull/785 ?

PR #788, opened 2015-08-10
Add a warning to all pages when not explicitly running in production (Fixes #780)

### LINKED ISSUE #780: Put a bigass warning banner above beta subdomain
Because for some reason people are linking it and it also shows up on Google.

We should have some kind of super obnoxious warning that this site is a developer sandbox and a big link to take people back to master

gamerlv:
The warning is controlled by a ENV variable called PHPENV.
Accepted values are: dev, staging, production

This fixes #780.

### DIFF
--- a/_templates/header.php
+++ b/_templates/header.php
@@ -140,7 +140,7 @@ $l10n->begin_html_translation();
                     <i class="warning fa fa-warning"></i>
                 </div>
                 <div class="icon-text">
-                    <h2 id="back-up-your-data">Warning: Development Site</h2>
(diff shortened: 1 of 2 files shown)


lewisgoddard:
Why not use the warning box we have on http://elementary.io/docs/installation#recommended-system-specifications ?

gamerlv:
Didn't think we already had a style like that. Can rewrite the error to make it look like that.

lewisgoddard:
I apologise in advance for my poor commit titles, the web editor does not lend itself to good practices.

PR #808, opened 2015-08-23
minor css clean

danrabbit:
* Move hero image stuff to home.css
* get rid of unused hr style
* add compatible vendor prefixes
* fix extra comma
* duplicate text-center
* 0 has no units
* Don't overwrite small button height
* Move #Download to home.css
* Use commas for things that are common

### DIFF
--- a/styles/main.css
+++ b/styles/main.css
@@ -264,10 +264,7 @@ nav > .nav-content > ul > li > .logomark {
     padding: 0;
 }
 
-nav .logomark .logomark-svg {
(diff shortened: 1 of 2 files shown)


styles/main.css
```diff
@@ -472,10 +422,6 @@ footer ul li a:focus {
     text-align: left;
 }
 
-.text-center {
```
lewisgoddard: I recall this being used in docs somewhere.

PR #823, opened 2015-08-25
Combined the slides in the installation instructions (Issue #800)

### LINKED ISSUE #800: Choose Windows, Linux, or Mac once in installation docs
Right now we have separate sliders for every section which is rather redundant. We should just have people choose once and set it all the way down.

Novanoid:
Fixed #800 by combining the installation instructions into a single slide for each operating system with a global switch at the top. It looks much clearer now in my opinion.

A couple of strings in there have been modified or added and will need to be translated.

### DIFF
--- a/scripts/docs/installation.js
+++ b/scripts/docs/installation.js
@@ -1,5 +1,5 @@
 $(function() {
-    // Parse user-agent to detect current plateform
+    // Parse user-agent to detect current platform
     function detectOS() {
         var ua = window.navigator.userAgent;
         if (ua.indexOf('Windows') >= 0) {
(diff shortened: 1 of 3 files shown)


danrabbit:
On the "Windows" tab, I don't think "Booting" should wrap around the rufus image since it's not related.

danrabbit:
Looks good to me :) Want to get a double check from Lewis before merging

scripts/docs/installation.js
```diff
@@ -15,49 +15,22 @@ $(function() {
...
+        operatingSystemSlider.slideTo('install-on-ubuntu');
     }
-
-    if (currentOs == 'osx') {
-        bootSlider.slideTo('booting-on-a-mac');
-    } else {
-        bootSlider.slideTo('booting-on-a-pc');
-    }
-
 })();
```
lewisgoddard: ` })();` should just be ` });` jQuery functions are automatically self-invoking, we don't need to do it ourselves.

Novanoid: You are right, the call was there already before I committed my changes and seems not to break anything. However since it is redundant I'll quickly commit a fix with the parentheses removed anyways.

PR #825, opened 2015-08-27
Generate new language files and update translations.

lewisgoddard:
Test easily with `php backend/translations-checker.php`


lewisgoddard:
This needs updating again.

KeitIG:
Looks absolutely good to me. Don't want to merge  it as I'm not really active these times.

Issue #828, opened 2015-08-28
Replace recommended hashing mechanism for Windows


lewisgoddard:
According to https://twitter.com/garethism/status/637166709007368192, DivHasher both prompts a security warning, and doesn't exist at the address we link to.

We should probably use something like https://raylin.wordpress.com/downloads/md5-sha-1-checksum-utility/ (note: does SHA256 too), the free version comes from http://download.cnet.com/MD5-SHA-Checksum-Utility/3000-2092_4-10911445.html

![Checksum Utility](https://raylin.files.wordpress.com/2014/01/pic1.jpg)

Updating the name, link, and screenshot should suffice, although we might want to drop the two column layout by the looks of things.

kbailey4444:
Cnet is known to [bundle adware with downloads](http://www.howtogeek.com/210265/download.com-and-others-bundle-superfish-style-https-breaking-adware/), so I don't think they should be used. From [this superuser post](http://superuser.com/questions/245775/is-there-a-built-in-checksum-utility-on-windows-7) I found [Certutil](https://technet.microsoft.com/en-us/library/Cc732443.aspx). It comes preinstalled in Windows and can be used to generate checksums.

lewisgoddard:
That doesn't provide support pre-Windows 7, and a significant number of our downloaders are XP. It also requires the user to use command-line tools on Windows, and probably type out a long and complex path. I would prefer to make this as accessible as possible.

CNet hasn't, to my knowledge, ever bundled software in without the developers consent (unlike SourceForge). If that is the case, then we are unlikely to be able to use the utility without going through them. **If they are not bundling Adware, then Download.com is a trusted domain, and should lend its reputation to us.**

**A potential solution to this could be [#Hash Tool](http://www.digitalvolcano.co.uk/hash.html) by Digital Volcano.** The tool is a simple GUI and is available direct from their site. (I have not been able to check if it is adware free, but [it appears to be](https://www.virustotal.com/en/file/afedbc9ff0a559ccde694c1de263d74f055f668a01b311f0a4630aad5402e3ce/analysis/)).

![#Hash Tool](https://cloud.githubusercontent.com/assets/831389/9618267/302aadcc-50fd-11e5-809f-fc0e18f07ffa.png)

PR #836, opened 2015-09-02
Replaces recommended hashing mechanism for Windows with Hash Tool

kbailey4444:
Closes issue #828. Changed to single column layout, with the Hash Tool screenshot above the directions.


lewisgoddard:
I found it a little weird that you recommended selecting the hash type before the file, considering the visual flow would follow left-to-right usually, but I now realise that the hash symbol is not in fact a button for that (it brings up the about box, weirdly), but that it hashes automatically.

Another thing that I noticed was it relies on .Net 2.0, which isn't enabled by default in Windows 8 or 10, possibly 7 too, but my machine did just handle it, so it shouldn't be a major issue. It's also a little annoying it's called "MD5 Hash" on the desktop shortcut, but that is outside of our control as well.

Issue #852, opened 2015-09-16
Update "Get Involved" milestone to loki-beta1


codygarver:
It's still showing Freya Update 1 with 0 bugs https://elementary.io/get-involved

PR #859, opened 2015-09-17
moved chart to loki-beta1

btkostner:
what the title says

### DIFF
--- a/get-involved.php
+++ b/get-involved.php
@@ -9,7 +9,7 @@
     include $template['alert'];
 ?>
 <section class="grid hero">
-    <h1>Check out our progress for Loki beta 1</h1>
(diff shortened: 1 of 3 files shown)


btkostner:
fixes #852

get-involved.php
```diff
@@ -9,7 +9,7 @@
     include $template['alert'];
 ?>
 <section class="grid hero">
-    <h1>Check out our progress for Freya Update 1</h1>
+    <h1>Check out our progress for Loki beta 1</h1>
```
lewisgoddard: We've previously capitalized both Freya and Beta, so it would make sense for this to be too.

`Loki Beta 1`

Issue #813, opened 2015-08-23
Link to "Learning the Basics" from installation guide


danrabbit:
We should recommend people read the getting started guide after they complete their installation

PR #860, opened 2015-09-17
fix 813

btkostner:
what the title says


btkostner:
fixes #813

lewisgoddard:
You should be able to move that outside the switching div so that it isn't replicated 3 times over.

lewisgoddard:
The side affect of that may be that it is statically below the length of the longest content, not the selected one. That, however, is an existing bug in the Slider.

btkostner:
It does some weird flow issues with the slider too
overflow

btkostner:
updated. should work after #868

PR #869, opened 2015-09-23
New hero image

snwh:
New hero/notebook image for the homepage


danrabbit:
scaled images are choppy AF

PR #871, opened 2015-09-27
Update human-interface-guidelines.md

lopter-sigyn:
Add link to associated resource


lopter-sigyn:
propose adding link to referenced resource

danrabbit:
This isn't the right place to link. Instead please add the link in the Contractor section below (where the current link points)

lopter-sigyn:
first pull amendment

Issue #693, opened 2015-06-11
Mention casting in code style guide


danrabbit:
According to a conversation that Rico had the other day with Christian:

`(widget as Gtk.Entry).max_width_chars` using `as` for casting is not a good idea
`(widget as Gtk.Entry)` can be `null`, if you know what the object is then use `((Gtk.Entry) widget).max_width_chars`. If you use `as` you need to add a `nullcheck` to do it properly.

So we should recommend:
`((Foo) Bar).method`

Not
`(Foo as Bar).method`

lewisgoddard:
Is this already in the code style and is wrong, or is it an entirely new section?

pavlicekl:
No, it's not mentioned in code style anywhere.
Deserves a new section, I suppose.

lewisgoddard:
Fixed by #876 876

PR #876, opened 2015-10-08
Casting section added to code style guide. Issue: #693.

pavlicekl:


fabianthoma:
ok to test

fabianthoma:
ok to test

lewisgoddard:
If @danrabbit is happy with the placement, then I'm happy with the code.

docs/code/reference.md
```diff
@@ -140,6 +140,17 @@ than good.
 Also worth referring that there should be no Hungarian notation, and no mix of
 any kinds of notations.
 
+## Casting {#casting}
+
+Prevent using `as` keyword when casting as it might give `null` as result, which could be easily forgotten to check.
```
lewisgoddard: From an language point of view, it might be better off said "Avoid using..." instead.

pavlicekl: I agree, "Avoid using ..." sounds better.

Issue #257, opened 2015-01-28
Show date for selected bar on "Get Involved" chart


emersion:
Show current date when we update donuts

PR #880, opened 2015-10-15
Show date for selected bar on "Get Involved" chart #257

pavlicekl:


lewisgoddard:
test this please

lewisgoddard:
This definitely works, but we might want to change how that date is formatted. Maybe even use relative times, like "3 days ago".

screenshot from 2015-10-15 19 09 21

danrabbit:
Yeah I agree that we should use a more "human readable" date format.

Regarding relative dates, I think "Today" and "Yesterday" make sense, but I'm not sure "3 Days ago" is as easily parsed as "October 12th". Then again, there's no real need to be super accurate here imo. So as long as it's a bit more "natural" than just straight numbers and slashes I think that would be good :)

pavlicekl:
I've changed the date format to more human readable form:
en - September 17, 2015
fr - 17 septembre 2015

Please consider that relative dates would require explicit localization.

lewisgoddard:
test this please

lewisgoddard:
Would it be possible to also get the ordinals? The th, nd, and rd's.

danrabbit:
I'm not sure we need the year if it's this year, to be honest

cassidyjames:
@lewisgoddard dates are written as cardinal numbers in American English.

lewisgoddard:
Are we saying just drop the year and we're good to merge? There's conflicting comments right now. I'll stick with cardinals to avoid translating ordinals.

pavlicekl:
Do we have a conclusion here?

lewisgoddard:
I think it was decided that we just wanted to drop the year.

pavlicekl:
I'm checking the possible implementation - strip year string via regexp looks as correct solution:

dateLocalized.replace(/,* \d{4}/, "")

But I can't ensure that there is a common solution covering all languages, especially the ar, bg, etc. ...

So I'd recommend to keep the year to prevent issues with localization of the date.

What do you think?

scripts/get-involved.js
```diff
@@ -36,9 +36,12 @@ $(function () {
         };
 
         for (var time in data) {
+            var dateLocalized = new Date(parseInt(time, 10) * 1000).toLocaleDateString(document.documentElement.lang, { year: 'numeric', month: 'long', day: 'numeric' });
```
lewisgoddard: Can we not just remove the year from the options on this line?

lewisgoddard:
test this please

danrabbit:
Hey @pavlicekl I just wanted to remind you that you have an unclaimed bounty since October https://www.bountysource.com/issues/8113199-show-date-for-selected-bar-on-get-involved-chart

PR #881, opened 2015-10-15
Better Mac Instructions

lewisgoddard:
This is a re-do of @linusbobcat's #781


lewisgoddard:
TODO: Clean up unnecessary images.

PR #882, opened 2015-10-15
Add Error Pages to nginx.conf, Fix #789

### LINKED ISSUE #789: Redirecting to 404.php does not work on subdomains
I tested this out on http://beta.elementary.io.

Basically, if you type http://beta.elementary.io/something-random you don't see that nice 404 page that you see when you type http://elementary.io/something-random.

Also, http://blog.elementary.io shows a 404 page that's inconsistent with http://elementary.io, but I'm not sure if Tumblr allows the customization of 404 pages. If so, then I think that this 404 page should definitely match the one set on http://elementary.io.

lewisgoddard:
Add Error Pages to nginx.conf, Fix #789

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -2,6 +2,11 @@
     access_log off;
     index index.php index.html;
 
+    # Error Pages
+    error_page 403 /403.php;
+    error_page 404 /404.php;
+    error_page 410 /410.php;
+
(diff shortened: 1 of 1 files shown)


fabianthoma:
Did add these lines to the nginx config so should work

PR #886, opened 2015-10-16
Repositions the UNetbootin screenshot

aleksandar-todorovic:
It was above the text even before #885. It should be placed here I guess.


lewisgoddard:
That looks better.

PR #888, opened 2015-10-16
merge installation and getting started docs

### LINKED ISSUE #812: Link to "Learning The Basics" from the support page
what it says on the tin. We should link the getting started guide here

### LINKED ISSUE #887: Combine Installation and Getting Started docs
To follow with "a click is a choice, a scroll is a continuation", I think we should combine these docs. They're fairly related imo and short enough that I don't think combining them would be a bad thing.

danrabbit:
fixes #887 
fixes #812


lewisgoddard:
So are we still not sure about this? I'm against merging single use and ongoing documentation, they are two distinct steps in my opinion, but I'll follow #ux if I have to.

FYI you'll need to get fabian to update the nginx.conf server-side, AFAIK it's still not automated.

PR #892, opened 2015-10-18
Update get-involved.json

thepieterdc:
Patreon is a brand and should therefore not be translated.


emersion:
Please contribute to translations from Transifex. See our translation guide : https://github.com/elementary/mvp/blob/master/TRANSLATE.md

PR #893, opened 2015-10-21
Update Chart and add error-printing when not writeable

lewisgoddard:


fabianthoma:
Alright, so a few things I got to note here:
- [ ] The chart.json should not live in the git repo
- [ ] Would be awesome if we could visualize multiple milestones in one view (Could be a new PR)
- [ ] Don't put all of the code in an else block, instead use PHP's exit functionality

lewisgoddard:
I agree it shouldn't live in the Git repo, but then we'll need something to use in development without setting up the file with permissions. As for else vs exit, I've always thought of exit as evil, but I suppose it makes more sense since this is effectively just a bash script.

Multiple milestones should definitely be a new PR.

lewisgoddard:
I'm going to close rather than rebase this, as some work has been merged from other branches.

Issue #324, opened 2015-01-31
Section buttons look like they relate to the bug graph


maffity:
On the Get Involved page, the section buttons are immediately below the bug graph. Since they share colors I thought they were somehow related.

If we want to keep the buttons the same colors, perhaps they could be in a new section with a different background color. Otherwise, I think we should change the colors on the graph or the buttons.

lewisgoddard:
This is temporarily fixed as the graph is hidden for the current release.

lewisgoddard:
The chart is back, and so is this bug.

PR #896, opened 2015-10-27
Section buttons look like they relate to the bug graph #324

pavlicekl:


fabianthoma:
ok to test

KeitIG:
@pavlicekl Check the indentation of your files.

KeitIG:
just referencing the pr to the issue. #324

lewisgoddard:
So this is how it looks on my screen. If I'm honest, I'm not sure that it's enough contrast, but it's probably about as much as the other section breaks are.

screenshot from 2015-11-08 00 03 34

pavlicekl:
Maybe adding a thin 1px line above the menu section in the same color as the quote bellow would be enough to improve the contrast.

PR #900, opened 2015-10-31
Replace "static notebook" explanation. Fixes #883

### LINKED ISSUE #883: Replace "static notebook" explanation
Although the concepts are the same, the Granite.StaticNotebook widget was deprecated in favor of Gtk.StackSwitcher. We should probably update the name here or at least make that reference to avoid confusing developers into using a deprecated widget

pavlicekl:


docs/human-interface-guidelines.md
```diff
@@ -641,11 +641,11 @@ See also: _[3 Ways to Make Checkboxes, Radio Buttons Easier to Click](http://uxm
 
 Notebooks are a type of widget that allow showing one of multiple pages in an app, also colloquially referred to as "tab bars."
 
-### Static Notebook {#static-notebook}
+### Stack Switcher (former Static Notebook) {#static-notebook}
```
lewisgoddard: This should go from `{#static-notebook}` to `{#stack-switcher}`, and check for other links to that.

btkostner:
Don't forget to change the image name too.

danrabbit:
I don't think this really solves the issue just replacing the words like that. StackSwitcher isn't a notebook. More info on the original report.

danrabbit:
No movement for a couple months here, so I'm closing this PR

PR #904, opened 2015-11-06
add to window, not this

danrabbit:
Fixes an issue pointed out by http://elementaryos.stackexchange.com/questions/2848/getting-started-with-gtk-grid-with-error


embik:
Looks good. It makes more sense now as the docs are only talking about putting everything in the main function.

PR #915, opened 2015-11-11
Translations section on the docs 

### LINKED ISSUE #905: Document making your project translatable [$50]
We should add a basic example of making a (small) set of strings translatable. This should include strings that need pluralization or that contain variables or any other kind of non-standard, but common use cases we can think of.

Philip-Scott:
Fixes #905

### DIFF
--- a/docs/code/getting-started.md
+++ b/docs/code/getting-started.md
@@ -425,15 +425,16 @@ That's a lot! You're well on your way to becoming a bonified app developer for e
 # Adding Translations {#Adding-Translations}
 Now that you've learned about CMake, the next step is to make your app able to be translated to different languages. 
 
-1. The first thing you need to know is how to convert strings in your code into translatable strings.
-
+The first thing you need to know is how to convert strings in your code into translatable strings. Here's an example:
         stdout.printf ("Not Translatable string");
         stdout.printf (_("Translatable string!"));
         
         string normal = "Another non-translatable string";
         string translated = _("Another translatable string");
 
-See the difference? We just added `_()` around the string! Well, that was easy!
+See the difference? We just added `_()` around the string! Well, that was easy! 
+
+1.- Go back to your project and make all your strings translatable by adding `_()`
 
 2. Create a directory named "po" on the root folder of your project. Then you have to add the following lines in the "CMakeLists.txt" file you created a moment ago:
 
@@ -461,6 +462,8 @@ See the difference? We just added `_()` around the string! Well, that was easy!
 
     ```bash 
     bzr add po/
+    bzr commit -m "Added translation files"
+    bzr push
     ```
 
 That's it! CMake will automatically add all the string marked with `_()` into a .pot template file, and a file for each available language where you'll place the translatable strings.
--- a/docs/code/getting-started.md
+++ b/docs/code/getting-started.md
@@ -442,7 +442,6 @@ See the difference? We just added `_()` around the string! Well, that was easy!
         add_definitions (-DGETTEXT_PACKAGE=\"${GETTEXT_PACKAGE}\")
         add_subdirectory (po)
 
-
 3. Inside of your po directory you will need to create another CMakeLists.txt file. This time, it's contents will be:
 
         include (Translations)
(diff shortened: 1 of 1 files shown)


danrabbit:
add to whitelist

docs/code/getting-started.md
```diff
@@ -422,6 +422,40 @@ Let's review what all we've learned to do:
 
 That's a lot! You're well on your way to becoming a bonified app developer for elementary OS. Give yourself a pat on the back, then take some time to play around with this example. Change the names of files and see if you can still build and install them properly. Ask another developer to branch your project from launchpad and see if it builds and installs cleanly on their computer. If so, you've just distributed your first app! When you're ready, we'll move onto the next section: Packaging.
 
+# Adding Translations {#Adding-Translations}
+Now that you've learn about cmake, the next step is to make your app able to be translated to different languages. The first thing you need to know is how to convert strings in your code into translatable strings.
+
+        stdout.printf ("Normal String");
```
danrabbit: Instead of normal, we should call this "not translatable" to make sure we're giving it a negative connotation

danrabbit:
- [ ] Can you update the strings that follow in the rest of the guide to be translatable? Specifically, the strings in the Gtk.Grid section.

docs/code/getting-started.md
```diff
@@ -422,6 +422,40 @@ Let's review what all we've learned to do:
...
+        stdout.printf (_("Translatable string!"));
+        
+        string normal = "Another normal string";
+        string translated = _(Another translatable string);
+
+See the difference? We just added _() around the string! Well, that was easy! Next you need to create a directory named "po" on the root folder of your project. Then you have to add the following lines in the "CMakeLists.txt" file you created a moment ago:
```
danrabbit: should be `_()` with the ` char when we have code or commands inline

docs/code/getting-started.md
```diff
@@ -422,6 +422,40 @@ Let's review what all we've learned to do:
...
+        )
+
+and finally, on your build directory execute the following commands:
+ 
+    ```bash
+    cmake -DCMAKE_INSTALL_PREFIX=/usr ../
+    make pot
+    ```
+
+That's it! CMake will automatically add all the string marked with _() into a .pot template file, and a file for each available language where you'll place the translatable strings.
```
danrabbit: again `_()`

docs/code/getting-started.md
```diff
@@ -422,6 +422,46 @@ Let's review what all we've learned to do:
 
 That's a lot! You're well on your way to becoming a bonified app developer for elementary OS. Give yourself a pat on the back, then take some time to play around with this example. Change the names of files and see if you can still build and install them properly. Ask another developer to branch your project from launchpad and see if it builds and installs cleanly on their computer. If so, you've just distributed your first app! When you're ready, we'll move onto the next section: Packaging.
 
+# Adding Translations {#Adding-Translations}
+Now that you've learned about CMake, the next step is to make your app able to be translated to different languages. The first thing you need to know is how to convert strings in your code into translatable strings.
```
danrabbit: Maybe add "Here's an example:" at the end here

docs/code/getting-started.md
```diff
@@ -422,6 +422,46 @@ Let's review what all we've learned to do:
...
+
+        stdout.printf ("Not Translatable string");
+        stdout.printf (_("Translatable string!"));
+        
+        string normal = "Another non-translatable string";
+        string translated = _(Another translatable string);
+
+See the difference? We just added `_()` around the string! Well, that was easy! Next you need to create a directory named "po" on the root folder of your project. Then you have to add the following lines in the "CMakeLists.txt" file you created a moment ago:
```
danrabbit: I feel like there's a disconnect here between the example above and what we're actually asking the reader to do. We might want to make it clearer that we're asking the reading to make their existing project translatable. This is where we should break between the example and step 1 which would be, "go back to your project and make all your strings translatable"

docs/code/getting-started.md
```diff
@@ -422,6 +422,49 @@ Let's review what all we've learned to do:
...
+    make pot
+    ```
+
+5. Don't forget to add this new directory and it's contents to bzr
+
+    ```bash 
+    bzr add po/
+    ```
+
+That's it! CMake will automatically add all the string marked with `_()` into a .pot template file, and a file for each available language where you'll place the translatable strings.
```
danrabbit: Let's finish up by reminding the reading to commit and push as well

PR #924, opened 2015-11-19
fix #846; Changes to CSS responsive 'danger zone'

### LINKED ISSUE #846: "Learn more" in Midori description(Meet our apps section) is badly visible
See screenshot, Windows 10, Chrome Version 45.0.2454.85 m
learn_more

verth:
Meet Our Apps responsive problem fix

* **<1350px:** Column with text defaults to display:block, this was inconsistent, ex. video stayed "inline" much longer than other slides.
* **>=1350px:** columns always stay next to each other (fixed spacings and widths are bit smaller), making sure "Learn More" is above the mountain fold
* **>=1400px:** Reset. Is same as before.

### DIFF
--- a/styles/home.css
+++ b/styles/home.css
@@ -133,8 +133,21 @@ input:focus + .focus-reveal {
     vertical-align: top;
 }
 
+#carousel .column.alert img {
+    height: 64px;
+    width: 64px;
+}
+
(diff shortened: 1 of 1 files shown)


fabianthoma:
ok to test

PR #928, opened 2015-12-03
Update index.json

Chalermkiat:
better and friendlier Thai translation


lewisgoddard:
Sorry, but we don't accept translation changes here. You can read our [translations guide](https://github.com/elementary/mvp/blob/master/TRANSLATE.md) about contributing translations on Transifex.

PR #944, opened 2015-12-07
Fixed unclosed code tag on spanish "getting started"

desiderantes:
It looks like this right now
tmp_8576-screenshot_2015-12-07-00-30-25-137411698


lewisgoddard:
Merging without checks because checks are broken and this is a 4 character change.

PR #954, opened 2015-12-10
Update Translations for 0.3.2

lewisgoddard:


KeitIG:
Merging it as soon as the first test passes.

lewisgoddard:
@KeitIG As far as I know, the first test isn't actually running, the tests have been broken for a few days. The second one actually does everything.

PR #966, opened 2016-01-03
Adds /capnet-assist

### LINKED ISSUE #814: Create a landing page for Capnet Assist [$25]
Right now it's just pointing to our home page which looks like an ad. We should make a simple "error" page it can point to that explains why this window has appeared/what capnet is/we messed up somewhere

Ideally, users should never see this, but for some reason they are.

--

[Bounty on Bountysource](https://www.bountysource.com/issues/26103693-create-a-landing-page-for-capnet-assist)

emersion:
Fixes #814 

I decided to hide header & footer with css rules to avoid hardcoding css styles, analytics, base html layout in `/capnet-assist.php`.

See https://beta.elementary.io/branch/origin/capnet-assist/capnet-assist

### DIFF
--- a/capnet-assist.php
+++ b/capnet-assist.php
@@ -1,0 +1,20 @@
+<?php
+    include '_templates/sitewide.php';
+    $page['name'] = 'capnet-assist';
+    $page['title'] = 'You\'re connected! &sdot; elementary';
+		$page['scripts'] .= '<link rel="stylesheet" type="text/css" media="all" href="styles/capnet-assist.css">';
+    include $template['header'];
+    include $template['alert'];
+?>
+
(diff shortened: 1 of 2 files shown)


lewisgoddard:
We might want a "close window" button.

danrabbit:
The window does have its own window controls so that is probably not necessary

lewisgoddard:
As far as I'm concerned, this is a design confirm. It's a little heavier than a static page would have been, loads the nav and footer but hides them, but at least it will keep in line with styles.

danrabbit:
Seems fine to me :) We can always come back and make it smaller if we need to

PR #967, opened 2016-01-03
Fix #730: Set up permalinks with redirects for app help links

### LINKED ISSUE #730: Set up permalinks with redirects for app help links [$25]
After some discussion it sounds like we want to set up some kind of permalinks (like "elementary.io/support/noise") that desktop apps can point to so we don't have to issue a package update if the help URL changes.

The problem here is that many of these apps still point to old answers. The current place to get app help is Launchpad answers. But we think the future place to get help is specific tags on StackExchange.

This is the issue in LP https://bugs.launchpad.net/scratch/+bug/1456733

--
[Bounty on Bountysource](https://www.bountysource.com/issues/23978816-set-up-permalinks-with-redirects-for-app-help-links)

emersion:
Fixes #730: http://beta.elementary.io/branch/origin/app-help/

- [x] Audience
- [x] Calendar
- [x] Camera
- [x] Files
- [x] Geary
- [x] Maya
- [x] Midori
- [x] Noise
- [x] Photos
- [x] Scratch
- [x] Snap
- [x] System Settings
- [x] Videos
- [x] Terminal


lewisgoddard:
This probably wants to be `header('Location: https://answers.launchpad.net/audience', true, 302);` so if we change it later people update.

emersion:
Added all remaining apps. A few questions:
* Audience & Videos are the same app. Which name should we use?
* I used StackExchange links for all apps, not Launchpad Answers. @danrabbit is it better?

danrabbit:
Videos is how we refer to it for users, so it's better to use Videos imo

Yes that's correct, StackExchange is where we want to point people. We should probably start to close the LP answers when links are fixed

lewisgoddard:
I would use _both_ when an app has a code name and a UI name.

danrabbit:
Missing:
- [ ] Camera
- [ ] Calendar

PR #975, opened 2016-01-06
Fix #421: Average Payments

### LINKED ISSUE #421: Display the average payment to encourage competitiveness for larger payments.
and some more statistics in realtime like HumbleBundle does?

maybe that motivates someone to pay more than the average or more than the average windows users.

but maybe that value will be very low because only 0,1% will pay. then one will think: "why should i pay if no one else does"

what do you think?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/8723136-display-the-average-payment-to-encourage-competitiveness-for-larger-payments?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github).
</bountysource-plugin>

lewisgoddard:
Fixes #421

- [x] Add JSON API for Payments
- [x] Log payments to a SQLite database
- [x] Log OS with payments

https://beta.elementary.io/branch/origin/track-payments/
https://beta.elementary.io/branch/origin/track-payments/backend/average-payments.php


lewisgoddard:
I could use [Authenticatron](https://github.com/eustasy/authenticatron), it's not really what it's built for but will give us a code for a given secret tied to a 30second period. Just embed the timecode in the page like the public stripe key and submit it with the request. It includes a verification system with changeable time allowances. We'll just check the 99th percentile or something.

lewisgoddard:
As far as I am aware, this works so long as the `data` directory is writable (one of the caveats of SQLite), and we should make our own `backend/authenticatron.secret.php` ([this page](http://labs.eustasy.org/authenticatron/examples.php) generates them, and I've included a default).

lewisgoddard:
### Test Output
```

{
    "total": {
        "0": "total",
        "OS": "total",
        "1": 310202,
        "Total": 310202,
        "2": 368,
        "Count": 368,
        "3": 842,
        "Average": 842
    },
    "windows": {
        "0": "windows",
        "OS": "windows",
        "1": 1,
        "Total": 1,
        "2": 1,
        "Count": 1,
        "3": 1,
        "Average": 1
    },
    "other": {
        "0": "other",
        "OS": "other",
        "1": 125000,
        "Total": 125000,
        "2": 94,
        "Count": 94,
        "3": 1329,
        "Average": 1329
    },
    "android": {
        "0": "android",
        "OS": "android",
        "1": 3300,
        "Total": 3300,
        "2": 3,
        "Count": 3,
        "3": 1100,
        "Average": 1100
    },
    "ios": {
        "0": "ios",
        "OS": "ios",
        "1": 0,
        "Total": 0,
        "2": 0,
        "Count": 0,
        "3": 0,
        "Average": 0
    },
    "os x": {
        "0": "os x",
        "OS": "os x",
        "1": 8000,
        "Total": 8000,
        "2": 4,
        "Count": 4,
        "3": 2000,
        "Average": 2000
    },
    "linux": {
        "0": "linux",
        "OS": "linux",
        "1": 0,
        "Total": 0,
        "2": 0,
        "Count": 0,
        "3": 0,
        "Average": 0
    }
}
```

backend/average-payments.php
```diff
@@ -0,0 +1,71 @@
...
+
+////    Error Handling
+function LastError($db) {
+    $Error = 'Error Code "'.$db->lastErrorCode().'" : '.$db->lastErrorCode();
+    var_dump($Error);
+}
+
+////    Open Database
+if ( $Processing ) {
+    if ( !is_writable($Database) ) {
```
gamerlv: It might be handy to also check if the folder where `$Database` is located is also writeable. If my memory serves, sqlite wants to put an lockfile in the same folder it's database file is

fabianthoma:
Since I'm assigned to this I guess I should be the one to review this as it needs server changes.
Sorry it took me so long to get around to it.

So there are a few questions I have:
- Why are we using sqlite, in production we should really be using something else as the database would get big and slow very soon. As well as access collisions could occur.
- Why are there so many new lines in the authenticatron source file, it makes it hard to read it in my opinion
- Why exactly do we need authentication by tokens, is that to prevent malicious counter tampering? If so, we should use it together with php sessions at the very least.
- The code apparently conflicts, so this PR needs updating

fabianthoma:
retest this please

lewisgoddard:
- _Why are we using sqlite, in production we should really be using something else as the database would get big and slow very soon. As well as access collisions could occur._

The database is a fixed size, since it doesn't store actual transactions, just the total and tally data required. It will never be more than a Kb or two in size. This should also reduce collisions to effectively zero.

- _Why are there so many new lines in the authenticatron source file, it makes it hard to read it in my opinion_

Authenticatron is an upstream library, so it's not something we need to worry about maintaining.

- _Why exactly do we need authentication by tokens, is that to prevent malicious counter tampering? If so, we should use it together with php sessions at the very least._

We only need to set the token for a transaction we wish to limit the frequency of, I'm unclear on the benefit of PHP sessions here.

- _The code apparently conflicts, so this PR needs updating_

I will sort this out a little later today.

lewisgoddard:
@btkostner Fetching count and average will still allow end users to get total amounts.

btkostner:
You are right. It shouldn't include count.

lewisgoddard:
I think we're actually done with this now, since it got a whole lot better. Over to @fabianthoma, we need an empty database made before merge. The create command is in the new code.

lewisgoddard:
`rsync: failed to set times on "/var/www/io.elementary.beta/branch/origin/track-payments/data": Operation not permitted (1)`

lewisgoddard:
This is still only failing because of file permissions in jenkins.

lewisgoddard:
I am unclear on why https://beta.elementary.io/branch/origin/track-payments/backend/average-payments.php fails to load when it works on my machine.

lewisgoddard:
```
/home/travis/build/elementary/mvp/scripts/homepage.js
   75:13  error  'stripeLanguages' is defined but never used             no-unused-vars
   76:13  error  'languageCode' is defined but never used                no-unused-vars
   77:2   error  Expected indentation of 4 space characters but found 0  indent
  104:32  error  Extra semicolon                                                        semi
  135:1   error  Trailing spaces not allowed                                       no-trailing-spaces
  144:1   error  Trailing spaces not allowed                                       no-trailing-spaces
  144:2   error  Mixed spaces and tabs                                              no-mixed-spaces-and-tabs
  145:5   error  Block must not be padded by blank lines                 padded-blocks
```

lewisgoddard:
We only have 1 error outstanding, but it's only a warning and wasn't introduced in this branch.

`PHP Deprecated:  Methods with the same name as their class will not be constructors in a future version of PHP; TwApi has a deprecated constructor in ./backend/fav-tweets.php on line 18`
`Deprecated: Methods with the same name as their class will not be constructors in a future version of PHP; TwApi has a deprecated constructor in ./backend/fav-tweets.php on line 18`

PR #979, opened 2016-01-10
Fix #903: Hardcoded ISO Size

### LINKED ISSUE #903: Do not hardcode release variables.
- [x] Any string containing the version number, "0.4".
- [x] Any string containing the release name, "Loki".
- [x] Any string containing the release code, "loki".
- [x] The filename.

The ISO Size cannot be done due to translations.

lewisgoddard:
Fix #903: http://beta.elementary.io/branch/origin/fix-903/

### DIFF
--- a/index.php
+++ b/index.php
@@ -54,7 +54,7 @@
                     ?>
                 </div>
                 <button type="submit" id="download" class="suggested-action">Download Freya</button>
-                <p class="small-label"><span data-l10n-off="1">1.15</span> GB (for PC or Mac)</p>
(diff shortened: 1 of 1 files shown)


danrabbit:
Hm does anything need to be translated except "for Mac or PC"? This seems wonky to translate to me

lewisgoddard:
You mean exclude the GB too?

tintou:
We need to translate this because of several reasons :
 * RTL support
 * In French we use the word Octets instead of Bytes
 * We also use a comma instead of a fullstop.
So you might want to do a "%f GB" and parse it the right way 😐

PR #984, opened 2016-01-12
make download button text dynamic based on purchase

btkostner:
It will now say 'Download elementary OS' or 'Purchase elementary OS' based on what you select

### DIFF
--- a/scripts/homepage.js
+++ b/scripts/homepage.js
@@ -224,5 +224,18 @@ $(function () {
         }, 3000);
     });
 
+    // Change Button text on payment click
+    function updateDownloadButton () {
+        var payment_amount = $('#amount-custom').val() * 100;
+        if ($('input#amount-custom').hasClass('checked') && payment_amount < payment_minimum) {
+            $('#download').text('Download elementary OS');
+        } else {
+            $('#download').text('Purchase elementary OS');
+        }
+    }
+
+    $('#amounts').on('click', updateDownloadButton);
+    $('#amounts input').on('change', updateDownloadButton);
+
     console.log('Loaded homepage.js');
 });
--- a/index.php
+++ b/index.php
@@ -53,7 +53,7 @@
                         }
                     ?>
                 </div>
-                <button type="submit" id="download" class="suggested-action">Download elementary OS</button>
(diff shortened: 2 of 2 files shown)


scripts/homepage.js
```diff
@@ -224,5 +224,21 @@ $(function () {
         }, 3000);
     });
 
+    // Change Button text on payment click
+    $(function() {
+      function updateDownloadButton () {
+        var payment_amount = $('#amount-custom').val() * 100;
+        if ($('input#amount-custom').hasClass('checked') && payment_amount <= payment_minimum) {
```
lewisgoddard: It's `payment_amount < payment_minimum`, not `<=`

scripts/homepage.js
```diff
@@ -224,5 +224,21 @@ $(function () {
...
+          $('#download').html('Download elementary OS');
```
lewisgoddard: Should probably use `.text()` instead of `.html()` seeing as it's just text.

scripts/homepage.js
```diff
@@ -224,5 +224,21 @@ $(function () {
...
+        } else {
+          $('#download').html('Purchase elementary OS');
+        }
+      }
+
+      $('#amounts').on('click', updateDownloadButton);
+      $('#amounts input').on('change', updateDownloadButton);
+      $('button#download').on('focus', updateDownloadButton);
```
lewisgoddard: I'm not sure why we need this last event listener.

scripts/homepage.js
```diff
@@ -224,5 +224,21 @@ $(function () {
         }, 3000);
     });
 
+    // Change Button text on payment click
+    $(function() {
```
lewisgoddard: Does this need to be within _another_ anonymous function?

scripts/homepage.js
```diff
@@ -224,5 +224,21 @@ $(function () {
...
+      function updateDownloadButton () {
```
lewisgoddard: Indentation should be 4 spaces, not 2.

scripts/homepage.js
btkostner: It doesn't but it seems to be all the rage in the file

PR #986, opened 2016-01-15
add new preview images and update metadata

### LINKED ISSUE #982: Remove duplicate metadata
Turns out the Twitter falls back to opengraph so we can:

- [x] remove `<meta name="twitter:title" content="Download elementary OS">`
- [x] remove `<meta name="twitter:description" content="A fast and open replacement for Windows and OS X. Pay what you want or download for free.">`
- [x] remove `<meta name="twitter:image" content="https://elementary.io/images/notebook.png">`

danrabbit:
fixes #982 

Twitter falls back to open graph, so these lines are redundant.
Set a sitewide image, so it'll never resolve as image = ''


_templates/header.php
```diff
@@ -40,21 +40,6 @@
         <meta name="author"      content="<?php echo !empty($page['author']) ? $page['author'] : $sitewide['author']; ?>">
         <meta name="theme-color" content="<?php echo !empty($page['theme-color']) ? $page['theme-color'] : $sitewide['theme-color']; ?>">
 
-        <?php
```
lewisgoddard: This little if block here makes twitter display stuff differently, something it never does with open-graph. https://dev.twitter.com/cards/types/summary-large-image

PR #1001, opened 2016-01-26
Don't replace location.hash on every scroll

stuartlangridge:
Don't replace location.hash on every scroll; not only is this inelegant, but when you replaceState some browsers, including Chrome, re-request favicon.ico. Instead, only replace when it's actually required to, so we don't hit favicon.ico on every scroll.


lewisgoddard:
ok to test

Issue #1000, opened 2016-01-26
Strange syntax highlighting for embedded numbers on Installation page


stuartlangridge:
The shasum commands and command results on the Installation page have strange syntax highlighting:

![screenshot at 16-01-26 04 30 16](https://cloud.githubusercontent.com/assets/4356350/12572526/897098d2-c3e5-11e5-9905-353d4967404c.png)

This is pretty much a bug in highlightjs (filed as https://github.com/isagalaev/highlight.js/issues/1062), but it may be worth working around it in the docs (perhaps by not declaring these snippets to be bash).

stuartlangridge:
Seems to be fixed upstream, so the website needs a new version of hljs.

btkostner:
Seems like it fixed some of the highlighting, but it still highlights the 'cd' and 'fc' in the output. I updated the script in the [fix-1000](https://github.com/elementary/mvp/tree/fix-1000) branch.

PR #1003, opened 2016-01-26
Fix 1000

btkostner:
Updates highlight script, and adds a 'nohighlight' class for use in docs. Implemented on installation page to avoid weird highlighting glitches.


btkostner:
GitHub didn't link it, but this fixes #1000

lewisgoddard:
I need a UX approval on these, I don't think they'll like terminal commands on a light background, or output.

screenshot from 2016-01-27 01 04 34

btkostner:
fixed that by making the `nohighlight` class optional on code blocks, so it still includes the `language-bash` class

Issue #1026, opened 2016-02-13
Team page should handle API errors


cassidyjames:
I don't see any handling in the team page for getting anything other than an `ok: true` response from the API. We should decide what should happen if Slack was down or the API key expired, then implement that.

Issue #1022, opened 2016-02-12
Slack API key should not be in public source code


cassidyjames:
Probably not a big deal, but since it's my API key, it could be abused by anyone and get me banned. :P Instead, we should treat it like the Stripe API key and include it only in production.

I can regenerate a new key for when we implement this.

PR #1027, opened 2016-02-13
moved slack token to config file

btkostner:


btkostner:
fixes #1022

btkostner:
As per Slack discussion, team page will silently fail without the people. No php errors will be logged. Fixes #1026

PR #1035, opened 2016-02-17
added more flow to purchase elementary

btkostner:
was working on the new store and got somewhat sidetracked.

Improves flow between purchasing elementary, keeping the cookie, dynamic release name, and purchase / download button updating.

_also adds `.editorconfig`_


backend/config.example.php
```diff
@@ -4,6 +4,8 @@
 
 $config = array(
 
+    'release' => 'Freya Update 2',
```
danrabbit: This should probably be "Freya 0.3.2"

.editorconfig
```diff
@@ -0,0 +1,9 @@
+root = true
+
+[*]
+end_of_line = 1f
+insert_final_newline = true
+indent_style = space
+
+[*.php]
+indent_size = 4
```
lewisgoddard: Can we not have `indent_size` on `[*]` ? This also applies to CSS and JS.

backend/config.example.php
```diff
@@ -4,6 +4,8 @@
 
 $config = array(
 
+    'release' => 'Freya 0.3.2',
```
lewisgoddard: Maybe we should have `release_title` and `release_version`?

backend/payment.php
```diff
@@ -15,11 +16,12 @@
             'amount' => $amount,
             'currency' => 'usd',
             'card' => $token,
-            'description' => 'elementary OS Freya 0.3.2',
+            'description' => $description,
             'receipt_email' => $email,
         ));
         // Set an insecure, HTTP only cookie for 10 years in the future.
-        setcookie('has_paid_freya', $amount, time() + 315360000, '/', '.elementary.io', 0, 1);
+        $encoded = urlencode(str_replace(' ', '', $description));
```
lewisgoddard: I'd like this to remain as descriptive as possible, so maybe have `'has_paid_'.$release_name` ?

index.php
```diff
@@ -5,10 +5,12 @@
     include __DIR__.'/_templates/sitewide.php';
     include $template['header'];
     include $template['alert'];
+    require_once __DIR__.'/backend/config.loader.php';
     require_once __DIR__.'/backend/classify.current.php';
 ?>
             <script src="scripts/slider.js"></script>
             <script>var stripe_key = '<?php include __DIR__.'/backend/payment.php'; ?>';</script>
+            <script>var release = '<?=$config['release']?>';</script>
```
lewisgoddard: I generally avoid shorthands like this for clarity, so I suppose we need to better define a code-style here. I'm sure we use `<?php echo $foo; ?>` elsewhere.

scripts/homepage.js
```diff
@@ -227,19 +228,22 @@ $(function () {
...
-            $('input#amount-custom').hasClass('checked') &&
-            payment_amount < payment_minimum &&
-            $('#amount-custom').val() != ''
-        ) {
+        if ($('#amounts').children().length <= 1) {
             $('#download').text('Download elementary OS');
         } else {
-            $('#download').text('Purchase elementary OS');
+            if (($('input.button#amount-custom').hasClass('checked') &&
+                 $('input.button#amount-custom').val() >= 1) || (
```
lewisgoddard: The custom amount field will never be checked while having an amount smaller than the minimum. You also seem to have lost the `payment_amount < payment_minimum`. I wouldn't worry about the button class, it's not relevant programatically, it's just a css target.

scripts/homepage.js
```diff
@@ -227,19 +228,22 @@ $(function () {
...
+                 $('button.payment-button').hasClass('checked'))) {
```
lewisgoddard:
```
            if (($('input.button#amount-custom').hasClass('checked') &&
                 $('input.button#amount-custom').val() >= 1) || (
                 $('button.payment-button').hasClass('checked'))) {
```

should be formatted as

```
            if (
                (
                    $('input.button#amount-custom').hasClass('checked') &&
                    $('input.button#amount-custom').val() >= 1
                ) || (
                    $('button.payment-button').hasClass('checked')
                )
            ) {
```

Brackets in `if`s should be easy to follow and group from the indentation.

.editorconfig
btkostner: It's only for php files.

Then again I _should_ add indent_size for css and js

backend/payment.php
btkostner: It gets the description name from the payment script, which should be the release variable above ('Freya 0.3.2'). How about just adding 'has_paid_' in front of description?

.editorconfig
lewisgoddard: That's what I meant, can we just apply this to the `[*]` block immediately under `indent_style`.

backend/payment.php
lewisgoddard: Yes, that would work, but see my other comments about the description.

lewisgoddard:
@btkostner If you fix the conflicts, I'm happy to merge.

scripts/homepage.js
```diff
@@ -230,19 +238,21 @@ $(function () {
...
-            $('input#amount-custom').hasClass('checked') &&
-            payment_amount < payment_minimum &&
-            $('#amount-custom').val() != ''
-        ) {
+        if ($('#amounts').children().length <= 1) {
             $('#download').text(translate_download);
         } else {
-            $('#download').text(translate_purchase);
+            if ($('input.button#amount-custom').val() * 100  >= payment_minimum ||
+                $('button.payment-button').hasClass('checked')) {
```
lewisgoddard:
```
if ($('input.button#amount-custom').val() * 100  >= payment_minimum ||
                $('button.payment-button').hasClass('checked')) {
```

should be

```
if (
    $('input.button#amount-custom').val() * 100  >= payment_minimum ||
    $('button.payment-button').hasClass('checked')
) {
```

scripts/homepage.js
```diff
@@ -230,19 +238,21 @@ $(function () {
         var translate_download = $('#translate-download').text();
         var translate_purchase = $('#translate-purchase').text();
 
-        if (
-            $('input#amount-custom').hasClass('checked') &&
-            payment_amount < payment_minimum &&
-            $('#amount-custom').val() != ''
-        ) {
+        if ($('#amounts').children().length <= 1) {
```
lewisgoddard: We never did this before because the default selection once paid is 0, which fails the `>= payment_minimum` check.

scripts/homepage.js
```diff
@@ -230,19 +238,21 @@ $(function () {
...
             $('#download').text(translate_download);
-        } else {
+        } else if (
+            $('button.payment-button').hasClass('checked') ||
+            payment_amount >= payment_minimum
```
lewisgoddard: Change to `$('#amount-custom').val() * 100 >= payment_minimum` and drop the variable, there's no point assigning something we only ever use once, and sometimes not at all.

lewisgoddard:
I'm happy for this to merge, but it needs some configuration files changing, so @fabianthoma will have to do it or it will break everything.

lewisgoddard:
retest this please

fabianthoma:
Config on the server has been changed. As @lewisgoddard said it's good to merge, I'll do that now

PR #1059, opened 2016-02-25
line-height changes to get the search text in the right place

Lukewh:
https://github.com/elementary/mvp/issues/1050


btkostner:
This fixes the line height on the search box for the text, but after looking at it again, it seems the whole slingshot area line height seems to be messed up. I'll put it on my todo list, but this PR looks good to merge

lewisgoddard:
So this is tested to fix the issue, but only does so partially?

btkostner:
it fixes the issue, but there is a bigger, overarching problem with the slingshot area

Issue #1054, opened 2016-02-20
Purchase/Download button not translatable


cassidyjames:
Since that button's text is being set directly in js, it's not translatable. As such, all languages show the button in English.

btkostner:
We could do that, and just not have any animation.

Option 2 being we could have two spans of text on the website, and then we have the javascript just replace the button text with the translated span text.

lewisgoddard:
Oh, I like Option 2, that's clever. Hidden down the bottom with the modals?

btkostner:
sounds good to me

PR #1060, opened 2016-02-26
set download button text in php for translations

btkostner:
makes button text set in php for translations for #1054


lewisgoddard:
I'm not sure if `hidden` is well [supported](http://caniuse.com/#feat=hidden) enough, unless we are willing to drop IE 10. Other things just user `style="display:none;"`

btkostner:
ah, i'll add that too then

PR #1064, opened 2016-03-01
added buttons to install docs

### LINKED ISSUE #1063: Make the "create elementary installer" link more visible
In the [docs](https://elementary.io/docs/installation#creating-an-installation-medium), there's link to software that installs the iso into a usb. 

If I'm just scanning the document because I only want the link to that installer I would like to be able to find it easily, a button would be nice.

btkostner:
Fixes #1063

### Changes Summary

- Added buttons for software on installation doc


danrabbit:
Hm, it looks like in every case that the download button is *after* the directions on how to use the tool. That seems kind of backwards. The user would want to download the tool first and then read how to use it, no? :p

PR #1067, opened 2016-03-02
Fix 1038

### LINKED ISSUE #1038: logotype.svg is inaccurately-named
The SVG includes "elementary OS", so it's not really the logotype and cannot be used to refer to anything other than the OS. I propose renaming it (and all references to it) to something like "logotype-os.svg"

btkostner:
Fixes #1038

### Changes Summary

- moved `logotype.svg` to `logotype-os.svg`


btkostner:
woops, seems like another commit tagged along for the ride. one moment...

btkostner:
Should be all good now

PR #1068, opened 2016-03-03
better scroll handling in docs

### LINKED ISSUE #1002: Best to not do work in a JavaScript scroll handler
A number of places in the elementary site do work in a scroll handler: examples include various `    $(document).on('scroll', function (event)` functions in `scripts/docs/main.js`. In general, it's best to not do actual work in a scroll handler, because slow devices (in particular mobiles) are likely to have jerky scrolling because of it. Instead, it's better to debounce; something like:

    var scrollDebounce;
    $(document).on('scroll', function (event) {
        clearTimeout(scrollDebounce);
        scrollDebounce = setTimeout(actuallyDoWork, 50);
    });

and then do the work in `actuallyDoWork()`. This prevents jerky scrolling in less-powerful devices.

btkostner:
This adds better handling of the scroll handler. Includes a time sensitive holder set to 5 seconds (if it hasn't been updated in 5 seconds the update it) and a 1 second debounce (wait 1 second to see if you are continue to scroll before update). It also decreases the upper limit for the current section from 1/3 of the page to 1/6 of the page, and adds fallback to the first section, so if anything weird happens (say an unknown hash in the url) we won't have an undefined hash in the url.

Fixes #1002


Lukewh:
Would something like
```javascript
var scrollElements = $('h1[id], h2[id]', '.docs');
var docElements = [];
for (var i = 0, ii = scrollElements.length; i < ii; i += 1) {
  docElements.push({
    top: $(scrollElements[i]).offset().top),
    ele: scrollElements[i]
  });
}

...

function scrollHandle() {
  var docViewTop = $(window).scrollTop();
  for (var i = 0; i < docElements.length; i++) {
    var elemTop = docElements[i].top;
}
```
be better?

1. Cache the document element top position, to avoid hitting the DOM a lot.
2. docViewTop should be checked once outside the loop, again to avoid DOM queries

btkostner:
technically it would be more efficient as it is not calling the DOM on every handle, but it does not take into account if they move position in the document, like on a resize. Honestly I don't think you would notice that much of a difference if it was moved out of the handle considering how fast computers are now-a-days.

Lukewh:
That's fair. My day job makes me notice things like that s'all.

lewisgoddard:
I see a drop from 300ms average to less than 100ms average script time, plus a maximum that got cut in half. There's still a couple of hiccups, but this is far better, and feels smoother.

PR #1079, opened 2016-03-10
Fix #1076 Fixed Download Instructions.

### LINKED ISSUE #1076: The instructions for verifying Downloads is incorrect
### Expected Behaviour
it should return the hashsum

### Actual behaviour
Assuming the ISO is in the downloads folder, copy and pasting "shasum -a 256 elementaryos-0.3.2-stable-i386.20151209.iso" into the terminal will fail as most terminals starts in the home folder

### Steps to reproduce the behaviour
1. Open a terminal
2. copy and paste the command line

### System
- OS X
- 10.11
- Firefox
- 44.0.2

A simple fix would be to have something like ``cd Downloads; shasum -a 256 elementaryos-0.3.2-stable-i386.20151209.iso``

Edit: this also affects Linux/Ubuntu obviously as it uses the same command.

Pioneer11X:
This is my first commit and pull request for this repo. Please let me know if I have done something wrong and I'll improve upon it.


Pioneer11X:
I'm not sure if the second commit is necessary.

lewisgoddard:
I agree that the second commit isn't necessary. We probably want to change this for OS X as well, seeing as that is command line too.

Pioneer11X:
This commits are just for OSX. I'll push another one for Linux in a few hours.

lewisgoddard:
ok to test

lewisgoddard:
ok to test

Pioneer11X:
Hey, sorry to bother, but what does ok to test mean?? Do I have to test it? or would you deploy it in your test servers? I'm new to contributing here.

PR #1093, opened 2016-03-23
use composer for lib folder

btkostner:
### Changes Summary

- Use composer for dependencies in `/backend/lib`
- Updated stripe library to latest version

### Why Even

Because some apis in a later branch (new-store) require use of composer (a long with most php libraries) and it would better to keep this in a smaller, more focused PR.


lewisgoddard:
add to whitelist

lewisgoddard:
Should we add documentation on how to update the libs?

btkostner:
We can, but considering all the libs are stored in the repo, I wouldn't consider it a high priority.

PR #1110, opened 2016-04-13
use google fonts for raleway

btkostner:
### Changes Summary

- Removes our local copy of Raleway, and replaces it with Google fonts version


KeitIG:
Kewis, does this Raleway version fix the encoding problem with some foreigner languages ?

lewisgoddard:
I haven't had a chance to test this yet, but you can see the available characters here: https://www.google.com/fonts/specimen/Raleway#charset

PR #1118, opened 2016-04-21
Fix #412: Quote tweet content not translated

### LINKED ISSUE #412: Quote tweet content not translated
We're translating the display text, but not the text that is tweeted.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/8658579-quote-tweet-content-not-translated?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github).
</bountysource-plugin>

lewisgoddard:
Fixes #412

### Changes Summary

- Add `data-tweet-suffix`
- Automatically update tweet link after load.
- Include script in homepage.


btkostner:
This does not work on page first load, you have to refresh on the language for it to take effect. The bigger issue being that with the translated text, it will usually be greater than twitter's character count limit.

Also, it seems weird to be doing this client side instead of with php, but that's just opinion.

lewisgoddard:
It works on first page load for me, I don't see any reason why it wouldn't unless your browser had just cached the page. I should probably truncate the quote right? or just tweet in english?

I don't want to do this in PHP because it means more fiddling with the DOM when translating, and that area is already a mess as it is. We have other scripts that modify links on page load.

lewisgoddard:
Ready for another review.

btkostner:
I adjusted it a bit so '…' would be inside the quotes. LGTM for me

lewisgoddard:
Awesome, can we approve each others bits?

PR #1124, opened 2016-04-22
Fix #1047: Redirect to non-www.

### LINKED ISSUE #1047: www subdomain does not redirect
`http://www.elementary.io` appears as a standalone site which hurts SEO. It should instead redirect to `http://elementary.io`

lewisgoddard:
Fixes #1047

### Changes Summary

- Add www. redirection


lewisgoddard:
When this is merged we're going to need to add the SSL block to _both_ server blocks, not just one.

PR #1129, opened 2016-04-24
Fix Issue #1112 : Make a note in the code style guide about translato…

### LINKED ISSUE #1112: Make a note in the code style guide about translator comments
Comments for translators use ///

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/32958837-make-a-note-in-the-code-style-guide-about-translator-comments?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ariessa:
Fixes #1112 

Added a note to use `///` for translators' comments.

This pull request  is ready for review.

 Found out about this issue at 'Up For Grabs'.

### DIFF
--- a/docs/code/reference.md
+++ b/docs/code/reference.md
@@ -117,6 +117,8 @@ than good.
         l = n * 2 + 4;   // Clear l variable
     }
 
+For translators' comments, use triple forward slash `///`.
+
 ## Variable names, class names, function names {#variable-names-class-names-function-names}
 
 
--- a/docs/code/reference.md
+++ b/docs/code/reference.md
@@ -118,6 +118,11 @@ than good.
     }
 
 For translators' comments, use triple forward slash `///`.
+```
+{
+    "elementary OS": false /// Can't be translated
+}
+```
 
 ## Variable names, class names, function names {#variable-names-class-names-function-names}
 


lewisgoddard:
You should give an example of how a translation comment might look in real code, maybe use an existing snippet.

lewisgoddard:
ok to test

docs/code/reference.md
```diff
@@ -117,6 +117,13 @@ than good.
         l = n * 2 + 4;   // Clear l variable
     }
 
+For translators' comments, use triple forward slash `///`.
+```
```
lewisgoddard: Generally, we use four space indentation instead of triple `

docs/code/reference.md
```diff
@@ -117,6 +117,13 @@ than good.
...
+{
+    "elementary OS": false /// Can't be translated
```
lewisgoddard: This should probably be `/// Shouldn't be translated because it is a brand name.`

Philip-Scott:
@ariessa Hey! :)

Translation comments are usually used in places were a translation IS taking place. This is usually a short note for translators that will appear on the `/po/app.pot` file

For example, if you have this in your code: 

    /// These are keyboard keys that must be valid since they will be used across the app
    case CHANGE_MODE:   return _("<Ctrl>M"); 

the .pot file will include it like:

    #. / These are keyboard keys that must be valid since they will be used across the app
    #: ../src/Application.vala:37
    msgid "<Ctrl>M"
    msgstr ""

So i would say that it needs a better example on the MR, and maybe also include it on https://elementary.io/docs/code/getting-started#Adding-Translations

Thank you!

danrabbit:
Yes, these comments are often used for disambiguation or to help in the creation of accurate translations. A real example I encountered recently:
```
/// TRANSLATORS: Refers to files the mimetype, not Files the app
_("Files")
```

This is potentially important especially for non-latin-character translations so the translator knows if the translation should be literal or just phonetic. Could be used to resolve a gender-agreement issue etc

btkostner:
Closing due to inactivity. Let me know if you plan to continue working on it and I will be happy to reopen.

PR #1158, opened 2016-05-14
[WIP] Loki showcase

btkostner:
A new showcase to replace our homepage carousel.

Things that still need work:
- [ ] Copy
  - [ ] Section header
  - [x] Mail
  - [ ] Calendar
  - [x] Files
  - [x] Terminal
  - [x] Scratch
  - [ ] Camera
- [ ] Screenshots
  - [ ] Midori
  - [ ] Mail
  - [ ] Camera

All files in the `/images` should be compressed when we are ready to merge


danrabbit:
I feel like the icon and copy should go above the apps. Currently, on smaller displays, you could click an icon and not even see the copy.

btkostner:
As far as the copy, the reason I did it below the screenshot was because it follows convention for the rest of the page, where copy would be below the image on smaller viewports.

Animating would me amazing. Slingshot's getting all the love at the bottom of the page `:P`, I'll put on my creative hat on and start thinking of code for that.

danrabbit:
Some minor nitpicks about Terminal:
- [ ] There should be a space after the ":" in the tab and title name. As in "Home: rm -rf" and not "Home:rm -rf"
- [ ] Looks like we're not currently loading the bold font weight (700) for open sans in the site <head>
- [ ] Non-interactive elements (titlebar buttons, tab buttons) should probably not be click-able
- [ ] Should we replace "user" with "ellie" as a subtle nod to our [fictitious demo user](https://drive.google.com/a/elementaryos.org/folderview?id=0B_sDFLg1fuxGb3ctUFduRWRqVWc&usp=sharing)
- [ ] The Terminal is way wider than tall on my 13" display. Any way to make it a bit more of a golden rectangle?

danrabbit:
- [ ] Need to add and use the [48px Terminal icon](https://github.com/elementary/icons/blob/master/apps/48/utilities-terminal.svg) instead of scaling the larger one

danrabbit:
- [ ] Fix midori-browser.org address. It isn't https

danrabbit:
- [ ] We should take HiDPI screenshots for apps that support it

Issue #1131, opened 2016-04-25
Burndown chart is too-aggressively cached


cassidyjames:
### Expected behavior

I hop over to https://elementary.io/get-involved#desktop-development and see the up-to-date burndown chart. It's not heavily cached since it's a daily-updating chart.

### Actual behavior

I hop over to https://elementary.io/get-involved#desktop-development and see a burndown chart from several days ago, or whenever I first visited the page. It seems to be cached, forcing a hard refresh to get the new graph.

### Steps to reproduce the behavior

1. Go to https://elementary.io/get-involved#desktop-development
2. Wait a few days
3. Revisit https://elementary.io/get-involved#desktop-development in the same browser

### System information

- **Operating system**: elementary OS 
- **Operating system version**: Freya and Lokii
- **Browser**: Chrome
- **Browser version**: 50.0.2661.86 (64-bit)
- **Enabled Browser Extensions**: Pocket

cassidyjames:
Not sure what our options are here since the caching is client-side. Can we set a header for the page/js that it's changed or something?

lewisgoddard:
I think we need to set an exclusion in Cloudflare so that this shouldn't be cached for more than a day, or set a header for the location in the Nginx config. https://support.cloudflare.com/hc/en-us/articles/200169266-Does-CloudFlare-honor-my-Expires-and-Cache-Control-headers-for-static-content-

PR #1133, opened 2016-04-25
Fixes #1131: Exclude `/backend/chart.json` from normal caching.

lewisgoddard:
Fixes #1131: Burndown chart is too-aggressively cached

### Changes Summary

- Exclude `/backend/chart.json` from normal caching.

This pull request is ready for review.

PR #1186, opened 2016-06-15
Prevent the browser from caching the chart data for more than one day

### LINKED ISSUE #1185: Get Involved chart is not updated
### Expected behavior

When accessing the Get Involved page, I expect the chart data to display the data in the chart.json file generated by chart.php.

### Actual behavior

If I have already visited the page in the last few days, it does not refresh the chart data, instead it uses a cached version of chart.json to display the chart on the Get Involved page.

### Steps to reproduce the behavior

1. Open the Get Involved page on day 1
2. Wait a few days and ensure that backend/chart.php has been triggered (I verified with another browser).
3. Open the Get Involved page on day 2 with the same browser as the one used on step 1
4. Look at the chart, it hasn't changed whilst if you open it with another browser that hasn't cached the chart.json file, it will display the latest version of the data.

### System information

- **Operating system**: any
- **Operating system version**: any
- **Browser**: any
- **Browser version**: any
- **Enabled Browser Extensions**: none

ggobbe:
Fixes #1185 (duplicate of #1131)

### Changes Summary

- Create a suffix in the form of YYYY-MM-DD
- Append the suffix to the chart.json url in the ajax request

### DIFF
--- a/scripts/get-involved.js
+++ b/scripts/get-involved.js
@@ -1,6 +1,8 @@
 $(function () {
+    var dateYearMonthDay = new Date().toISOString().substring(0, 10);
+    var preventingCacheSuffix = '?' + dateYearMonthDay;
     $.ajax({
-        url: 'backend/chart.json',
+        url: 'backend/chart.json' + preventingCacheSuffix,
         dataType: 'json'
     }).done(function (data) {
         var labels = {


lewisgoddard:
This would work, but would also be fixed in #1133

ggobbe:
Sorry for the duplicate. About this solution I think the main difference is that it will keep working even when/if using a different server than nginx, i.e. the code marks the intent, not the server configuration. But it is up to you to decide what you prefer.

PR #1189, opened 2016-06-18
keywords, not x-gnome-keywords

danrabbit:
doesn't need the x-gnome bit


fabianthoma:
Yup, it's in the freedesktop specification :+1:

PR #1195, opened 2016-06-28
[WIP] Webtorrent

lewisgoddard:
### Changes Summary

- [x] Switch on modal to an auto-starting WebTorrent if available.
- [ ] Change default to an auto-starting direct download.
- [ ] Implement save button for WebTorrents.

This pull request is not ready for review.

http://beta.elementary.io/branch/origin/webtorrent/

https://gitter.im/feross/webtorrent


lewisgoddard:
For some reason, Sintel starts for me fine, but the elementary download does not. I know I need to update CORS on most servers, but I expect it to work on AMS3.

fabianthoma:
@lewisgoddard well do we have webtorrent compatible seeders? I don't think we currently do, even the ones on the servers probably aren't capable of that

lewisgoddard:
@fabianthoma The webseed should take care of that, but I just realised that that won't send metadata.

lewisgoddard:
It seems CORs stops the webseed working, which we need to fix for this to even start.

lewisgoddard:
I would like a design decision on this before any more work is done. The basic technology is there, but we need to decide if we prefer people downloading in-page with webtorrent, or out-of-page with direct HTTP.

danrabbit:
I think that web torrent is probably more ideal than http. Theoretically users should get the fastest, most reliable download over torrent.

I still think users should be able to choose magnet link in either case. Personally, I would rather use my native torrent client than have to keep my browser open to complete a torrent. So I think we should consider web torrent a replacement only for the HTTP download and not for magnet links.

Hopefully that was the answer to the question you had in mind :p

lewisgoddard:
In that case, I have a few questions that need addressing before moving forward:

Should the webtorrent automatically start when the download is opened?
Should a direct download automatically start if the webtorrent is not available?
Should the webtorrent continue if the model is minimized?
Should the webtorrent automatically initiate save if the tab is focused?
Should the webtorrent automatically initiate save if the tab is not focused?
If a user cancels the webtorrent, should we offer to resume the webtorrent, offer direct HTTP instead, or both?

danrabbit:
> Should the webtorrent automatically start when the download is opened?
If by this you mean when the payment is completed, no. You should see a dialog that allows you to choose between magnet and webtorrent. The same as how now you choose between http and magnet

> Should a direct download automatically start if the webtorrent is not available?
If the link for webtorrent is chosen and webtorrent is not available, then yes we should fall back to http

> Should the webtorrent continue if the model is minimized?
I don't think we should introduce a concept of minimize as then we'd have to have some kind of window management and it just gets really messy. The modal should either be closed or open.

> Should the webtorrent automatically initiate save if the tab is focused/no focused?
After completion yes I think it would make sense to automatically initiate save. I don't see why someone would perform the download and then not save.

> If a user cancels the webtorrent, should we offer to resume the webtorrent, offer direct HTTP instead, or both?
Presumably if the browser is capable of web torrent we want people using this instead of http right? I was under the impression that we're essentially trying to deprecate regular http downloads. I think it would make sense to try to resume a download if that's possible.

lewisgoddard:
Okay, thank you!

emersion:
@lewisgoddard Do you need help on this one?

lewisgoddard:
Please don't delete yet.

Issue #1123, opened 2016-04-21
Magnet links do not use webseeds.


lewisgoddard:
### Expected behavior

Torrents should use the distributed webseeds.

### Actual behavior

Because the webseeds are timestamp protected, we didn't initially offer this. I now realise that this is silly easy to implement.

PR #1210, opened 2016-07-18
Validate custom amount in browsers that does n't support number type.

### LINKED ISSUE #1192: Entering $0 switches back to $10 in Firefox
Reported via Twitter

 https://twitter.com/dkomm/status/746053028822810624?s=09

niyasc:
Fixes #1192 

### Changes Summary

- Check each character that is being entered to input field. If it anything other than arrow keys, delete, backspace, ., or digit prevent default action.

### Known issue.

- Consequent decimal points as in `1..` can't be validated. Because, when value in input field is `1.`, `.val()` method returns just `.`.

### DIFF
--- a/scripts/homepage.js
+++ b/scripts/homepage.js
@@ -33,10 +33,27 @@ $(function () {
             updateDownloadButton();
         }
     };
+    var amountValidate = function(event) {
+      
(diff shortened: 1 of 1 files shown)


btkostner:
I cleaned this up a bit in #1123. Consider this merged

PR #1212, opened 2016-07-19
nginx.conf: redirect elementary.io/loki somewhere

codygarver:
nginx.conf: redirect elementary.io/loki somewhere

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -90,7 +90,7 @@ server {
         return 301 /docs/human-interface-guidelines;
     }
     location = /loki {
-        return 301 http://blog.elementary.io/;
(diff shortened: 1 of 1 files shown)


nginx.conf
```diff
@@ -89,6 +89,9 @@ server {
     location /docs/human-interface-guidelines/ {
         return 301 /docs/human-interface-guidelines;
     }
+    location /loki {
```
lewisgoddard: This would redirect `/loki-something/something` as well. Use `location = /loki` to exact match the URL.

lewisgoddard:
Merging changes to the config doesn't automatically update it.

fabianthoma:
Server Side Edit done.

PR #1216, opened 2016-07-20
Font revert revert

btkostner:
This reverts the revert of using Google fonts with a local copy of Raleway.


lewisgoddard:
https://beta.elementary.io/branch/origin/font-revert-revert/?lang=ru

PR #1221, opened 2016-07-25
Removing description of old capnet-assist behavior

matzipan:
The behavior described on the capnet-assist page has been fixed in http://bazaar.launchpad.net/~elementary-apps/capnet-assist/trunk/revision/78


danrabbit:
The only case this page would ever appear in is the one described in that paragraph. If that's not the case,  capnet will never load this page.

PR #1222, opened 2016-07-26
update translations

btkostner:
pull from transifex


liddack:
Some of the translations are updated. It's time to pull from Transifex again.

btkostner:
Will do after we get unit tests back up. It may take a while for them to show on the site though.

PR #1223, opened 2016-07-26
Number validation

btkostner:
cleaned up some whitespace from #1210


btkostner:
I just cleaned up some whitespace and some operators (using `===` vs `==`) and merged master in.

It does not fix all use cases (I was still experiencing multiple decimal points on my android phone), but it improves validation on desktop (chrome) quite a bit.

PR #1224, opened 2016-07-28
update stripe lib and add ups lib

btkostner:
this updates the stripe lib and adds the ups lib needed for the `new-store` branch.


lewisgoddard:
https://beta.elementary.io/branch/origin/ups-lib/

PR #1234, opened 2016-08-05
Make Open Source page prettier

danrabbit:
* Re-title as "open source"
* Use colors from our palette for sub items
* Make platform items flat
* Make anchor elements surround text as well as icons
* Add anchor elements for elementary platform items
* Use button class to re-use styles
* Use grid instead of row
* Fix dotted line and platform item margins on mobile


danrabbit:
Sam said he approves in Slack

PR #1236, opened 2016-08-05
Travis

btkostner:
This adds travis unit test scripts for mvp so we no longer rely on our internal Jenkins server for PR testing.

As of right now tests are failing. Fixing tests will come after this is merged.


lewisgoddard:
Are you intending to add dans "use new network icon"? It seems out of place.

btkostner:
It's already in master, I think my git history got messed up somewhere.

PR #1240, opened 2016-08-05
Js lint

btkostner:
This fixes the linting issues with Javascript files.

Most changes consist of:
- removing `;` from end of line
- spaces around mathematical operations
- indentation fixes
- headlessCamelCase variables


developer.php
```diff
@@ -9,9 +9,9 @@
     require_once __DIR__.'/backend/classify.current.php';
 ?>
 
-<script>var release_title = 'Loki';</script>
-<script>var release_version = '0.4-beta2';</script>
-<script>var download_region = '<?php echo $region; ?>';</script>
+<script>var releaseTitle = 'Loki';</script>
```
lewisgoddard: There are several of these inline scripts that still show `;`

PR #1241, opened 2016-08-05
fixes linting issues with css styles

btkostner:
This lints all of the css files with csscomb

Biggest changes being:
- indentation
- leading zeros
- sorting properties
- lower case hex color values


styles/docs.css
```diff
@@ -180,10 +180,10 @@ nav {
     border-right: 1px solid #f0f0f0;
     display: -webkit-flex;
     display: -ms-flexbox;
-    display: flex;
+    display:         flex;
     -webkit-flex-direction: column;
-    -ms-flex-direction: column;
-    flex-direction: column;
+        -ms-flex-direction: column;
+            flex-direction: column;
```
lewisgoddard: As I said before, CSSComb does not cope well with these multiple displays.

styles/docs.css
```diff
@@ -180,10 +180,10 @@ nav {
     border-right: 1px solid #f0f0f0;
     display: -webkit-flex;
     display: -ms-flexbox;
-    display: flex;
+    display:         flex;
```
lewisgoddard: As I said before, CSSComb does not cope well with these multiple displays. You suggested switching it out for something else? We only need the `-webkit-` for Android really.

lewisgoddard: http://caniuse.com/#feat=flexbox

PR #1244, opened 2016-08-06
DOC : missing french word

glutamatt:


btkostner:
Please don't edit the translation files directly. All of our translations are handled by transifex [here](https://www.transifex.com/elementary/elementary-mvp/).

PR #1157, opened 2016-05-13
Loki

btkostner:
# Road to Loki!

Home page:
- [x] Update hero screenshot
- [x] Update app screenshots
    - [x] Epiphany
    - [x] Calendar
    - [x] Camera (has insensitive buttons in header)
    - [x] App Center
    - [x] Files
    - [x] Mail
    - [x] Music
    - [x] Photos
    - [x] Scratch
    - [x] Terminal
        - [x] Broken on mobile
    - [x] Videos
- [x] Update apps in Slingshot
    - [x] Remove software updater, font viewer, software & updates, archive manager
    - [x] Split apps and settings in search
- [x] Update Ubuntu security notice link

Installation Guide:
- [x] Remove 32bit stuff

Getting Started:
- [x] Use `apt` instead of `apt-get`
- [x] Hello world
    - [x] Keywords not X-GNOME-Keywords
- [x] The build system
    - [x] Use build modules package instead of branching

Developer:
- [x] Update hero section

Support:
- [x] Midori :arrow_right: Epiphany
- [x] Geary :arrow_right: Mail

Privacy Policy:
- [x] Update browser icons

#### Release
- [x] Pull translations
- [x] Update release variables
- [x] Make release links use release variables (filename, fixes #903)
- [x] Update magnet hash
- [x] Update disk image size
- [x] Update Filenames & Hashes in the Installation Guide
- [ ] Empty Average Payments database

PR #1252, opened 2016-08-09
Add Countdown Flipclock

danrabbit:
- [x] Set the correct time 
- [x] Figure out why jquery isn't loading 
- [x] Load countdown.css in countdown.php template


scripts/countdown.js
```diff
@@ -0,0 +1,9 @@
+var clock = $('.countdown').FlipClock({
```
lewisgoddard: This should all be wrapped in a jQuery `onload`.

_templates/countdown.php
```diff
@@ -0,0 +1,9 @@
+<script src="https://code.jquery.com/jquery-3.1.0.min.js" integrity="sha256-cCueBR6CsyA4/9szpPfrX3s49M9vUU5BgtiJj06wt/s=" crossorigin="anonymous"></script>
```
lewisgoddard: I don't understand what you're doing with the script tags here, jQuery should already be loaded by the page anyway, why are we including it again?

_templates/countdown.php
```diff
@@ -0,0 +1,9 @@
...
+
+<div class="countdown-wrapper">
+    <div class="countdown"></div>
+    <a class="read-more" href="#">Click to Continue</a>
+</div>
+
+<script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/flipclock/0.7.8/flipclock.min.js"></script>
```
lewisgoddard: We're mixing CDNs, we'd be better off updating jsDelivr and loading from there.

_templates/countdown.php
danrabbit: Yeah I noted this in the PR comments. For some reason jquery isn't being loaded until later, maybe. I dunno. It's not working though.

lewisgoddard: We load it asynchronously, so you need to wrap `scripts/countdown.js` with the usual jQuery onload parameter that only gets fired once it's there.

_templates/header.php
```diff
@@ -100,6 +100,7 @@
         <script>
             <?php include __DIR__.'/../scripts/jql.min.js'; ?>
             jQl.loadjQ('https://cdnjs.cloudflare.com/ajax/libs/jquery/3.1.0/jquery.min.js')
+            jQl.loadjQdep('https://cdnjs.cloudflare.com/ajax/libs/flipclock/0.7.8/flipclock.min.js')
```
lewisgoddard: Could this not be moved to the `countdown.php`, keep it neater? These dependencies don't have to be loaded together.

btkostner: I was having some trouble with it being called in the countdown.

- If it's after the `boot` call then it won't work as intended
- If the `boot` call is put in the footer then there is a chance the other scripts will load and jQl won't intercept the `ready` calls
- If it's a regular (non async) call then it won't always load after jQuery is loaded and give an undefined error

btkostner: Once this is merged with master, we can update the loki branch to take it out so on release we are all clean.

lewisgoddard:
I hereby approve this code.

_templates/header.php
lewisgoddard: I've made a pull request for flipclock to be auto-updated on JSDelivr: https://github.com/jsdelivr/jsdelivr/pull/13425

lewisgoddard:
This now has conflicts, and we might be able to remove a lot of the CSS as I have asked JSDelivr to host it.

danrabbit:
We have a date. September 5th. So as soon as we get this sorted it can be put up.

danrabbit:
- [ ] Can we add a "Days" section or would that be too much work?
- [ ] Do we know what timezone this is set to? Is it UTC? We need it to be on the same timezone no matter who's viewing so it ends a the right time for everyone

btkostner:
The countdown on it's own is a little vague for someone who is first visiting the site. We should include some hype text above it so people know what we are counting down to.

- hype text

danrabbit:
It's intentionally vague. We never include text indicating what it's for

lewisgoddard:
https://github.com/jsdelivr/jsdelivr/pull/13425 being merged will neaten this up.

lewisgoddard:
I'm not sure if we have modified it here at all, but there's now a CDN hosted version of `flipclock.css` at https://cdn.jsdelivr.net/g/flipclock@0.7.7(flipclock.css)

btkostner:
When we push this to master, we should push translations for #1157 and hopefully get most of the content translated before release.

- [ ] Push Loki branch translations

PR #1254, opened 2016-08-11
New new store

btkostner:
At this point the store is ready for some more eyes. Everything from the store index to creating the order works. Now it's time to test and refine.

Things that need refinement:
- [x] cart form styles
- [x] copy for store items in `/data/store.json`
- [ ] cart localization
- [ ] figuring out a way to translate all the content

Store flow:
`index` -> `cart` -> `checkout` -> `order`
The `inventory` page is for setting cookies with javascript.


danrabbit:
- [ ] I read recently that using inline buttons instead of comboboxes increases conversions. Since we have to take up all that horizontal space anyways, can we use buttons (like the selector on the homepage) for sizes?

- [ ] I still think we need an explicit "remove" button somewhere in the cart for items

store/cart.php
```diff
@@ -0,0 +1,124 @@
...
+        <div class="list__footer">
+            <h4>Sub-Total: $<?php echo \Store\Cart\get_subtotal() ?></h4>
+        </div>
+    </div>
+
+    <div class="whole">
+        <h2>Shipping Information</h2>
+    </div>
+
+    <div class="whole grid grid--address">
```
cassidyjames: I think we should add labels to the fields in this form; it's too easy to start typing and then not know the requirements of the field. See: https://medium.com/simple-human/always-use-a-label-a39ceab554e6#.t2zduptpp

scripts/store/cart.js
```diff
@@ -0,0 +1,108 @@
...
+        total += t
+    })
+
+    $('.list--product .list__footer h4').text('Sub-Total: $' + parseFloat(total).toFixed(2))
+}
+
+/**
+ * POSTs to inventory endpoint to update cart quantities without page refresh
+ */
+$('.list--product .list__item input[name$="quantity"]').on('change', function (e) {
```
cassidyjames: We should also do this after a timeout of no entry, probably around 500ms. We do this on the System76 cart because it's not obvious that unfocusing the entry will update live, so people end up hitting enter.

store/cart.php
```diff
@@ -0,0 +1,124 @@
...
+
+            <div class="list__item" id="product-<?php echo $index ?>">
+                <img src="<?php echo $product['image'] ?>"/>
+                <div class="list__info">
+                    <b><?php echo $variant['name'] ?></b>
+                    <span><a href="/store/inventory?math=set&quantity=0&id=<?php echo $id[0] ?>&variant=<?php echo $id[1] ?>">Remove</a></span>
+                </div>
+                <div class="list__detail">
+                    <input type="hidden" name="product-<?php echo $index ?>-id" value="<?php echo $index ?>">
+                    <input type="hidden" name="product-<?php echo $index ?>-price" value="<?php echo $variant['price'] ?>">
```
cassidyjames: I haven't followed the logic completely here (still reviewing) but this stands out to me as odd/possibly dangerous. Is this just being used as as convenience cache?

backend/store/address.php
```diff
@@ -0,0 +1,332 @@
...
+
+        $this->email = htmlspecialchars($in, ENT_XML1, 'UTF-8');
+    }
+
+    public function set_phone ($in) {
+        if (!isset($in)) {
+            throw new \ValidationException('Phone number is not set');
+        }
+
+        if (!preg_match("/^[\+0-9\-\# ]{7,}$/i", $in)) {
```
cassidyjames: Can we just sanitize this for the user? It's awful to allow the user to type in (555) 123-4567 and then not actually accept it.

cassidyjames:
Clearing the address form on error is pretty bad; can we retain that when the page loads with an error?

scripts/store/cart.js
lewisgoddard: Or add a sugar-pill button next to it which says update, that would trigger the onchange instantly without having to even do anything.

lewisgoddard:
Client-side validation would help alleviate that.

store/cart.php
btkostner: It's used for JavaScript and updating price values. Everything on the order page is double checked before any API is called to make sure there is no weird tampering with data.

backend/store/address.php
btkostner: Yes we can. The phone number isn't needed for any machine function. It's just in case we need to contact them. I'll add parentheses to the regex and try to sanitize it to just numbers

btkostner:
At this point, I really wish mvp was more mcv styled as sending information from one page to another is a real pain. Not to mention trying to render different pages depending on logic is super convoluted. I can implement some basic client side validation to help with errors, but at this point I **REALLY REALLY REALLY** want to get a client side framework and make the store (almost completely) client side. This will help _a lot_ with internationalization, validation, and sending information between pages.

store/cart.php
btkostner: As far as I know we have no other user input (aside from the single `input` on the homepage) . I will need design for the input label.

btkostner:
### A detailed flow
**I'm not a novel writer**

#### Things to know
Product - a general item that includes the same name, description, image, type, and overall design files.

Variant - a version of a product. Changes can include, what's being printed on, color, name, size, cost, and price.

#### 1) `index.php`
Rendered by php with all products found in `data/store.json`. All products include a modal with data for the first variant with a hidden `variant` `input`. On any change of options (color, size, quantity) javascript goes through the `data/store.json` variants and tries to find the new variant id. It then replaces the hidden input with the new value. On `add to cart` we move to `1.5`.

#### 1.5) `inventory.php`
This file has no actual page, but instead turns a product id, variant id, and quantity into a cart array to be set as a cookie. On success it redirects to the cart.

#### 2) `cart.php`
Rendered by php with all variants in the cart, it includes the address form. Javascript listens for when quantity changes (or removes) occures and does a call to `inventory.php` so we don't refresh the page. It also listens to the `country` selector to hide, show, and update the `state` selector. It currently does _not_ have any form validation. It `POST`s all data (including the cart items) to `3`.

#### 3) `checkout.php`
This page validates all `POST` information internally, and then calls API calls to Google maps, and Printful. Google maps returns an image we display. Printful API returns shipping prices for all the variants we want to order. The page includes a `radio` select for different shipping types. Javascript listens for the radio change to change various shipping information (price, time, tax) on the page. On `order` button click, a stripe modal appears to take payment option. The price _is_ set by javascript, but the actual order does not take place on the client. After stripe successfully gets the payment option we move to `4`.

#### 4) `order.php`
This page validates all `POST` information, including calling the Printful API again to verify shipping price. It also verifies _every_ value it receives (price, tax, shipping, total) from `3` to make sure what we are doing is what is expected. We then start doing the juicy part:
1) We charge the payment option with the _calculated_ amount
2) We try to create the Printful order
3) If any error happens with the Printful order (response other than 2xx, an API error, or PHP error) we try to issue a refund to the stripe payment
4) If we fail on the refund then we log a PHP error and hope someone sees it `¯\_(ツ)_/¯`
On success we show a little "Order successful" message and a link to `1`
On failure we show the error message and a link to `1`

backend/emails/shipping_confirmation.html
```diff
@@ -0,0 +1,205 @@
...
+                    font-size: 15px !important;
+                }
+            }
+        </style>
+    </head>
+    <body style="max-width:580px;margin:0 auto;padding:20px;font-family:'Helvetica Neue', HelveticaNeue, Helvetica, 'Lucida Grande', arial, sans-serif;font-weight:300;font-size:14px;color:#222222;line-height:20px;">
+        
+        <div style="width:auto;margin:0 0 30px 0;" id="logo">
+            <a href="https://elementary.io" title="elementary">
+                <img src="https://elementary.io/images/logotype.png" height="28" alt="elementary" title="elementary">
```
lewisgoddard: Do we want this to be a tiny image in the corner and not full width?

backend/emails/shipping_confirmation.html
```diff
@@ -0,0 +1,205 @@
...
+
+<p style="font-weight:300;font-size:15px;color:#222222;line-height:20px;">
+    Here's what we shipped in this order:
+</p>
+
+<table width="100%" cellspacing="0" cellpadding="0">
+    <tbody>
+        <tr>
+            <td width="25%" valign="top" style="padding-top:10px;padding-bottom:10px;">
+                <img src="https://d1yg28hrivmbqm.cloudfront.net/files/6aa/6aa7c59b099906c64a9983a3b5377080_thumb.png" width="100%" alt="" style="background-color:#f0f0f0;">
```
lewisgoddard: This looks super-weird. Will this be a product image?

backend/emails/shipping_confirmation.html
danrabbit: Yeah I'd rather not have the logo too overstated

backend/emails/shipping_confirmation.html
danrabbit: Yeah I'm not sure what's the best way to get this but it should be the same image as the thumbnail in the store.

lewisgoddard:
.grid--address label {
        display: block;

is over-riding the `display: none;` being set by the `hidden` attribute.

btkostner:
Note before pushing to master:
- [ ] Update printful api hook endpoint to `elementary.io` vs `beta.elementary.io`

danrabbit:
It seems like things on the "Check Out" page are a little out of order

- [ ] The radios for selecting shipping method should probably be above the "ship to" and "estimated delivery" lines. In this way things kind of flow in order. You select a shipping method and then you get the estimated delivery date.

- [ ] Maybe the whole line item and total should be at the bottom of the page instead of the top. So you're selecting a shipping method first, then you're seeing the consequences of that (the estimated delivery date), and then you're reviewing the order and the final costs before you place it.

lewisgoddard:
I hereby approve this code.

PR #1256, opened 2016-08-16
Build

btkostner:
This adds a `gulpfile` to the project. It _is_ required for building assets, but it's rather light weight compared to some other processes I was looking into *wink* *wink* `webpack`.

Using it is as simple as `npm install` and `npm run build`. 

As of right now all it does is apply browser prefixes, but the world is at our finger tips. Just don't get too crazy.

But why do we need this? To be honest I wasn't thinking about `css` files, but this will be invaluable when doing the store v2 as most of it will be client side with `vue.js`. Plus, all the cool projects are using build tasks.


lewisgoddard:
The code here looks good, I will run a real-world test a little later, and then we need to talk about these server-side changes.

lewisgoddard:
I'm happy with this code, but I'm not sure of the state of server-side deployment right now. @btkostner has been doing it all, I guess it's best for him to deploy this..?

README.md
```diff
@@ -68,13 +68,17 @@ Then simply run `vagrant up`.
...
+* [Node.js](https://nodejs.org/)
 * `php7.0-curl`
 * `php7.0-intl`
 * `php7.0-json`
 * `php7.0-mbstring`
 * `php7.0-xml`
+* packages installed with `npm install`
 
-Then inside the project directory, run `php -S localhost:8000 router.php`. Next, just navigate to http://localhost:8000/ to view the site.
+Then inside the project directory, run `npm run build` to build the assets, and
```
lewisgoddard: You need to add `npm install` and `npm run build` to `CONTRIBUTING.md` as well.

PR #1259, opened 2016-08-19
Scripts clean

btkostner:
This seperates the `$page['scripts']` tag to three different tags:

- `$page['styles']` the most obvious one. This holds all the styles
- `$page['script-plugins']` loads any jQuery plugins after jQuery, but before any script that depends on jQuery.
- `$page['scripts']` loads all the other scripts. By default it loads everything async.

Other notables:
- changes cdn to `cdnjs.com` as it has a _very_ large library, is supported by cloudflare, and has all the plugins we need.


_templates/header.php
```diff
@@ -71,32 +76,18 @@
         <link rel="stylesheet" type="text/css" media="all" href="https://fonts.googleapis.com/css?family=Open+Sans:300,400,600,300italic,400italic|Droid+Sans|Roboto+Mono">
         <?php } ?>
 
-        <link rel="stylesheet" type="text/css" media="all" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.5.0/css/font-awesome.min.css">
+        <link rel="stylesheet" type="text/css" media="all" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.5.0/css/font-awesome.min.css">
```
lewisgoddard: 4.6.3?

_templates/header.php
```diff
@@ -71,32 +76,18 @@
...
-            <?php include __DIR__.'/../scripts/twitter-links.js'; ?>
-            <?php include __DIR__.'/../scripts/external-links.js'; ?>
-        </script>
-
-        <?php echo !empty($page['scripts']) ? $page['scripts'] : false; ?>
+        <?php foreach ($page['styles'] as $style) { ?>
+        <link rel="stylesheet" type="text/css" media="all" href="<?php echo $style ?>">
+        <?php } ?>
 
+        <script async src="https:<?php echo $sitewide['branch_root'] ?>backend/hsts.php"></script>
```
lewisgoddard: Won't this will block the pageload from firing for a few seconds in Midori?

_templates/header.php
btkostner: I didn't want to update it in this PR, but yes, it is available

_templates/header.php
btkostner: I can confirm this is not a problem with midori on 16.04

screenshot from 2016-08-20 09 56 50

PR #1260, opened 2016-08-20
use new grid for index

danrabbit:
Uses the new grid system for the home page. Didn't touch stuff for the carousel and below to avoid conflicts with Loki branch


btkostner:
much better at smaller sizes :+1:

PR #1272, opened 2016-08-24
update chartjs with cdnjs newest version

btkostner:
This removes the local copy of `chartjs` and replaces it with one from `cdnjs`. Sadly I couldn't find one that worked out of the box, so this also includes the code needed to update it to the latest version. Slight style changes did occur due to sizing changes.


lewisgoddard:
We seem to have started out with JSDelivr and are now gradually sneaking in CDNJS without me realising that we're mixing our CDNs. That's not very efficient with HTTP2 rising, and my favourite thing about JSDelivr is always loading the latest version so long as they have an update manifest for the file. We should probably make a decision about which we're using.

JSDelivr has automated updating (if you want it and they have a manifest), allow pull requests for new libraries, and are actually 3 CDNs behind the scenes, which gives us a little more redundancy. See http://labs.eustasy.org/cdns/compare-cdns

We currently use:

- JSDelivr: [2](https://github.com/elementary/mvp/search?utf8=%E2%9C%93&q=jsdelivr)
- CDNJS: [2](https://github.com/elementary/mvp/search?utf8=%E2%9C%93&q=cdnjs)

get-involved.php
```diff
@@ -7,12 +7,8 @@
         'styles/get-involved.css'
     );
     $page['scripts'] = array(
-        'scripts/Chart.custom.min.js' => array(
-            'async' => false
-        ),
-        'scripts/get-involved.js' => array(
-            'async' => false
-        )
+        'https://cdnjs.cloudflare.com/ajax/libs/Chart.js/2.2.1/Chart.min.js',
```
lewisgoddard: I've made a pull request for this to be added to JSDelivr: https://github.com/jsdelivr/jsdelivr/pull/13424

lewisgoddard:
Works for me.

screenshot from 2016-08-26 18 23 05

danrabbit:
Guys the burndown chart is burning up :p

PR #1275, opened 2016-08-24
WebTorrent as default download option.

lewisgoddard:
## WebTorrent

- Reduces reliance on load-balanced servers, potentially reducing cost, and increasing speed.
- Validates in the browser, so a step can be removed from "Installation".

---

### Changes Summary
- [x] Change default to a direct download.
- [x] Switch on modal to an auto-starting WebTorrent if available.
- [x] Implement save button for WebTorrents.
- [x] Move StreamSaver to JSDelivr (added in jsdelivr/jsdelivr#13597).
- [x] Move web-streams-polyfill to JSDelivr (added in jsdelivr/jsdelivr#13599).
- [x] Load libraries like [_scripts/lib/modal.js](https://github.com/elementary/website/blob/master/_scripts/lib/modal.js).
- [x] Fix web-stream-polyfill inclusion so Firefox can save.
- [x] Fix styles for all browsers.
- [x] Wait for fix for [Chromium Issue #375297](https://bugs.chromium.org/p/chromium/issues/detail?id=375297), which should be resolved with the release of Chrome 57, ETA: March 14th.
- [ ] Check streamSaver is saving when running WebTorrent.
- [ ] Pause or Stop WebTorrent when exiting the modal (see [this comment](https://github.com/feross/webtorrent/issues/580#issuecomment-183539789)).
- [ ] Check what happens when you use more RAM than you have, failover.
- [ ] Move all styles to stylesheets.

---

### Testing

This pull request is built at https://beta.elementary.io/webtorrent-2/
Skip 0 entry with `document.cookie = 'has_paid_Loki_0.4=1000'`
This request is auto-deployed for testing.

| Browser | [WebTorrent Supported](http://caniuse.com/#feat=rtcpeerconnection) | [StreamSaver Supported](https://github.com/jimmywarting/StreamSaver.js#supported-browsers) | StreamSaver Requires |
| ------- | ------------- | ----------------- | ----------------------- |
| Chrome  | Yes, from <51 | Yes, 52+          |                         |
| Firefox | Yes, from <46 | Yes, with Polyfill | Streams                 |
| Safari  | No            | No                | Streams, Fetch, SW      |
| Edge    | No            | No                | Streams, SW             |
| Opera   | Yes, from <42 | Yes, 39+          |                         |
| IE      | No            | No                | Everything (IE is dead) |

---

### Libraries
WebTorrent from https://github.com/feross/webtorrent
StreamSaver from https://github.com/jimmywarting/StreamSaver.js
Web Streams Polyfill from https://github.com/creatorrr/web-streams-polyfill


lewisgoddard:
screenshot from 2016-08-25 20 05 36
screenshot from 2016-08-25 20 30 43

lewisgoddard:
It appears we have something of a problem that until browsers themselves change how they handle things this branch will be blocked. Currently, most browsers max out blob files in the page at a few hundred Megabytes, as this mostly runs in memory. Because of this, Chrome returns a "file invalid" error when trying to save (it dumps the early data once it starts going over the limit), and Opera throws a "no file" error.

Only Firefox downloads and saves correctly, which is less than 20% of our users. Until the 70% that use Chrome work as intended, I can't recommend we include this as the new download mechanism.

---

| Browser        | Constructs as | Filenames    | Max Blob Size | Dependencies |
| -------------- | ------------- | ------------ | ------------- | ------------ |
| Firefox 20+    | Blob          | Yes          | 800 MiB       | None         |
| Firefox < 20   | data: URI     | No           | n/a           | [Blob.js](https://github.com/eligrey/Blob.js) |
| Chrome         | Blob          | Yes          | [500 MiB][3]  | None         |
| Chrome for Android | Blob      | Yes          | [500 MiB][3]  | None         |
| Edge           | Blob          | Yes          | ?             | None         |
| IE 10+         | Blob          | Yes          | 600 MiB       | None         |
| Opera 15+      | Blob          | Yes          | 500 MiB       | None         |
| Opera < 15     | data: URI     | No           | n/a           | [Blob.js](https://github.com/eligrey/Blob.js) |
| Safari 6.1+*   | Blob          | No           | ?             | None         |
| Safari < 6     | data: URI     | No           | n/a           | [Blob.js](https://github.com/eligrey/Blob.js) |

[Reference: FileSaver.js - Supported Browsers](https://github.com/eligrey/FileSaver.js/#supported-browsers)

[3]: https://code.google.com/p/chromium/issues/detail?id=375297

emersion:
One possible solution would be to use https://github.com/jimmywarting/StreamSaver.js#get-a-node-stream-from-webtorrent

Works in Chrome and Opera.

emersion:
For reference, the upstream issue is: https://github.com/feross/webtorrent/issues/319

lewisgoddard:
StreamSaver is a workaround, not a solution. For a file as large as ours that will have significant impact performance, saving will likely take many seconds, if not minutes.

emersion:
As far as I can tell, StreamSave is streaming, so the file is saved as long at it is downloaded. There is no "write file to disk" step.

lewisgoddard:
Also on hold pending Loki.

lewisgoddard:
I'm not entirely sure if stream saver isn't a better option for Firefox too, gets rid of the write-to-disk step that is present in webtorrent, but we need to add a check if streamSaver was cancelled or not.

emersion:
1. `streamSaver` is a better option for all platforms, as the file is written while it's been downloaded
2. `streamSaver` is not supported on Firefox yet
3. Not sure we can detect if the download was cancelled. The only way I an think of is to detect if the HTTP request is cancelled in the service worker.

emersion:
Additionally, `streamSaver` is not available on jsDelivr, is there a way to add it there?

lewisgoddard:
1. I agree.
2. Is there an ETA for streams in firefox?
3.Right now WebTorrent just carries on even if it's not saving in Chrome, this is kind of key.
4. I will propose streamSaver to JSDelivr.

emersion:
2.See https://bugzilla.mozilla.org/show_bug.cgi?id=1128959 and https://bugzilla.mozilla.org/show_bug.cgi?id=1283191

lewisgoddard:
So that first issue is 18 months old, and the person working on it recently stopped due to work constraints. I'm guessing we should get used to how it is now.

feross:
FYI Chrome finally fixed the 500MB blob size limitation! https://bugs.chromium.org/p/chromium/issues/detail?id=375297#c107

lewisgoddard:
Thanks for the nudge!

> The ETA of the stable Chrome 57 is March 14th.

emersion:
Note: there might be some issues on systems with low RAM. All the ISO is saved to RAM, regardless of streamSaver being enabled or not, because webtorrent needs access to the whole file in-memory to be able to send parts of the file to other peers.

Here are some stats about users' RAM: https://metrics.mozilla.com/firefox-hardware-report/#detail-ram. It seems that there are still a lot of people (20%) having 2GiB of RAM, which is not a lot when loading 1GiB in RAM.

1. Look what happens when user runs out of RAM (setup a VM with 1GiB and test)
2. See if it's possible to query the available RAM (seems not possible)
3. See if it's possible for webtorrent to only keep a random part of the file in RAM, and only send this part to other peers

feross:
You can use a custom `store` option and specify https://www.npmjs.com/package/idb-chunk-store as the store and IndexedDB will be used instead of in-memory.

danrabbit:
Gonna close this since there's been no progress in a couple years now

Issue #1282, opened 2016-08-25
Organise folders and files and document their usage.


lewisgoddard:
Currently the folder system in this website is a little muddled and could do with being tidied.
### Not externally accessible.

Any file or folder prefixed with a `.` or a `_` is not externally accessible.
- [x] `.github` is for special GitHub files only.
- [x] `.tx` is for special Transifex files only.
- [x] `_backend` is for storing backend files and libraries.
- [x] `_data` is for private data that is handled by some part of the site before being exposed.
- [x] `_lang` (previously `lang`) contains the translated Transifex files.
- [x] `_scripts` is for raw javascript, before being processed.
- [x] `_styles` is for raw stylesheets, before being processed.
- [x] `_templates` is for reusable sections of HTML.
- [x] `_tests` is for tests.
### Externally accessible.
- [x] `api` (previously `hook`) should contain hooks hit by other services, like printful, or ourselves.
- [x] `data` is for public data, like the slingshot json. This is loaded over HTTP.
- [x] `docs` is a site run by MDR, and should otherwise only contain `*.md` files.
- [x] `fonts` contains fonts we rely on.
- [x] `images` is for images.

btkostner:
I would also like to add, that we should create an `api` folder, and move `backend` to `_backend`. This should be a lot easier to understand for new people contributing as everything in the `_backend` directory would be pure PHP functions, no echos or outputs. 

Things that we can include right now
- `api/configuration` for sending site variables like the release and stripe public key. Moving them away from inline `script` tags
- `store/inventory` -> `api/store/inventory` for setting store cookies
- `api/stripe/payment` for making payments

btkostner:
I would like to make note that we probably don't want to move `lang` to `_lang` as it may come in handy for translating client side code unless we want to do an api endpoint (probably unnecessary).

btkostner:
And here are my future plans for the `_scripts` folder in a bit more detail

- `_scripts/helpers` is for small snippets used in multiple parts of code
- `_scripts/lib` is for any and all external loaded libraries
- `_scripts/pages` is for all compiled files. One per page with the exception of `main.js`
- `_scripts/widgets` is for all small reusable components

PR #1294, opened 2016-08-27
Webpack (War on Globals)

btkostner:
I'm sorry that git has such a hard time understanding the `mv` command. Even after `git mv` and `git commit`. 

**Things that `git` sucks at detecting:**
- Files being moved from `scripts` to `_scripts`
- Adding header to all JS files so people have an idea of what they do
- Indenting most of the code because of a single line addition

**The major changes in this PR:**
- The `scripts` directory was moved to `_scripts`
- Added webpack to build script files
- Move away from external scripts served in HTML and do it in JS

**Webpack:**
- Moved away from `gulp` to `webpack` so we can bundle javascript. This means you can now use `include jQuery from 'lib/jquery'`. +3 to charisma
- Added babel for transcoding. We can now do proper classes and use a cleaner syntax. +2 to charisma
- Simplified the CSS building and testing

**What is this external script loading thing?**
So instead of having an HTML document that would load `jQuery` and `Chart.js` and a bunch of other dependencies, we load a single (hopefully) file, and that JS file is responsible for loading what it needs. This should help a lot with modularity and expansion.

**Other small changes:**
- Added a `main.js` page that loads all of those pesky snippets we have like `smooth-scrolling` and `external-links`. -3 HTTP calls
- Moved `highlight.js` to jsdelivr CDN.
- Added a `npm run start:watch` command that launches PHP server and runs `build:watch`


lewisgoddard:
Is the build command still `npm run build` or is webpack a separate command?

lewisgoddard:
Also, git seems to have understood fine https://github.com/elementary/mvp/pull/1294/commits/dc986f5d64565916b679e4554013db18a9ab22da

btkostner:
All the commands are the same
- `npm start` for php server 
- `npm install` for deps
- `npm run build` for building
- `npm run build:watch` for building on new things 

New command
- `npm run dev` starts php server and `build:watch`

btkostner:
Considering we are nearing Loki release. Please don't merge this until after.

lewisgoddard:
Ping #1309

btkostner:
Alright this should be good and ready. Sorry if I overstepped a bit while cleaning up the code. There is more changes than what I probably should have done, but on the bright side, I got a head start on #1282. 

Some of the changes:
- I added an `api/config.php` file that outputs JSON variables used throughout the site, so no more inline script variables (except on the error pages).
- I cleaned up the `_scripts` folder a bit (needs more) by stuffing all the compiled scripts to `_scripts/pages`
- Moved Google analytics to a lib just like everything else
- Cleaned up some PHP headers while I was removing script tags

I would recommend double checking and going around the pages with dev console up and making sure there are no errors

_scripts/lib/jquery.js
```diff
@@ -0,0 +1,16 @@
...
+ * Loads jquery from cdn address
+ *
+ * @exports {Promise} default - a promise of the jQuery library
+ */
+
+import Promise from 'core-js/fn/promise'
+import Script from 'scriptjs'
+
+export default new Promise((resolve, reject) => {
+    Script('https://cdn.jsdelivr.net/jquery/3.1.0/jquery.min.js', () => {
```
lewisgoddard: Most of these are in the format `https://cdn.jsdelivr.net/g/jquery@3.1.0`

btkostner: Fixed

PR #1300, opened 2016-08-28
make the countdown a bit more responsive

btkostner:
This makes the countdown a bit more responsive by collapsing at widths smaller than 800px.

Why is this a PR?
Because we are close to pushing the countdown branch and I didn't want to break anything before release. This does edit the normal flipclock.js DOM, which (from what I could find) is not supported, so this needs some testing just in case.

### DIFF
--- a/_styles/countdown.css
+++ b/_styles/countdown.css
@@ -21,6 +21,7 @@
 .countdown-wrapper {
     overflow: auto;
     padding: 16px;
+    width: 100%;
     text-align: center;
     user-select: none;
 }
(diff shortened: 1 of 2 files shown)


danrabbit:
Can confirm that this reflows to a vertical layout on mobile.

However, it seems like we a small problem which is that scrolls pass through to the website underneath unless you're dragging one of the numbers. This makes the "continue" button fairly inaccessible. I'm actually having trouble activating the button altogether, especially since the bottom toolbar in safari comes up and hides it when you tap the page. This anchor probably needs to be made a much larger tap target (with padding)

btkostner:
@danrabbit try the padding now. I used `30px` but it still seems in the same spot. I can increase it if need be.

danrabbit:
Yay works for me :)

PR #1303, opened 2016-08-29
Fix #1075: Add print-friendly CSS

### LINKED ISSUE #1075: Style printed documents better
Got this idea from reading this Reddit post https://www.reddit.com/r/elementaryos/comments/49715h/pdf_version_of_the_elementary_os_developer/

This might be a good idea to automatically generate PDF versions of our docs (user guide, dev guide, HIG, etc). Of course it's important that users have the latest versions of these docs, but being able to get a downloadable copy might be good.

lewisgoddard:
Fixes #1075

### Changes Summary

- Remove header from printed documents.
- Remove navigation from printed documents.
- Remove footer from printed documents.
- Add a `.dont-print` class.

### DIFF
--- a/_styles/main.css
+++ b/_styles/main.css
@@ -1097,7 +1097,8 @@ i.fa.fa-close.close-modal {
 @media print {
     header,
     nav,
-    footer {
+    footer,
+    .dont-print {
         display: none;
     }
 }
--- a/_styles/main.css
+++ b/_styles/main.css
@@ -1093,3 +1093,11 @@ i.fa.fa-close.close-modal {
     display: block;
     margin: 0 auto;
 }
+
+@media print {
+    header,
+    nav,
+    footer {
+        display: none;
+    }
+}


btkostner:
Do we want to limit this to just the docs? I'm not sure how applicable printing the homepage and other pages are (maybe brand).

Also I wouldn't consider this a complete fix, there are still some places that look bad (code blocks) that should be fixed before this issue is completely fixed.

lewisgoddard:
I don't see why you'd ever want the header, footer, or navigation. They are mostly just links to other pages, which make no sense on paper.

Code blocks look bad for you? The look sans-color to me (translates far better without backlight).

btkostner:
It's better with the no background without a doubt, but the border is a bit weird.
screenshot from 2016-08-29 13 01 57


Not high priority though, so I'm ready to merge as is if you are happy with it.

lewisgoddard:
I'd prefer to merge this, it's simple and shows significant improvement. The border is a little weird but renders differently based on the browser, it's probably best to let them use whatever defaults the user sets and only override to make smart decisions.

PR #1316, opened 2016-09-02
Auto-select country entry.

### LINKED ISSUE #1265: Use an autocompleting entry for Country
Using a dropdown for country is kind of painful. If would be nicer if it were an entry with autocompletion. We could use this: https://github.com/JamieAppleseed/selectToAutocomplete

lewisgoddard:
Improves upon #1265 Use an autocompleting entry for Country

### Changes Summary

- Detect the country.
- Select it automatically.
- Trigger an update.

### DIFF
--- a/backend/classify.functions.php
+++ b/backend/classify.functions.php
@@ -130,3 +130,30 @@ function ipHash($hostname, $debug = false) {
     }
 }
 
+function getCurrentCountry($hostname, $debug = false) {
+
+    try {
+        if ( $debug ) {
+            echo $hostname."\n";
+        }
+        if (!class_exists('GeoIp2\Database\Reader')) {
+            throw new \Exception('Class GeoIp2\Database\Reader not found');
+        }
+        $reader = new Reader(__DIR__.'/GeoLite2-City.mmdb');
+        $record = $reader->city($hostname);
+        if ( $debug > 1 ) {
+            var_dump($record);
+        }
+        $country   = $record->country->isoCode;
+
+    } catch (\Exception $e) {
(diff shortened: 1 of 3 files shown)


lewisgoddard:
How do we want this to deal with no country detected (hardly ever happens). Currently defaults to the first in list.

backend/classify.functions.php
```diff
@@ -130,3 +130,30 @@ function ipHash($hostname, $debug = false) {
...
+        }
+        $country   = $record->country->isoCode;
+
+    } catch (\Exception $e) {
+        echo '<!-- '.$e->getMessage().' -->'."\n";
+        $country   = false;
+    }
+
+    if ( $debug ) {
+        echo 'Country: "'.$country.'"'."\n";
```
btkostner: Why are we echoing data? Isn't it possible to leak sensitive data in the error?

lewisgoddard: @btkostner That file has always done that, and only with the debug mode enabled, it's what you see at https://elementary.io/backend/classify.get.php

There's no sensitive data there, only the individuals detected country/region.

danrabbit:
If we're using an entry instead of a combobox, it should be blank so users can see the hint text.

If this branch doesn't intend to use the entry and is just about automatically grabbing the country, I would default to whatever our country with the most visits is (I'm gonna guess US or maybe Brazil)

lewisgoddard:
@danrabbit This isn't an entry, just an auto-selector, so keep US?

lewisgoddard:
@danrabbit I've modified the title for you.

PR #1320, opened 2016-09-02
switch ko to ko_KR translations

btkostner:
As per Jung-Kyu Park's request, we should be using the more complete `ko_KR` translations instead of the current `ko` translations.

Not sure if there is anything else we need to do

### DIFF
--- a/_backend/l10n.php
+++ b/_backend/l10n.php
@@ -22,7 +22,7 @@ class Translator {
         'id_ID' => 'Bahasa Indonesia',
         'it' => 'Italiano',
         'ja' => '日本語',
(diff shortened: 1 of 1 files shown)


emersion:
We tend to prefer language codes without region codes. It would be better to delete `ko`, replace it by `ko_KR`.

lewisgoddard:
@emersion Do we know if we can do that in Transifex easily? Otherwise, this is our next-best option.

btkostner:
Unless I'm missing something in the Transifex UI, it is not possible to easily switch the language code.

Issue #983, opened 2016-01-11
Solarized stylesheet is different from desktop


danrabbit:
all the yellow stuff should be red (like "var")

danrabbit:
Alright to get more specific, here's what I notice that needs to be changed:

- [ ] `public`, `class`, `protected`, `override`, `static`, `var`, `new`, `return` are green. Should be red.
- [ ] `void`, `int`, `string` are green. Should be yellow 
- [ ] `ref` is not highlighted. Should be red
- [ ] `false` is not highlighted. Should be green
- [ ] `stdout

PR #1322, opened 2016-09-02
Fix #1321: Use hljs from CDN

### LINKED ISSUE #1321: update hljs
We're several versions behind. Might incidentally fix #983

emersion:
- [x] Solarized dark has solarized light theme for line numbers
- [x] ~~Line numbers broken in `/developers`~~

Fixes #1321


developer.php
```diff
@@ -5,15 +5,14 @@
...
+        'https://cdn.jsdelivr.net/g/jquery.leanmodal2@2.5,highlight.js@9'
     );
     $page['scripts'] = array(
-        'scripts/highlight.pack.js',
         'scripts/developer.js'
     );
     $page['styles'] = array(
         'styles/developer.css',
-        'styles/solarized_light.css'
+        'https://cdn.jsdelivr.net/highlight.js/9.*/styles/solarized-light.min.css'
```
lewisgoddard: Alternatively you can use a URL to match the existing format: https://cdn.jsdelivr.net/g/highlight.js@9(styles/solarized-light.min.css)

danrabbit:
Hm, dark theme means Terminal so it shouldn't have line numbers at all. Probably the API for choosing line number or not changed?

emersion:
Ahah, so actually the problem might be that the dark theme is always used, for both bash and vala.

danrabbit:
It looks like the language isn't being detected correctly (it thinks it's a different language in like every snippet). I don't see a vala style upstream in hljs. Did we carry our own highlighting scheme for this?

emersion:
So we basically created our own theme to highlight bash with solarized dark. We may want to keep themes hosted in the repo, especially if we want to customize it (e.g. to fix #983 which is not fixed with newer themes).

danrabbit:
Hm I think we're missing something. I think maybe we added a custom set of highlight rules.

In this branch, on developer page it says the language is "cpp". In master it says "vala"

emersion:
Okay, the problem here is that hljs from the CDN is missing vala.

danrabbit:
Yeah so I guess this is blocked then. We need to fix #983 and upstream the vala highlight mode

emersion:
No, we can upstream the vala highlight mode. hljs from the CDN is only including 22 languages, because including more would make the library too big. So we need to continue to host it.

lewisgoddard:
Easy fix! Use both the `highlight.js` and the language file. Which do we need?

Like this: `https://cdn.jsdelivr.net/g/highlight.js@9.6.0(highlight.min.js+languages/vala.min.js)`

lewisgoddard:
Works for me, maybe @danrabbit can just double-confirm that this works. It highlights the vala on `/developer` fine here though.

danrabbit:
Yeah this is pretty much the same as master. I think there's some issues with our solarized-light styles and the vala highlighter is not 100% correct. But it's parity with master, so it works for me :)

If we're using these things from CDN, we should delete the local copies

lewisgoddard:
`scripts/highlight.pack.js` has been deleted already.

PR #1335, opened 2016-09-05
fix configuration backend loader for new beta site

btkostner:
This fixes the configuration loading for the new https://beta.elementary.io site

### DIFF
--- a/backend/config.loader.php
+++ b/backend/config.loader.php
@@ -3,9 +3,9 @@
 if ( is_readable(__DIR__.'/config.php') ) {
     // for configured local
     require_once __DIR__.'/config.php';
-} else if ( is_readable(__DIR__.'/../../_backend/config.php') ) {
+} else if ( is_readable(__DIR__.'/../../master/backend/config.php') ) {
     // for configured hosted
-    require_once __DIR__.'/../../_backend/config.php';
+    require_once __DIR__.'/../../master/backend/config.php';
 } else {
     // for un-configured local
     require_once __DIR__.'/config.example.php';
--- a/backend/config.loader.php
+++ b/backend/config.loader.php
@@ -1,15 +1,12 @@
 <?php
 
-if (
-    !empty($_SERVER['REQUEST_URI']) &&
-    substr($_SERVER['REQUEST_URI'], 0, 8) == '/branch/'
-) {
-    // for Branches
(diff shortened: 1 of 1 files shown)


fabianthoma:
Just to clarify, this is going to use the same config for all of the branches now, right? even master?
Because from what I understand of gits it does deploy all branches to one directory, right?

btkostner:
That is correct. If we want we can use a different config for master we can just add the `config.php` file in master's `backend` folder, but I'm not sure that we need to.

lewisgoddard:
This won't work until #1282 is fixed, will it?

backend/config.loader.php
```diff
@@ -1,15 +1,12 @@
...
-    substr($_SERVER['REQUEST_URI'], 0, 8) == '/branch/'
-) {
-    // for Branches
-    require_once __DIR__.'/../../../backend/config.php';
-} else if ( is_readable(__DIR__.'/config.php') ) {
-    // for Configured MASTER
+if ( is_readable(__DIR__.'/config.php') ) {
+    // for configured local
     require_once __DIR__.'/config.php';
+} else if ( is_readable(__DIR__.'/../../_backend/config.php') ) {
```
lewisgoddard: Shouldn't this now be `/../../master/_backend/config.php` ?

btkostner: I put the `_backend` folder outside of branches, so the tree looks like this:
```
_backend
  config.php
master
  _backend
config-loader
  _backend
```

btkostner:
`_backend` is outside of a branch, so it's not related to #1282

backend/config.loader.php
lewisgoddard: I feel like you're duplicating files too much. What would be different in the `_backend` folder than the `master/_backend` folder? Why not just have branches fall back to master?

btkostner: The only problem I could possibly see with that is if we ever wanted to have the beta branches use a different config than master. Otherwise I'm good with using the configuration file in master.

lewisgoddard: I would have thought that 99% of the time they'd be the same, and we'd just override the branch we want different.

btkostner: Updated to use `master/backend/config.php`. Will need to be updated when #1282 is being worked on.

PR #1338, opened 2016-09-06
Fix #1334: Use a button to show the countdown overlay

### LINKED ISSUE #1334: Use a FAB to hide/show the countdown overlay
People are confused when the countdown is gone and they can't find a way to bring it back, since there's no way to access it again anywhere on the site. A Floating Action Button would probably be a good way for users to access the countdown.

lewisgoddard:
Fixes #1334: Use a button to show the countdown overlay

We currently have a `/countdown` page, but nothing links to it.

### DIFF
--- a/_templates/header.php
+++ b/_templates/header.php
@@ -146,6 +146,12 @@ $l10n->begin_html_translation();
                     <?php } ?>
                 </ul>
                 <ul class="right">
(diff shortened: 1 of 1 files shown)


danrabbit:
Simple. Subtle. Cute. I like it.

PR #1339, opened 2016-09-08
OS X on the name change of the current

igor-dyatlov:
- [x] Homepage
- [ ] `/docs/installation`

### DIFF
--- a/index.php
+++ b/index.php
@@ -50,7 +50,7 @@
                     ?>
 
                 </div>
-                <h4>A fast and open replacement for Windows and OS X</h4>
+                <h4>A fast and open replacement for Windows and macOS</h4>
             </div>
         </section>
         <div class="hero"></div>


cassidyjames:
This sounds good to me now that macOS has a release date, though I'm not sure if we want to wait for that date or switch now.

danrabbit:
Apple has macOS on their website now. I'm okay with whatever

lewisgoddard:
I guess they can retrospectively apply macOS to older versions, but the first true macOS release comes out on the 20th of September.

cassidyjames:
I'm good with merging. @igor-dyatlov can you update this from master and then we can get it merged?

png2378:
Please, don't break translations before the release :worried:

lewisgoddard:
Yeah, I'd prefer to block this from `master` for a couple of weeks. It's really not a major change.

lewisgoddard:
Also, this needs to be updated in some of the docs, notably installation.

lewisgoddard:
Closed in favour of #1348

PR #1342, opened 2016-09-08
fix homepage title for social previews

### LINKED ISSUE #1340: Change home page title
Maybe better to change the page title to the tagline "A fast and open replacement for Windows and OS X" since this shows up on social media previews and such

btkostner:
Fixes #1340

### Changes Summary

- Do not update the page title with JS
- Update page title to "A fast and open replacement for Windows and OS X"


lewisgoddard:
This will break the translation on the page titles.

btkostner:
@lewisgoddard seems to be working as intended https://beta.elementary.io/homepage-title/es/

lewisgoddard:
Huh, I didn't think that worked for title tags.

lewisgoddard:
I'm fairly certain it will be bad for SEO for us not to have "elementary OS" in the title tag.

btkostner:
Worth noting that because I added the `elementary` to the title we have no translations for it

lewisgoddard:
Should it be "elementary" or "elementary OS"?

The OS is what the description is describing.

btkostner:
I just copied the ending we use on all the other pages:

Support: `Support &sdot; elementary`
Store: `Store &sdot; elementary`
Developer: `Developer &sdot; elementary`

lewisgoddard:
The store, support, and developer are about things other than the OS, but the homepage is currently OS only. I feel like it might help with SEO, our rankings are a little tentative.

PR #1349, opened 2016-09-09
Fix #903: Never have to modify the links directly again

### LINKED ISSUE #903: Do not hardcode release variables.

lewisgoddard:
Fixes #903: Do not hardcode release variables. 

### Changes Summary

- [x] Magnet Hash
- [x] Chart Milestone


btkostner:
Continuing with this, we should move some other variables to the `config.php` page:

- sha sums
- chart milestone (hide it when `!isset`)

Possibles:
- Would it be too specific if we had a beta section in the `config.php` file for use in the developer header?

lewisgoddard:
There is a caveat. Docs can't use PHP, so anything there has to be re-written each time. No variables.  This means shasums are out.

PR #1358, opened 2016-09-10
Translation update

btkostner:
Updates translations and adds open-source page and some store pages to transifex


btkostner:
Needs another update

PR #1362, opened 2016-09-11
Explicitly state you can enter zero.

lewisgoddard:


lewisgoddard:
even0

danrabbit:
This is a bad idea and the only thing I can see happening is us going out of business.

PR #1365, opened 2016-09-12
rewrite download dialog

### LINKED ISSUE #1352: Download dialog is bad
Super helpful bug title I know. Okay here we go!

- [x] doesn't need an 'x' button since it has "Cancel"
- [x] 'Loki' is not a great button label. Should be an explicit action like "Download"
- [x] Buttons should be right aligned not center

Just check out this guy for reference: https://elementary.io/docs/human-interface-guidelines#dialogs

- [x] A good icon would be apps/48/ubiquity.svg

danrabbit:
Fixes #1352 

screenshot from 2016-09-12 11 44 26

Please double check responsiveness and that it doesn't break docs/human-interface-guidelines#dialogs


index.php
```diff
@@ -369,14 +369,15 @@
...
-                </div>
-                <div class="column linked">
+        <div id="download-modal" class="dialog">
+            <img alt="Download elementary OS icon" src="images/icons/apps/48/ubiquity.svg">
+            <div class="content-area">
+                <p class="primary">Choose a Download</p>
+                <p>Download from a localized server or by magnet link. For help and more info, see the <a class="read-more" href="docs/installation" target="_blank">installation guide</a></p>
+            </div>
+            <div class="action-area">
+                <a class="button close-modal" style="margin:0;" href="#">Cancel</a>
```
lewisgoddard: Because this has `href="#"`, it scrolls the page to the top. Leave it blank and add the class `clickable` to cure.

lewisgoddard: Also, inline styles.

_styles/main.css
```diff
@@ -942,6 +942,79 @@ input:focus:-ms-input-placeholder {
...
+
+.dialog .action-area {
+    display: -ms-flexbox;
+    display: flex;
+    -ms-flex-flow: row;
+    flex-flow: row;
+    -ms-flex-pack: end;
+    flex-wrap: wrap;
+    justify-content: flex-end;
+    margin-top: 12px;
```
btkostner: also remove all the browser prefixes here. we automatically add them during build.

PR #1369, opened 2016-09-12
lower shirt price

danrabbit:
Velocity is kind of low on these right now. After some discussion, we think we should lower the price

### DIFF
--- a/data/store.json
+++ b/data/store.json
@@ -112,35 +112,35 @@
                 "size": "S",
                 "color": "Tri-Black",
                 "cost": 18.50,
-                "price": 28.00
+                "price": 25.00
             }, {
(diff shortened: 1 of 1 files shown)


data/store.json
```diff
@@ -112,35 +112,35 @@
...
-                "price": 28.00
+                "price": 25.00
             }, {
                 "id": 5871,
                 "name": "2XL Track Men's",
                 "size": "2XL",
                 "color": "Tri-Black",
                 "cost": 20.25,
-                "price": 30.00
+                "price": 36.00
```
lewisgoddard: Is this price supposed to be going up, or was it supposed to be `26.00`?

PR #1371, opened 2016-09-13
update nginx config with one used in production

btkostner:
This updates the nginx example file with the one currently in production minus the SSL configuration.

We might want to consider adding the SSL configuration here, but I'm not sure.


lewisgoddard:
This probably shouldn't be merged until after elementary/ansible#53 is because there might be more changes.

PR #1379, opened 2016-09-17
Update README.md

btkostner:
Let's add all the badges we can to the readme


codygarver:
I think the bountysource badge should be the first on the row of badges

PR #1391, opened 2016-09-22
nodejs-legacy instead of nodejs

danrabbit:
npm run build fails with nodejs package. Needs nodejs-legacy


danrabbit:
I'm on Loki and installing nodejs didn't work. Had to install nodejs-legacy

btkostner:
Hmm, I'm using `nvm` and am too lazy to uninstall everything to test it. I'll take your word for it.

PR #1392, opened 2016-09-22
Adding the coffee mug to the store

danrabbit:
Please double check product IDs. They could be totally wrong :p


data/store.json
```diff
@@ -191,5 +191,27 @@
...
+        "name": "Coffee Mug",
+        "description": "The elementary 'e' logo sublimated onto an 11oz, dishwasher and microwave safe coffee mug.",
+        "image": "images\/store\/accessories\/mug_11oz.png",
+        "type": "Accessories",
+        "variants": [
+            {
+                "id": 1320,
+                "name": "11oz Coffee Mug",
+                "size": "11oz",
+                "color": "Black",
```
btkostner: sure we don't want the "White Glossy Mug 11oz" to be white colored?

data/store.json
```diff
@@ -191,5 +191,27 @@
                 "url": "https:\/\/elementary.io\/images\/store\/designs\/logotype_white.psd"
             }
         ]
+    },
+    {
+        "id": 1320,
```
btkostner: The product id should be `19`

data/store.json
danrabbit: Isn't this internal? My rational is that it's the color of the print. So if we added a "blue" variant it would be the color of the print that changes and not the mug (since the mug can't change color)

btkostner: It is internal. Printful api only uses the `id`. Just wanted to make sure.

PR #1399, opened 2016-09-25
Sentry

btkostner:
Adds sentry error tracking to server side PHP. Most of this is the sentry library.


lewisgoddard:
Maybe now would be a good time to drop those libraries in the repository and let composer just pull them in?

btkostner:
@lewisgoddard works for me. I'll open another PR to remove the libraries

PR #1411, opened 2016-09-28
Fix workspace overview shortcut

png2378:
Replacing Cmd+S instead of Cmd+A for workspace overview.


png2378:
Sorry, this is my offence.

PR #1423, opened 2016-10-01
add some warning messages for legacy browsers

### LINKED ISSUE #1393: Redirect everything to HTTPS
### Expected behavior

Site is browsed only via HTTPS.

### Actual behavior

Site is accessible via HTTP to allow support for Midori.

btkostner:
Fixes #1393

This should take care of the website side for http -> https redirects issue.

- Adds a message when you browse the website with `http` (should never happen in production)
- Keeps the original IE upgrade message, but applies to anything less than IE11
- Adds a message targeting _specifically_ the version of midori in Freya. (needs copy)

All of the regex should be fairly strict, but it doesn't hurt to test to make sure we are targeting what we need.


_templates/legacy.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+/**
+ * _templates/legacy.php
+ * Holds all legacy browser messages warnings
+ */
+
+$l10n->set_domain('layout');
+
+$user_agent = $_SERVER['HTTP_USER_AGENT'];
```
lewisgoddard: Sometimes the user agent is not set.

_templates/legacy.php
```diff
@@ -0,0 +1,43 @@
...
+
+$is_HTTP = (empty($_SERVER['HTTPS']) || $_SERVER['HTTPS'] !== 'on');
```
lewisgoddard: This is only supposed to check for a non-empty string or "off". Some servers use "On", or simply `true` to mean it is on.

_templates/legacy.php
```diff
@@ -0,0 +1,43 @@
...
+$is_old_IE = preg_match('/MSIE [2-10]/i', $user_agent);
+
+if (getenv('PHPENV') === 'production' && $is_HTTP) {
```
lewisgoddard: Will this ever be true now we have a server-side redirect?

btkostner: Not on our production servers, but I felt it was needed in case someone else trys to host, or if we setup a new server that does not have the certificate generated yet.

PR #1430, opened 2016-10-04
cursory typos, concision, clarity

owcz:
### Changes Summary

Only changed what I thought was the most egregious as I read. Rarely ever a need to use "allow" instead of "let", or "utilize" instead of "use". Happy to help more if wanted


docs/human-interface-guidelines.md
```diff
@@ -90,7 +90,7 @@ Most users don't want to read through help docs before they can use your app. In
 
 ### Use Understandable Copy {#use-understandable-copy}
 
-Avoid technical jargon and assume little-to-no technical knowledge. This allows your app to be "self-documenting."
+Avoid technical jargon and assume little-to-no technical knowledge. Thus, your app becomes "self-documenting."
```
danrabbit: Not a fan of this change tbh. Just doesn't sound natural

docs/human-interface-guidelines.md
```diff
@@ -266,64 +266,64 @@ For more info, see the FreeDesktop.Org [menu entry](http://standards.freedesktop
...
+Keywords should be single words (in [title case](#title-case)) separated by semicolons.
 
 --------------------------------------
 
 See also: [Desktop Entry Specification](http://standards.freedesktop.org/desktop-entry-spec/latest/index.html) from FreeDesktop.org
 
 ## Contractor {#contractor}
 
-Contractor is a service and a protocol for exposing services easily between apps. It allows an app to interact with various other apps/services without hardcoding support for them. You simply add [contractor](https://launchpad.net/contractor) support, and then any service registered with contractor is now available for your app to use. Your app can integrate with contractor in two different ways:
+Contractor is a service and a protocol for exposing services easily between apps. It lets apps interact with other apps and services in lieu of hardcoded support. You simply add [Contractor](https://launchpad.net/contractor) support, and then any service registered with Contractor is now available for your app to use. Your app can integrate with Contractor in two different ways:
```
danrabbit: I dislike "in lieu of". I think this is way outside the voice and the other wording was more clear.

docs/human-interface-guidelines.md
```diff
@@ -266,64 +266,64 @@ For more info, see the FreeDesktop.Org [menu entry](http://standards.freedesktop
...
 
-* Register one of it's functions as a service that can be used from other apps
+* Register one of its functions as a service that can be used from other apps
 * Implement the contractor menu or otherwise access contractor to receive a list of services that can handle the data your app manages
 
 ### Displaying Results from Contractor {#displaying-results-from-contractor}
 
-The most typical way to present Contractor results to users is in menu form. Keep in mind the following things when presenting contractor results:
+Contractor results are typically presented to users in menu form. Keep the following in mind when presenting Contractor results:
```
danrabbit: Ah this is way better. This was a train wreck before. Thanks :)

docs/human-interface-guidelines.md
```diff
@@ -266,64 +266,64 @@ For more info, see the FreeDesktop.Org [menu entry](http://standards.freedesktop
...
 
 * If the item acted upon is one of many items visible, such as in an icon view, be sure to display results in a way that is directly related to the item such as in a context menu.
 * If the item acted upon is the only item currently visible, such as a web page or a contact in Dexter, contractor can be displayed in a toolbar menu.
 * Never display the target app's name. The menu string is the intended display name for that app. Doing so can lead to redundant messages such as, "Set image as wallpaper (wallpaper)" and it is always irrelevant to a user's needs.
 
 ## Dock Integration {#dock-integration}
 
-By integration with Pantheon's dock, apps are able to communicate their status to the user at a glance.
+Apps integrated with Pantheon's dock communicate their status to the user at a glance.
```
danrabbit: I think this change makes it too passive. I think the previous wording was more of a call to action for developers to add this integration instead of talking about what some other apps already do.

docs/human-interface-guidelines.md
```diff
@@ -266,64 +266,64 @@ For more info, see the FreeDesktop.Org [menu entry](http://standards.freedesktop
...
+A badge shows a count of actionable items managed by your app. Its purpose is to inform the user that there are items that require user attention or action without being obtrusive. This is a passive notification. A badge should not show totals or rarely changing counters. If the badge is not easily dismissed when the user views your app, it is likely that this is not a good use of a badge.
 
-* **Good Example**: Unread messages in Geary
+* **Good Example**: Unread messages in Mail
 * **Bad Example**: Total number of Photos in Shotwell
 
 ## System Indicators {#system-indicators}
 
-Indicators are small icons that live on the top panel. They give users a place to glance for a quick indication of various settings or events. Clicking the icon shows a small menu with related actions available to the user.
+Indicators are small icons that live in the menu bar (wingpanel). They give users a place to glance for a quick indication of various settings or events. Clicking the icon shows a small menu with related actions available to the user.
```
danrabbit: We don't refer to the panel as the "menu bar". I think that's a Mac-ism.

docs/human-interface-guidelines.md
```diff
@@ -266,64 +266,64 @@ For more info, see the FreeDesktop.Org [menu entry](http://standards.freedesktop
...
 
 ![](/images/docs/human-interface-guidelines/system-indicators/systray.png)
 
-### Does Your App Need An Indicator? {#does-your-app-need-an-indicator}
+### Does Your App Need an Indicator? {#does-your-app-need-an-indicator}
 
-The indicator area is notorious for becoming cluttered and having inconsistent behavior. Given that users will probably install many third party apps, we need to be careful about how many indicators we're showing and how they behave. Keep in mind that not every application needs an indicator. Only a very small set of apps will benefit from one. You do not need an indicator if:
+The indicator area is notorious for clutter and inconsistent behavior. Given that users will probably install many third-party apps, we must be careful about the number of indicators we show and how they behave. Keep in mind that not every application needs an indicator. Only a very small set of apps will benefit from one. You do not need an indicator if:
```
danrabbit: I think "becoming cluttered" is necessary because in its default state it is not cluttered. It only becomes cluttered when users install apps that overuse it.

docs/human-interface-guidelines.md
```diff
@@ -332,27 +332,27 @@ The indicator area is notorious for becoming cluttered and having inconsistent b
...
 
-* **View Windows** are the most literal kind of "Window" available. They show content such as presentations, web pages, documents, etc. These Windows act as a sort of frame around the specific content they display. They typically allow multiple instances of an application to be shown and may utilize tabs to help keep content organized without displaying too many windows.
-* **App Windows** are the main windows of apps that are not document-based. These windows typically only allow one instance of an application to be shown.
+* **View Windows** are the most literal kind of "Window" available. They show content such as presentations, web pages, documents, etc. These Windows act as a sort of frame around their specific, displayed content. They typically afford the option of opening multiple instances of an application, and may use tabs to structure an overabundance of windows.
```
danrabbit: I don't really like the rewording here. I think it's much less clear and outside the casual voice. "afford the option of opening" is especially wordy. It feels like you're trying too hard to avoid using the word "allow" instead of actually trying to make the text more clear.

docs/human-interface-guidelines.md
```diff
@@ -332,27 +332,27 @@ The indicator area is notorious for becoming cluttered and having inconsistent b
...
+* **App Windows** are the main windows of apps that are not document-based. These windows typically let applications show only a single instance.
```
danrabbit: I'm not sure about this one either. Seems off for some reason

docs/human-interface-guidelines.md
```diff
@@ -390,11 +390,11 @@ Make both the primary and secondary text selectable. This makes it easy for the
 
 ### "OK" is not Okay {#ok-is-not-okay}
 
-When presenting a dialog to a user, always use explicit action names like "Save...", "Shut Down", etc. Consider that using "OK" enables users to proceed without understand the action they are authorizing. Not all users will read the question or information presented to them in a dialog. Using specific action names will make it harder for a user to select an unintended action and may even encourage them to read the information presented before making a selection.
+When presenting a dialog to a user, always use explicit action names like "Save" or "Shut Down". Consider how "OK" lets users proceed without understanding the action they are authorizing. Not all users will read the question or information presented to them in a dialog. Using specific action names will make it harder for a user to select an unintended action and may even encourage them to read the presented information before making a selection.
```
danrabbit: The ellipsis here is intentional, don't remove it.

docs/human-interface-guidelines.md
lewisgoddard: "Thus" is a little too old school for technical documentation.

docs/human-interface-guidelines.md
lewisgoddard: I agree, "without hardcoding support for them" is clearer, the rest is good.

owcz:
@cassidyjames, ah, sorry about that. Happy to split it up however you suggest—whatever would be most helpful

docs/human-interface-guidelines.md
owcz: Rephrased into imperative

docs/human-interface-guidelines.md
owcz: Rephrased

docs/human-interface-guidelines.md
owcz: More looking to avoid passive voice than trying to force a change for "allow" ("let" is always an easy substitution and windows technically don't have agency to allow but they can afford, especially since this is a design doc). I think part of the issue is the original copy—multiple instances feels like it's implied but I tried to work it in anyway. Rephrased

docs/human-interface-guidelines.md
owcz: Rephrased

lewisgoddard:
This branch requires a merge from master to bring it up it date, and a re-review from @danrabbit

docs/human-interface-guidelines.md
danrabbit: I still think this needs to be worded in a way that instructs the developer to add this integration. This reads as if we're talking about some other people's apps and not your app that you are working on and what you should do.

docs/human-interface-guidelines.md
danrabbit: We use the term "panel" and this dates back to the GNOME 2 days. We don't say menu bar to refer to the panel. We use menubar to refer to the Gtk+ widget `Gtk.Menubar` which we recommend against using in applications.

docs/human-interface-guidelines.md
```diff
@@ -323,7 +323,7 @@ Indicators are small icons that live in the menu bar (wingpanel). They give user
 
 ### Does Your App Need an Indicator? {#does-your-app-need-an-indicator}
 
-The indicator area is notorious for clutter and inconsistent behavior. Given that users will probably install many third-party apps, we must be careful about the number of indicators we show and how they behave. Keep in mind that not every application needs an indicator. Only a very small set of apps will benefit from one. You do not need an indicator if:
+The indicator area is prone to clutter and inconsistent paradigms. Given that users will probably install many third-party apps, we must be careful about the number of indicators we show and how they behave. Keep in mind that only a very small set of applications need or benefit from an indicator. Avoid adding an indicator if:
```
danrabbit: I don't think using the word "paradigms" make this more clear. I'm okay with the other changes

danrabbit:
Thanks looks good! :)

PR #1435, opened 2016-10-08
Turkish translations.

OzgurGorgulu:


btkostner:
We use [transifex](https://www.transifex.com/elementary/elementary-mvp/dashboard/) for our translations. We usually update them every 2 or so weeks.

PR #1437, opened 2016-10-13
Fork Our website on github Button is little offset

akshayvinchurkar:
this small change will fixed that


lewisgoddard:
Actually it seems like all our `.button.flat.suggested-action` render a few pixels to the left of `.button.flat`

akshayvinchurkar:
this is design decision i dont know but if this is then this doesn't make any sense

danrabbit:
My guess it that it was some lazy CSS based on the assumption that there would be a button to the left of it that it's trying to be padded from

akshayvinchurkar:
i take this as mistake if this is then merge the request

cassidyjames:
I think the issue is that the buttons are `display: inline-block` and there's whitespace between them, so the _next_ button is getting a space rendered before it. If I manually smack them up against each other in the HTML, it renders properly.

I am not 100% sure why it's showing up here and not elsewhere. We should probably revisit the button styles to avoid this more globally, though.

lewisgoddard:
This is a well documented phenomenon with several resolutions, although an adjustment of 4px can correct for this, flexbox is also a possible solution. Paging @btkostner 

Reference: https://css-tricks.com/fighting-the-space-between-inline-block-elements/

btkostner:
Did someone say flexbox?

In all seriousness I want to redo the CSS for buttons and button groups as it's a bit of a mess, and has a bunch of size differences between the styles. I had similar problems with while doing store styles.

btkostner:
Closing from #1556

PR #1439, opened 2016-10-14
Windows: Use Rufus to check hash (#1361)

### LINKED ISSUE #1361: Verify your download should use Rufus (Docs on installation from Windows)
Recent versions of Rufus support calculating image hashes (click the `#` in the status bar, since [Version 2.3 (2015.08.28)](https://github.com/pbatard/rufus/blob/b3c76b1f61ce380461edf1f8dbccb93626d174da/ChangeLog.txt)) so the [installation instructions on veryfing the download](https://elementary.io/docs/installation#verify-your-download) should also recommend using this feature. As Rufus is needed anyway to create the installation medium this allows users to use one program less and makes the whole process easier.

quassy:
Fixes #1361

### Changes Summary

- Update & streamline installation instructions for Windows users so only Rufus is needed (and not a second program to verify the image hash).

---

> As Rufus is needed anyway to create the installation medium this allows users to use one program less and makes the whole process easier...

I'm not sure yet my text is the best way to streamline the installation instructions, but maybe it can serve as a basis for discussion. 

Issues:
* Less emphasis on verifying being optional, which I believe is a good thing. With verifying being so easy it shouldn't be called optional anymore. (In my opinion instructions for Linux and macOS should follow suit.) 
* #verify-your-download is removed as a distinct section, otherwise there would be quite a lot of redundancy. I don't think it is needed.
* There is no screenshot showing the computed hashes. (But is it really needed? Especially because a screenshoted hash different from the one given in the text could confuse users.)


quassy:
* I have updated all Rufus images. `rufus_select_iso.png` shows the same parameters as the previous one, is a newer version than on the Rufus website and the usb key is appropriately named "elementary" (but sadly no HiDPI). 
* The header is moved.
* I believe `rufus.png`, `windows_image.png`, `hash-tool.png` can be removed (as they are / will not be used anywhere), but this should happen separately.
* Added `rufus_hash_selected.png`, which could be used to show users which hash to compare (even if it's pretty obvious already).
* The commits are a bit of a mess I'm afraid, maybe they can be squashed?

danrabbit:
Yep, will be a Squash merge. Thanks for your work on this @quassy!

Also I tried to send you an email earlier :p Maybe send me an email if you get a chance? daniel at elementary.io

PR #1444, opened 2016-10-19
Attempt to fix average payments for OS types.

lewisgoddard:
### Changes Summary

- Move OS detection before first call.
- Define variable with other post values.


backend/average-payments.php
```diff
@@ -69,7 +65,7 @@ function LastError($db) {
     }
 
     ////    Update
-    $query  = 'UPDATE `AveragePayments` SET `Total` = `Total` + \''.$amount.'\', `Count` = `Count` + 1, `Average` = ((`Total` + \''.$amount.'\') / (`Count` + 1)) WHERE `OS`=\''.$OS.'\' OR `OS`=\'total\';';
+    $query  = 'UPDATE `AveragePayments` SET `Total` = `Total` + \''.$amount.'\', `Count` = `Count` + 1, `Average` = ((`Total` + \''.$amount.'\') / (`Count` + 1)) WHERE `OS`=\''.$os.'\' OR `OS`=\'total\';';
```
btkostner: using a POST variable directly in a SQL statement is a really bad idea. we should be using `mysql_real_escape_string` on all the inputs for security

lewisgoddard: I wonder why we weren't doing that with `$amount` already. Maybe because the charge will fail if it isn't an `int`?

lewisgoddard: Fixed, although we can't use `mysql_real_escape_string` without a mysql link.

PR #1458, opened 2016-10-27
Updated spanish translation in getting started

desiderantes:
### Changes Summary

- In spanish, Terminal is a female name, so it should be treated as «la Terminal»
- Translated a missing paragraph about SSH
- Removed unnecessary «Usted» treatment, as we have tacit nouns in Spanish


btkostner:
All of those files are automatically generated from transifex, so those changes will not be persistent. Please change the translations [here](https://www.transifex.com/elementary/elementary-mvp/translate/#es/docs_code_getting-started/47153018) instead.

desiderantes:
That link gives me a 404

btkostner:
Try this link instead https://www.transifex.com/elementary/elementary-mvp/dashboard/

Issue #1374, opened 2016-09-14
Screenshots on Learning the Basics need HiDPI versions


danrabbit:
There are only LowDPI versions [here](https://elementary.io/docs/learning-the-basics#learning-the-basics)
- [ ] Slingshot (We already have the CSS one, we should copy that and make sure it's translatable)
- [ ] Plank
- [ ] Panel (maybe this should be drawn with CSS so it's translatable)
- [ ] Window Buttons
- [ ] System Settings

PR #1460, opened 2016-10-31
Add HiDPI images for Learning the Basics page

therladbsgh:
This is for issue #1374 which asks for HiDPI images for the Learning the Basics page. It includes images of Slingshot, Plank, Panel, Windows Buttons, and System Settings, although all the images are in .png format and are not drawn with CSS. I have never tried a pull request before so let me know if I need to change something.


lewisgoddard:
So in [`images/screenshots`](https://github.com/elementary/website/tree/master/images/screenshots) we use the `@2x` suffix to denote hidpi images, rather than your `_hidpi`. I don't know much else for certain, but I would have thought we also want identical downscaled versions, as close to the originals as possible. I'll leave that to #ux

snwh:
If we were to include high DPI versions of the documentation screenshots it'd be best to have them be the same as the original (just pixel doubled or what not) and showcasing the default elementary OS experience –i.e. not having apps like "DrRacket" in the screenshot.

I agree with using the `@2x` suffix for consistency with other asset file names.

lewisgoddard:
I'm going to close this due to lack of activity.

PR #1461, opened 2016-10-31
Fix #1355: Log errors instead of just echoing them.

### LINKED ISSUE #1355: Stop echoing to oblivion
**This is somewhat related to having an `api` folder.**
**Also somewhat related to #1282**
## The problem

There are a number of places in `backend` that we echo errors, even if nothing is catching them. This is bad! We should be using PHP's `error_log` and / or sending it to the client. This is _very_ important when dealing with production bugs, because if we are echoing the error our PHP server logs will not log them!

On top of this, we should setup new relic to listen to the PHP logs for better bug management.
## The end game
- Anything in the `backend` folder should handle or re-throw errors. Never `echo`!
- Any errors that we can't really handle should be send to `error_log`

lewisgoddard:
Fixes #1355

### Changes Summary

- Add `log_echo` function.
- Log all errors.
- Print errors in debug mode.


backend/chart.php
```diff
@@ -105,4 +102,4 @@ function log_info($msg) { // Basic logger
 
 file_put_contents('./chart.json', json_encode($chart, JSON_PRETTY_PRINT));
 
-log_info('Done.');
+log_echo('Chart Update - Done.');
```
btkostner: These should just be echo statements unless it's an error like the first one.

backend/classify.functions.php
```diff
@@ -30,16 +31,19 @@ function ipCheck($hostname, $debug = false) {
...
         $continent = $record->continent->code;
         $country   = $record->country->isoCode;
         $longitude  = $record->location->longitude;
 
     } catch (\Exception $e) {
         if ( $debug ) {
-            echo '<!-- '.$e->getMessage().' -->'."\n";        
+            log_echo('<!-- '.$e->getMessage().' -->'."\n");
+        } else {
+            error_log($e->getMessage());
```
btkostner: I feel these should be flipped. If it's in debug, it should output the message, but not send it to sentry because someone is already debugging it. If it's in production (or not debug mode) it should send the message to sentry (although not echo the error out loud?).

_templates/sitewide.php
```diff
@@ -1,6 +1,7 @@
 <?php
 
 require_once __DIR__.'/../backend/lib/autoload.php';
+require_once __DIR__.'/../backend/log-echo.php';
 require_once __DIR__.'/../backend/config.loader.php';
 
 // Setup sentry error logging
```
btkostner: considering sentry is already setup with the `log-echo` file, we could remove the next couple of lines

PR #1472, opened 2016-11-08
TinyPNG Responsive Images & Screenshots

lewisgoddard:
These images are a third of the size, but with noticeable quality variations.


btkostner:
@lewisgoddard What is the hold on this? Do we just need to work with #ux to find a sweet spot for size and quality?

brejoc:
Sorry for the vague comment, but I'm completely lacking overview in the project. Wouldn't it be possible to automate the image optimization during the deployment process?

lewisgoddard:
Probably possible to automate, but we need to avoid the banding in the images here. I think they are actually compressed as much as is possible without significant loss of quality.

brejoc:
An other option might be a lossless image optimisation, with which you wouldn't have to worry. Has someone compared the sizes yet?

brejoc:
I've made some tests with `optipng`, which should only use lossless techniques. I've checked out the repo and gave `optipng` a try on every png image in `images` with the highes possible retry rate: `find ./images -name "*.png" -print0 | xargs -r0 optipng -o7 > output_optipng`

Before:  
```
$ du -s images/
18572   images/
```

After:  
```
$ du -s images/
17828   images/
```

While the results are somewhat underwhelming, there is a reduction in file size. But I should also mention, that using `optipng` with `-o7` takes some time. I guess the cost-benefit ratio depends on your CPU speed.


`optipng` ouput: [output_optipng.txt](https://github.com/elementary/website/files/642270/output_optipng.txt)

lewisgoddard:
I'm going to trash this, since compressing significantly seems to result in a significant loss of quality, and since lossless compression saves less than 5% of the size I don't think it's worth automating.

PR #1479, opened 2016-11-16
add black and white hero tees

danrabbit:
I'm guessing this will need changes to the store since these are technically the same product ID? Or maybe I just don't know what I'm doing :D


btkostner:
@danrabbit this should be ready for merge now if you want to just double check before merging.

danrabbit:
Let's do it :D

PR #1480, opened 2016-11-17
Retail update

btkostner:
So this changes the data structure of our `store.json` file to something more flexable, but considerably less DRY. As such a bunch of logic all over the place had to change. Luckely not much had to change in the checkout and order pages.

This overview is quite large and hard to understand. Going commit by commit should stay to a few files each, and will be a lot easier to understand what has changed.

**DO NOT MERGE THIS TO MASTER YET**. This changes how the shipped hook finds the product, so it will need to be pushed while no outstanding orders exist.


lewisgoddard:
`+966 −529` is quite a lot of repetition, unless this adds some features too.

btkostner:
This doesn't add any features aside from being able to have multiple products with the same printful product. There are some files moved around considering there are better spots for them now. One being the API error message is now in a template, and second being the store inventory file is now in the API folder.

btkostner:
Looking back at this PR, there is quite a bit I changed that ~~could~~ _should_ go into another PR, and be simplified down. Let me know if you want me to redo this and I will.

btkostner:
Closing in favor of #1555

PR #1486, opened 2016-11-29
start keeping track of the projects we donate to

danrabbit:
Start keeping track of the projects we've donated to in a public manner. So far I have
- Bountysource
- Debian
- Fontawesome
- Inkscape
- Open Street Map
- rEFInd
- Software Freedom Conservancy


open-source.php
```diff
@@ -155,6 +155,38 @@
     </div>
 </div>
 
+<div class="grid">
+    <div class="whole">
+        <div class="two-thirds">
+            <h2>Projects We've Supported</h2>
```
cassidyjames: `Projects We Support` sounds better I think, as some of these are recurring and not just one-off "yeah we supported them one time".

danrabbit: Yeah I was thinking about that as well. I think I personally would rather lean towards the more conservative language even if it's less generous to us just because I feel like it's more honest since some of them are one-offs. I don't want to mislead anyone into thinking that we continually pass money to these projects every month if we've only done it once or twice

danrabbit: We could try using a more active phrase like "We Support Open Source Projects"

cassidyjames: Yeah that makes sense. I like `We Support Open Source Projects`

lewisgoddard: You wanted this to be "We Support Open Source Projects" ..?

open-source.php
```diff
@@ -155,6 +155,38 @@
...
+            <p>elementary makes a point to send funds back to the projects we rely on. When you purchase a copy of elementary OS, you're also supporting great projects like these.</p>
+        </div>
+    </div>
+    <div class="donation-grid">
+        <a href="https://salt.bountysource.com/teams/bountysource" target="_blank">
```
lewisgoddard: You shouldn't need to `target="_blank"` these links, the website does that automatically.

PR #1487, opened 2016-12-01
add developer redirect to developer.elementary.io

btkostner:
Redirects developer page to `developer.elementary.io` subdomain.

We will need to update translations to remove the stale ones once this goes to master.


danrabbit:
Adding "on hold" while we work out cert issues and the like

btkostner:
This should be good to merge into master now as everything should be setup and working as intended

danrabbit:
I opened [a small issue](https://github.com/elementary/houston/issues/314) to avoid a regression :)

danrabbit:
I can confirm that all the metadata twitter socially stuff is working again.

btkostner:
In the famous words of developers everywhere, "I really hope this doesn't break in production."

PR #1494, opened 2016-12-08
move social links to the top of the page

danrabbit:
That's a bold strategy, Cotton

This branch:
* puts all the header links that stay on our site on the left side
* puts all the social links on the right side
* Demotes the "blog" link to just be a Medium social network icon
* Moves the links in the footer where the social media links used to be

If the design here is approved, we'll need to do the same for the templates at Developer before merging and then merge those at the same time

### DIFF
--- a/_styles/main.css
+++ b/_styles/main.css
@@ -271,7 +271,7 @@ nav a:visited {
     display: inline-block;
     fill: inherit;
     margin: 0;
-    padding: 0 16px;
+    padding: 0 12px;
 }
 
 nav a:focus {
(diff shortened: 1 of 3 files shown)


lewisgoddard:
Would this imply the destruction of the Tumblr-hosted blog.elementary.io domain? Perhaps redirection to medium?

danrabbit:
@lewisgoddard, I don't think we should change the blog subdomain because it would create a whole bunch of broken links. I imagine we'll still cross-post there until at least the traffic to our website from Medium overtakes it in analytics

PR #1506, opened 2016-12-27
simplify async loading of analytics

btkostner:
There is no need to put async loading of Google analytics while we are already promising the function. This reduces the confusion and loading problems with analytics by just waiting for it to load before running the code.


btkostner:
I'll push this now and keep an eye on analytics and sentry tonight / tomorrow to see if there is any regression.

PR #1512, opened 2016-12-29
double down on call to action

### LINKED ISSUE #1500: Double down on the call to action
Been noticing  a lot of websites put their call to action at the bottom of the page as well as the top. We should throw down a big suggested action button at the bottom of the home page that scrolls you back up to the Pay-What-You-Want widget

danrabbit:
Fixes #1500 

screenshot from 2016-12-29 10 08 37

### DIFF
--- a/_styles/main.css
+++ b/_styles/main.css
@@ -156,9 +156,14 @@ img.h1 {
     opacity: 0.8;
 }
 
+div.inline,
 img.inline {
     display: inline-block !important;
     vertical-align: middle;
+}
+
+img.inline + div.inline {
+    margin-left: 12px;
 }
 
 .inline-tweet {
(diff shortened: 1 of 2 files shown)


lewisgoddard:
I'm not sure I'm happy with the action of that button. Scrolling to the header makes sense, but only takes it so the header is at the very top of your screen. On large screens, this makes the quotes from press front and center. On smaller screens, I would assume it works much better.

danrabbit:
I agree it's less than ideal, but I don't really know what else to do. To make it centered or even at the bottom of the display, there would have to be an ID somewhere in the middle of the laptop image. I'm thinking that this "hero" should probably be redesigned to fit better on a desktop layout so that the whole thing sits inside 100vh

PR #1514, opened 2016-12-29
redo css to make index hero 100vh

btkostner:
This changes the CSS for the homepage hero to show the elementary logo, hero image, and payment options all on screen when the page loads. At odd sizes it will revert back to regular flow to avoid a super small hero image. Because we change to use an `img` tag instead of CSS `background-image`, we lost some image responsiveness.

When testing this, the development hero will mess up alignment, so I would recommend running it with 
`PHPENV=production php -S 0.0.0.0:8000 router.php`.

### DIFF
--- a/_styles/main.css
+++ b/_styles/main.css
@@ -435,7 +435,7 @@ footer ul li a:focus {
     object-fit: contain;
 }
 
-@media (min-aspect-ratio: 1/1) and (min-height: 600px) {
+@media (min-aspect-ratio: 1/2) and (min-height: 800px) and (max-height: 2000px) {
     .section--stretched {
         height: calc(100vh - 56px);
     }
--- a/_styles/main.css
+++ b/_styles/main.css
@@ -472,7 +472,7 @@ footer ul li a:focus {
         width: calc(33% - 88px);
     }
 
-    .column + .column {
+    .column + .column + .column {
         margin-left: 16px !important;
     }
 }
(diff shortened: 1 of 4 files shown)


danrabbit:
I think this would need to be paired with layout out the pay-what-you-want section horizontally. Otherwise we just lose too much vertical space and the image is too small on laptop sizes
screenshot from 2016-12-29 12 51 04

lewisgoddard:
If we're going to do that, do we want the selection buttons to be the same padding as the final button is?
Also, since there is always text under the final button, do we want the "Enter any amount" under "Custom" to always be visible?

lewisgoddard:
image

image

_styles/main.css
```diff
@@ -410,6 +410,43 @@ footer ul li a:focus {
     outline: none;
 }
 
+/***********
+* Sections *
+***********/
```
danrabbit: Shouldn't this go in home.css since this is all specific to the homepage?

lewisgoddard: Possibly, depends on if we intend to reuse it. I know blake wants more flexbox.

btkostner: I do like flexbox.

I tried to keep it as generic as possible so we could reuse it, but I'm not sure if any other page would use it. Maybe the get involved page if we had an image or something to take up a lot of space.

As of right now this could be moved to `home.css`

btkostner:
The spacing between the button and the "Pay what you want:" label seems a bit off, but it's not that big of a deal.

screenshot from 2016-12-30 13 27 50

PR #1521, opened 2017-01-01
iconify and reduce vertical space in get involved/funding

danrabbit:


get-involved.php
```diff
@@ -40,49 +40,40 @@
 <section id="funding" class="grey">
     <div class="grid">
         <div class="two-thirds">
-            <h1>Funding</h1>
+            <h2>Funding</h2>
```
cassidyjames: I agree with this change, but now it doesn't match the other sections. Should we update the other sections to have h2 headings and h3 subheadings?

danrabbit: pushed

PR #1524, opened 2017-01-01
Update GPL Header

Philip-Scott:
Address removed and aligned the *


danrabbit:
Gonna close this since there wasn't any other discussion about switching to LGPL by default

PR #1546, opened 2017-01-18
run hide function on page load

### LINKED ISSUE #1543: Shipping shows USA states for other countries on load.
If you are not placing order from USA, states are shown anyway. They do hide later if you manually select some other country and appear back when choosing "United States".

Tried on Firefox 50 and Opera 42

btkostner:
Fixes #1543

### DIFF
--- a/_scripts/pages/store/cart.js
+++ b/_scripts/pages/store/cart.js
@@ -97,13 +97,16 @@ Promise.all([jQuery, analytics]).then(([$, ga]) => {
         })
 
         /**
-         * Hide the inputs we don't need depending on the country
-         * NOTE: JQuery in a loop does not work as intended :sad_face:
+         * updateAddressForm
+         * Updates state field based on current selected country
+         *
+         * @return {void}
          */
-        $('form[action$="checkout"] select[name="country"]').on('change', function (e) {
-            var value = $(this).val()
-            var $state = $(this).siblings('select[name="state"]')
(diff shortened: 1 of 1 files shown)


_scripts/pages/store/cart.js
```diff
@@ -97,13 +97,16 @@ Promise.all([jQuery, analytics]).then(([$, ga]) => {
...
-            var value = $(this).val()
-            var $state = $(this).siblings('select[name="state"]')
-            var $statelabel = $(this).siblings('label[for="state"]')
+        const updateAddressForm = () => {
+            const form = $('form[action$="checkout"]')
+            const value = $('select[name="country"]', form).val()
+            const $state = $('select[name="state"]', form)
+            const $statelabel = $('label[for="state"]', form)
 
             ga('send', 'event', 'Cart', 'Country Change', value)
```
lewisgoddard: This event should only be tracked when the user changes the country, so that we know the prediction engine was wrong. Can we remove it when the pageload fires? It would make tracking more consistent and easier to analyse.

lewisgoddard: This event should only be tracked when the user changes the country, so that we know the prediction engine was wrong. Can we remove it when the pageload fires? It would make tracking more consistent and easier to analyse.

btkostner:
@lewisgoddard fixed now

PR #1551, opened 2017-01-20
Update Copy to reflect the move to HeaderBar

danrabbit:


btkostner:
It might be worth mentioning that the granite app menu icon is not dragable.

danrabbit:
@btkostner ah damn I didn't realize that. I think that widget shouldn't be used any more though, so I guess it could be considered a bug in the app

PR #1558, opened 2017-01-25
change shipping title based on country user is from

btkostner:
### Changes Summary

- Changes the 'Worldwide Shipping' title based on the country the user is from

https://beta.elementary.io/store-shipping/store/


store/index.php
```diff
@@ -144,7 +145,19 @@
...
+            $country_code = getCurrentCountry($ip);
+            $shipping_text = 'Worldwide Shipping';
+
+            if ($country_code && $country_code !== 'US') {
+                $country_name = getCurrentCountryName($ip, $page['lang']);
+
+                if ($country_name) {
+                    $shipping_text = 'Now shipping to ' . $country_name;
+                }
+            }
```
lewisgoddard: You need an else here to handle Cuba, Iran, and North Korea.

lewisgoddard:
Code looks fine, but the beta system doesn't seem to run.

lewisgoddard:
> Now shipping to United Kingdom
Hm, really it is "The United Kingdom"

lewisgoddard:
There's quite a few with that sort of exception, does that database not have a nice-name field? https://en.wikipedia.org/wiki/List_of_sovereign_states

btkostner:
No it does not. It's all translated fields.

btkostner:
Final thoughts on this. Should we setup a catch all for countries that need "the" in front of the name, or just role as is right now?

lewisgoddard:
I'd appreciate the catch-all, but this isn't going to be easily translatable, is it?

btkostner:
It's currently using translations from the library.

Translations for Japan:
```
{
["de"]=> string(5) "Japan"
["en"]=> string(5) "Japan"
["es"]=> string(6) "Japón"
["fr"]=> string(5) "Japon"
["ja"]=> string(6) "日本"
["pt-BR"]=> string(6) "Japão"
["ru"]=> string(12) "Япония"
["zh-CN"]=> string(6) "日本" 
}
```

lewisgoddard:
Sure then, we should try and catch those exceptions and see whether or not that seems better once translated.

btkostner:
I'm going to close this PR because translations are going to be a huge pain for very little gain.

PR #1566, opened 2017-02-05
draw panel in learning the basics with css

danrabbit:
Part of #1374

https://beta.elementary.io/panel-css/

Draws the panel with HTML and CSS which makes it slightly responsive, look nice on HiDPI, and translatable.

Only problem is that the icon colors are wrong. I tried to add a php import here, but it seems to get commented out automatically?


lewisgoddard:
Yeah, you can't add a PHP import inside a markdown file, that's a valid markdown comment syntax. You should still be able to style them as img's.

PR #1587, opened 2017-02-11
Getting Started guide example fixes

cimm:
_New to elementaryOS and going over the "[Getting Started](https://elementary.io/docs/code/getting-started)" guide for developers. I am a programmer but never programmed in Vala nor GTK before so this is a n00b trying things out.  The guide is well explained, easy to follow, and coherent. Thanks for that!_

I did run into problems near the end of the guide:

1. The "[Functionality in Gtk.Grid](https://elementary.io/docs/code/getting-started#functionality-in-gtk-grid)" section introduces the `this` concept at the end of the example code:

```vala
this.add (grid);
```

It's the first time it's used. The guide uses ` window.add (grid)` in all other examples and it does not explain where it comes from. Wouldn't it be better to keep using `window` here?

2. If I understand correctly the section is supposed to explain communication between the button and the label:

> ... but instead of just changing the label of the button, we’re going to use it to change an empty label to a message.

Now, the code to update said label:

```vala
button.clicked.connect (() => {
    button.label = _("Hello World!");
    button.sensitive = false;
});
```

I might not understand it but this seems like it will update the text _on the button_, not the label created earlier. If so, it misses the point completely as it's basically identical to the earlier example.

Shouldn't it be something like:

```vala
button.clicked.connect (() => {
    label.set_text (_("Hello World!"));
    button.sensitive = false;
});
```

I might misunderstand the explanation but this could mean others have troubles with that section as well.

### DIFF
--- a/docs/code/getting-started.md
+++ b/docs/code/getting-started.md
@@ -600,7 +600,7 @@ This time when we created our grid, we gave it another property: `row_spacing`. 
 Now, let’s hook up the button to change that label. To keep our code logically separated, we’re going to add it below `window.add (grid);`. In this way, the first portion of our code defines the UI and the next portion defines the functions that we associated with the UI:
 
     button.clicked.connect (() => {
-        label.set_text (_("Hello World!"));
+        label.label = _("Hello World!");
         button.sensitive = false;
     });
 
--- a/docs/code/getting-started.md
+++ b/docs/code/getting-started.md
@@ -593,14 +593,14 @@ Let’s create a Window with a vertical Grid that contains a Button and a Label:
     grid.add (button);
     grid.add (label);
 
-    this.add (grid);
+    window.add (grid);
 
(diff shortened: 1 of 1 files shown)


docs/code/getting-started.md
```diff
@@ -593,14 +593,14 @@ Let’s create a Window with a vertical Grid that contains a Button and a Label:
...
+    window.add (grid);
 
 This time when we created our grid, we gave it another property: `row_spacing`. We can also add `column_spacing`, but since we’re stacking widgets vertically we’ll only see the effect of `row_spacing`. Notice how we can create new widgets outside the grid and then pack them into the grid by name. This is really helpful when you start using different methods to change the properties of your widgets.
 
-Now, let’s hook up the button to change that label. To keep our code logically separated, we’re going to add it below `this.add (grid);`. In this way, the first portion of our code defines the UI and the next portion defines the functions that we associated with the UI:
+Now, let’s hook up the button to change that label. To keep our code logically separated, we’re going to add it below `window.add (grid);`. In this way, the first portion of our code defines the UI and the next portion defines the functions that we associated with the UI:
 
     button.clicked.connect (() => {
-        button.label = _("Hello World!");
+        label.set_text (_("Hello World!"));
```
danrabbit: instead of using the `set_text` method we can still use properties here and do `label.label = _("Hello World!");`

cimm:
Using the property instead of the setter method now, thanks for the feedback.

Issue #1287, opened 2016-08-26
Line items on Square receipt


cassidyjames:
If possible, we should provide order line items on the Square receipt for merch (i.e. 1× L Logotype Tee Men's: $25) instead of just a single "elementary store" line item.

btkostner:
Stripe itself only allows a charge description which is what we are currently using. Sadly it's only a single line. The ~~easier~~ better approach would be to setup a hook for Stripe for a payment and do our own emails.

lewisgoddard:
I thought the plan was always to replace Stripes emails over time.

danrabbit:
Gonna add "needs design" since we need to create a template for this

lewisgoddard:
We should probably implement a similar thing for purchases of the OS, but with download links. That would take care of #34 and #24

lewisgoddard:
Do we want to use Mandrill for this as well?

btkostner:
I would assume we will use mandrill for this as well. No reason to use yet another service.

PR #1598, opened 2017-02-21
Implement custom receipts for OS purchases

### LINKED ISSUE #1455: Implement custom receipts for OS purchases.
### Pre-deployment

- [ ] Add function to check Order ID exists
- [ ] Add API endpoint to set cookie like order happened in this browser.

### onSuccessfulOrder

- [ ] Add order ID to link in email
- [ ] Send custom, responsive receipt through Mandrill.

### onDownloadAgain

- [ ] Hit `/api/download-again?order-id=SOMETHING_SECRET_AND_LONG`
- [ ] Log the use and set the cookie.
- [ ] Redirect to homepage, who is none the wiser and treats the user like they just ordered it.

#### Ordered Cookie

```
// Set an insecure, HTTP only cookie for 10 years in the future.
$encoded = urlencode(str_replace(' ', '_', 'has_paid_'.$description));
setcookie($encoded, $amount, time() + 315360000, '/', '', 0, 1);
```

[api/payment.php, Line 24](https://github.com/elementary/website/blob/master/api/payment.php#L24)

#### Check Order ID

```
// Check the stripe checkout to make sure we have a _real_ call
try {
    $ch = \Stripe\Charge::retrieve($res['data']['order']['external_id']);
} catch (Exception $e) {
    header('HTTP/1.0 400 Bad Request');
    echo 'Inaccurate data';
    die();
}
```

[hook/store/index.php, Line 39](https://github.com/elementary/website/blob/master/hook/store/index.php#L39)

btkostner:
Fixes #1455

### Changes Summary

- Add a hook endpoint for stripe to send custom email receipts
- Add an endpoint for returning users to set cookie
- Add a email template for receipts

https://beta.elementary.io/checkout-emails/


cassidyjames:
I just tested this on the beta site and got the normal Stripe email. Am I missing something?

btkostner:
I wrote this to use the stripe hook when a charge is successful. So we will need to setup hooks for it to work.

Although on second though, I could just send the email when the payment passes on our system...

btkostner:
@cassidyjames I just switched it to send the email when we get a payment instead of on hook. You should be able to test it on the beta site now.

lewisgoddard:
Have you considered #1287 in this at all? The two are heavily related and could probably share some assets.

lewisgoddard:
Generally we use the PHP native `exit;` or `exit(1);` instead of its alias `die();`

btkostner:
I have not, but I can include that on this PR. I'll work on it tomorrow, and probably into next week.

btkostner:
@lewisgoddard We don't have the store email template done. I'll do the store emails in another PR, but I consider this one done as is, right now.

PR #1611, opened 2017-03-03
avoid button text flash on already downloaded

btkostner:
Renders the button with correct text on the server to avoid a quick flash text change.


btkostner:
No. We have two javascript strings at the bottom of the page that get translated, and we use those translations during rendering. Confirmed Japanese translations worked without a problem.

PR #1619, opened 2017-03-21
New Projects page

tintou:
This adds a new project page to get a clear list of all the available projects under the elementary umbrella


btkostner:
I'm going to ping @elementary/ux about the design and what a page like this would need.

danrabbit:
What's the purpose of this page? What is intended to do that can't be achieved by visiting https://github.com/elementary ?

tintou:
@danrabbit GitHub UI isn't equivalent to what we had with Launchpad, it's now difficult to find a project on it, with this page everything is sorted and categorized. Plus this is very useful when some projects are Not on GitHub

danrabbit:
@tintou Can you describe more the specific problems you have? What exactly are you using this page to do that you can't do on GitHub and why not? I think before we commit to having another page on the website we should be clear about what the problem we're solving is.

Almost all repositories are migrated to GitHub. We'll finish it up pretty soon, so if that's the driving motivation it will not be relevant much longer.

tintou:
@danrabbit The real problem is that the main page is not sorted alphabetically and the projects are available in several pages so it's a pity to navigate there. Plus the amount of project growing won't make it any easier.

lewisgoddard:
I agree that the discover-ability on the matching GitHub page is lacking, especially if you're not sure what to search for. I also agree this needs to list repositories automatically. We might be better off having some table system, with the API we can also list how many issues each repo has and display more information.

btkostner:
We could create an API script similar to our average payment data generator script. Dump all of the info into a JSON file. Run it with systemd every hour or something.

btkostner:
I think the consensus here is that maintainability will be too hard. We have no way to get the icon from the GitHub API, and keeping a hardcoded list is out of the question.

If we want to continue this, open an issue so we can plan this out better.

PR #1624, opened 2017-03-27
New Hardware section

cassidyjames:
Since we're starting to see hardware distributors ship elementary OS, we want to publicize our guidelines on the branding page.

### DIFF
--- a/brand.php
+++ b/brand.php
@@ -104,7 +104,7 @@
     <section id="hardware">
         <h2>Hardware Distributors</h2>
         <p>We want to ensure that as long as our software carries the elementary branding, the experience will be consistent whether it was downloaded it from our website or pre-installed on a hardware product.</p>
-        <p>The software components of elementary OS may be modified and redistributed according to the terms of the software&rsquo;s licensing. However, our brand marks may only be redistributed under one or more of the following conditions:</p>
+        <p>The software components of elementary OS may be modified and redistributed according to the terms of the software’s licensing. However, our brand marks may only be redistributed under one or more of the following conditions:</p>
         <ol>
             <li>The software remains unchanged, including pre-installed applications, stylesheets and iconography, configuration files, etc., or</li>
             <li>The modifications are approved in writing by elementary.</li>
--- a/brand.php
+++ b/brand.php
@@ -103,15 +103,19 @@
     
     <section id="hardware">
         <h2>Hardware Distributors</h2>
-        <p>When a user sees or experiences elementary OS, we want to be sure they are seeing it as we intended. We also want to ensure they will have a consistent experience no matter how they access it, be it by downloading it from our website or using it pre-installed on a hardware product.</p>
-        <p>A modified version of elementary OS may be re-distributed pre-installed on a hardware product so long as it meets one or more of the following conditions:</p>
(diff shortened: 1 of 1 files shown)


cassidyjames:
Available at https://beta.elementary.io/branding-hardware/brand#hardware

brand.php
```diff
@@ -100,6 +100,19 @@
         <img src="<?=$sitewide['root'];?>images/brand/community-black.png" alt="elementary Community Logo in black"/>
         <p>This helps establish the product as part of the overall elementary community while reducing confusion that can arise from using the main logomark.</p>
     </section>
+    
+    <section id="hardware">
+        <h2>Hardware Distributors</h2>
+        <p>When a user sees or experiences elementary OS, we want to be sure they are seeing it as we intended. We also want to ensure they will have a consistent experience no matter how they access it, be it by downloading it from our website or using it pre-installed on a hardware product.</p>
+        <p>A modified version of elementary OS may be re-distributed pre-installed on a hardware product so long as it meets one or more of the following conditions:</p>
+        <ol>
+            <li>It is completely rebranded, containing none of the elementary brand marks,</li>
```
danrabbit: Probably should add a line stating that they need to follow the terms of the licensing of the including software, specifically the GPL

brand.php
```diff
@@ -100,6 +100,19 @@
         <img src="<?=$sitewide['root'];?>images/brand/community-black.png" alt="elementary Community Logo in black"/>
         <p>This helps establish the product as part of the overall elementary community while reducing confusion that can arise from using the main logomark.</p>
     </section>
+    
+    <section id="hardware">
+        <h2>Hardware Distributors</h2>
+        <p>When a user sees or experiences elementary OS, we want to be sure they are seeing it as we intended. We also want to ensure they will have a consistent experience no matter how they access it, be it by downloading it from our website or using it pre-installed on a hardware product.</p>
+        <p>A modified version of elementary OS may be re-distributed pre-installed on a hardware product so long as it meets one or more of the following conditions:</p>
```
danrabbit: Maybe we should rephrase this a bit and put the focus on our brand. They can redistribute the software as long as they follow the terms of the software's licensing. But they can only redistribute our brand marks under the terms of their licensing, which is as long as the OS remains unmodified.

danrabbit:
Something I just thought of is that we should make it easy to comply with removing trademarks if that's the route they choose. We should call out what exactly this entails and maybe even how to do it. As far as I remember, the only modifications that would be needed would be:

* LSB Release
* The boot/shutdown screen
* The `distributor-logo` icon

cassidyjames:
Working on this directly with @danrabbit in a Google Doc. Will update the PR once we've settled on copy.

cassidyjames:
Go go gadget @danrabbit

brand.php
```diff
@@ -100,6 +100,23 @@
         <img src="<?=$sitewide['root'];?>images/brand/community-black.png" alt="elementary Community Logo in black"/>
         <p>This helps establish the product as part of the overall elementary community while reducing confusion that can arise from using the main logomark.</p>
     </section>
+    
+    <section id="hardware">
+        <h2>Hardware Distributors</h2>
+        <p>We want to ensure that as long as our software carries the elementary branding, the experience will be consistent whether it was downloaded it from our website or pre-installed on a hardware product.</p>
```
danrabbit: it was downloaded it :p

cassidyjames:
@danrabbit ^

PR #1630, opened 2017-04-02
Docs: executables also in RDNN

danrabbit:
Right?


cassidyjames:
Yeah I agree that we should settle on this convention. Could potentially check for it in Houston eventually.

Issue #1649, opened 2017-04-13
No purchase window for dollars > 0 and < 1


acagastya:
### Expected behavior
The download button should read "Purchase Elementary"

### Actual behavior
Even though I use custom amount for donation (US$ 0.99), I can not make the payment since the window won't appear and the button reads "Download Elementary".

### Steps to reproduce the behavior

1. Go to <elementary.io>
2. Enter 1 in the Custom textbox for "Pay What You Want".
3. The button reads "Purchase elementary OS"
4. Now change the value of Custom input to 0.99
5. The button new reads "Download elementary OS"

btkostner:
This is intended behavior. It's not cost effective to process payments under $1.

PR #1650, opened 2017-04-13
return floatval() for os_payment_getcookie

acagastya:
Fixes #
Else, even if the user specifies money in the range of 0.01 to 0.99, the button won't say "Purchase elementary OS", and if user decided to pay $0.99, payment window won't trigger.

### Changes Summary
- change return type floatval() for os_payment_getcookie from int to float


lewisgoddard:
As mentioned under #1649, this is the intended behavior due to processing fees.

We appreciate the contribution.

PR #1652, opened 2017-04-14
make empty strings return original text when translating page

btkostner:
When a translation line contains an empty string, we will no longer translate it. This is the first step to fixing weblate showing 100% translations for everything.


btkostner:
Once this is merged, I'll submit a huge PR fixing all of the language files with untranslated strings.

PR #1656, opened 2017-04-16
Translation script update

btkostner:
General changes:
- Added a developer cli library so we get nice progress bars for console commands.
- Setup composer to autoload the `Lib` directory (more to come)
- Added some useful variables and method to `L10n` class
- Moved composer lib folder to `vendor` as this is more standard, and lets us use the `lib` folder
- Added a bootstrap file that loads composer files and config, but _does not_ send any data
- Removes dead transifex code

Changes to how translation files are written:
- Untranslated lines now output a blank string instead of the source string
- Checks for untranslated lines (fixes weblate showing 100% translated all the time)

File path changes:
-  `l10n.php` -> `Lib/L10n.php`
- `extract-l10n.php` -> `Console/Extract.php`
- `.tx/prepush.sh` -> `Console/Translation.php`

I would recommend running `php ./_backend/Console/Translation.php` and check language files to ensure everything is working as intended.


btkostner:
ping @lewisgoddard: when you get a chance, can you please review this. I would like to get weblate updating all sorted out sooner rather than later.

_backend/bootstrap.php
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * _backend/bootstrap.php
+ * Entry file for the whole website. Handles configuration loading.
+ */
+
+require_once __DIR__ . '/vendor/autoload.php';
+
+// TODO: why are these not in composer?
+require_once __DIR__ . '/../docs/_mdr/Parsedown.php';
+require_once __DIR__ . '/../docs/_mdr/ParsedownExtra.php';
```
lewisgoddard: These are only needed in the docs folder, and were part of an upstream app we butchered. They could be moved so long as we check anywhere else they are referenced.

btkostner: Moved to composer, or these lines moved out of the bootstrap file?

lewisgoddard: Moved to composer, so long as we're sure they'll still load everywhere where docs need them.

btkostner: Alright. I'll open it as an issue.

PR #1665, opened 2017-04-23
Huge language file update

btkostner:
This is the first run of the new extract script.

### Changes
- Updates all translation sources with the current website ones.
- Removes untranslated strings
- Removes the english language translations


btkostner:
Travis says it's ok, so it's probably ok.

PR #1666, opened 2017-04-23
add more languages to website support

btkostner:
Adds some more languages to the website.

Needs input from the translation team.


_backend/Lib/L10n.php
```diff
@@ -99,21 +99,29 @@ public static function languageDirectory($lang)
     }
 
     protected $available_langs = array(
-        'en' => 'English',
```
lewisgoddard: English must be at the top of the array so it always appears in a separate section at the top of the list.

_backend/Lib/L10n.php
```diff
@@ -100,20 +100,29 @@ public static function languageDirectory($lang)
...
         'ca' => 'català',
         'cs_CZ' => 'čeština',
         'de' => 'Deutsch',
         'es' => 'Español',
+        'fi' => 'Finnish',
         'fr' => 'Français',
         'it' => 'Italiano',
         'ja' => '日本語',
+        'ko' => '한국어',
+        'ko_KR' => '한국어 (대한민국)'
```
sukso96100: Seems like you forgot to add `,` at the end of this line!

PR #1667, opened 2017-04-23
remove ko_KR language

btkostner:
### Changes Summary

- Remove `ko-KR` language because it's now condenced to just `ko`.


btkostner:
Once this is merged we can delete the language in weblate

PR #1679, opened 2017-04-30
Fix #1676: HIG sidebar list styling inconsistent between browsers

### LINKED ISSUE #1676: HIG sidebar list styling inconsistent between browsers
The sidebar appears to look and function properly on stable Chrome and Epiphany, but has various states of screwy on Firefox and upcoming versions of Chrome.

## Chrome stable and Epiphany:

chrome-stable

## Latest Firefox:

firefox

## Chrome Unstable:

screenshot from 2017-04-28 10-26-05

---

It looks like our list styling needs to be revisited as it will get real screwy on future versions of Chrome and also looks oddly different on Firefox.

lewisgoddard:
Fixes #1676: HIG sidebar list styling inconsistent between browsers

Building at https://beta.elementary.io/fix-1676/docs/human-interface-guidelines

### Changes Summary

- Add more definitive list definitions to sidebar styles.


_styles/docs.css
```diff
@@ -200,7 +200,9 @@ nav {
     color: #403757;
 }
 
-.sidebar > ul {
+.sidebar > ul,
```
btkostner: Can we make this one like `.sidebar ul` to look like the others, or does it break css specificity?

lewisgoddard: That would break any levels past the first three, and that matches the first one fine like that.

lewisgoddard: There shouldn't be that many levels, but I'd prefer to cover it.

lewisgoddard:
Do we want to fix the sidebar not being fixed in Firefox too, or keep that separate?

btkostner:
Alright, I added a fix for the sidebar in firefox. @lewisgoddard want to double check this.

PR #1696, opened 2017-05-10
Add Terminal tee

### LINKED ISSUE #1621: Add Terminal tee to store
Just a reminder that we want to do this now that we can multiple designs with the same base product.

danrabbit:
Fixes #1621


cassidyjames:
Oh TIL we have the others there, too. Huh.

danrabbit:
Yeah because we have to send it with the order to printful so they know what to print :p

lewisgoddard:
I'm unclear on if we can merge this now or if we're waiting on Printful for something?

PR #1707, opened 2017-05-16
Improve homepage showcase perf

loklaan:
### Changes Summary

- Remove un-optimizable height transition from showcase slider  
  [Transitioning element height](https://csstriggers.com/height) is a very expensive operation for browsers, because it forces reflow.


btkostner:
The code looks good. I'll pass this over to @elementary/ux for their design approval.

btkostner:
The more I think about this, the more I think it would be a better approach to get the max height of the slides, set the container to that height, and center all the slides in the middle. This would avoid changing heights all together.

loklaan:
You could do that. The vertical whitespacing wouldn't be right though.

If you're hesistant with this fix, _because it introduces sudden-jank_, I'd encourage you to think about how the user's perception is immediately driven by the horizontal transitions of the slides entering and exiting.

PR #1711, opened 2017-05-16
Update press logos on mouse hover and click

user501254:
Fixes https://github.com/elementary/website/pull/431

### Changes Summary

- increase opacity of press logos when hovering over them
- move the press logos slightly into the page on click

~This is how it looks:~
gif-3-b6d8dc5ce0


lewisgoddard:
If we add this we'd probably want a CSS transition.

danrabbit:
I dig it. I agree I'd like a simple `ease-in-out`

user501254:
@lewisgoddard @danrabbit 
I have made a new commit to add the `ease-in-out` transition with a duration of 250ms, keeping it consistent with rest of the transitions on the page..

user501254:
We could probably also use the same transition affect on `.oss-logo` in [Open Source ⋅ elementary](https://elementary.io/open-source) towards the bottom of the page. The opacity change looks quite abrupt.

lewisgoddard:
That seems like a good idea.

PR #1723, opened 2017-05-20
Update TRANSLATE.md

btkostner:
This removes a bunch of the old translation documentation from transifex. I did remove some additional things that shouldn't be relevant to the new weblate instance.


lewisgoddard:
I feel like notes about reviewing are still pertinent, if different.

PR #1728, opened 2017-05-25
allow download endpoint for any version matching current major

### LINKED ISSUE #1726: Download link in receipts does not trigger download
Getting multiple reports of this lately

btkostner:
Fixes #1726 

### Changes Summary

- Allows any os purchase with the same major version to download the current version from sent email receipt.

### DIFF
--- a/api/download.php
+++ b/api/download.php
@@ -50,19 +50,27 @@ if (isset($charge['metadata']['products'])) {
 $isoVersion = false;
 foreach ($products as $product) {
     if (substr($product, 0, 3) === 'ISO') {
+        // Set $isoVersion to the ISO Version number like '0.4.1' from the purchased product string 'ISO_0.4.1'
         $isoVersion = substr($product, 4);
+        // Set $isoMajor as the first number from $isoVersion
         list($isoMajor) = explode('.', $isoVersion);
-
+        // If that worked
         if ($isoMajor != null) {
+            // Set $currentMajor as the first number from the current release version
             list($currentMajor) = explode('.', $config['release_version']);
-
+            // If the purchased major matches the current major
             if ($isoMajor == $currentMajor) {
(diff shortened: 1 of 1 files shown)


api/download.php
```diff
@@ -22,40 +22,49 @@ function go_home() {
...
 
-$iso_version = false;
+$isoVersion = false;
 foreach ($products as $product) {
     if (substr($product, 0, 3) === 'ISO') {
-        $iso_version = substr($product, 4);
+        $isoVersion = substr($product, 4);
+        list($isoMajor) = explode('.', $isoVersion);
+
+        if ($isoMajor != null) {
```
lewisgoddard: What happens if this if fails?

api/download.php
```diff
@@ -22,40 +22,49 @@ function go_home() {
...
+        return go_home();
     }
 }
 
-$iso_version = false;
+$isoVersion = false;
 foreach ($products as $product) {
     if (substr($product, 0, 3) === 'ISO') {
-        $iso_version = substr($product, 4);
+        $isoVersion = substr($product, 4);
```
lewisgoddard: Is this literally taking the first 4 characters? That won't work with double digits, why not just handle them with explode like you do later?

lewisgoddard: Ah, okay, that's set in the product definitions, right?

api/download.php
btkostner: If we can't detect the major release version from the receipt, then we don't try to match it and only process the exact match. If for some reason it's like `ISO_release` then a purchase cookie for `release` gets set.

api/download.php
btkostner: This is for the release payment. When you buy it we add meta data to the charge in stripe with a line like `ISO_0.4.1` to indicate that the 0.4.1 release was bought.

PR #1733, opened 2017-05-28
Update docs regarding .desktop categories

danrabbit:
* Remove GTK because it's not a valid category
* Make sure people know they can add more than one category
* Capitalize Scratch because we're here and I saw it


btkostner:
@danrabbit GTK is a valid category. Should we still remove it (I think so)?

danrabbit:
Yeah TIL about the additional categories. I still think it's not a useful category in this example though since we expect all apps to be Gtk+

PR #1606, opened 2017-03-01
Correct typo in docs

seyDoggy:
The docs, under "Adding Translations", reads `make pot` but when I ran
this command it failed. I tried running with `make po` and that was
successful. I assume this is a typo.

Fixes #

### Changes Summary

- all instances of `make pot` replaced with `make po`.

This pull request [ is ] ready for review.

PR #1734, opened 2017-05-28
Correcting small typo

danieljsamson:
When going through tutorial I discovered that it needs to be make po instead of make pot

Fixes # 1

### Changes Summary
rename line.


lewisgoddard:
Duplicate of #1606

Also, this seems to be contested, our developers use `make pot`.
What system were you running that used `makr po`?

PR #1747, opened 2017-06-12
Spelling, MIME, … : A etc

comradekingu:


comradekingu:
@lewisgoddard 
The idea being that : Bigversal is always correct, and : smallversal sometimes isn't. It is a matter of choice in English to do so for some things, whereas for other languages not.

btkostner:
I'm going to merge this as all the changes seem good. If we have a reason not to capitalize a letter we can change it later.

comradekingu:
@lewisgoddard: The reason for always doing ": Bigversal" is because it is never in error.
You also end up with fewer errors in translation, as that tends to be a strict requirement elsewhere.

PR #1761, opened 2017-06-20
Clarify Title Case

cassidyjames:
- Link previous mention of Title Case to the section for consistency
- Add link to AP style
- Break rules into bullet points
- Clarify capitalization based on word length
- Remove ellipsis from third example (since we don't use ellipses in search anymore)
- Add example of a welcome widget since it's a super common use
- Add mention of English-specificity


docs/human-interface-guidelines.md
```diff
@@ -965,20 +965,26 @@ Used for labels and descriptive text.
...
 
-Capitalize the first and last words.
-All nouns, pronouns, adjectives, verbs, adverbs, and subordinate conjunctions (as, because, although) are capitalized.
 Used for titles, buttons, menus, and most other widgets.
 
 * ex. "Open File" title in a dialog.
 * ex. "Delete 13 Songs" on a button.
-* ex. "Search Contacts..." in a search bar.
+* ex. "Search Contacts" in a search bar.
+* ex. "No Videos Open" on a welcome screen.
```
btkostner: There should probably be some examples where not every first letter is capitalized.

cassidyjames: I thought about that but was having a hard time thinking of a frequent case like that.

PR #1765, opened 2017-06-24
Explain .desktop needs to be executable in Getting Started

cimm:
Adds a single sentence to the [Getting Started Guilde](https://elementary.io/docs/code/getting-started#the-desktop-file) explaining the `.desktop` file needs to be _executable_.

As far as I know the `.desktop` file will not be picked up by the system if it isn't executable. This might be common sense for experienced developers but took me way too long to figure out as a new Linux developer.

Not sure if we should go more in detail: does it needs to be executable by world for example or only by the user?

This pull request is ready for review but feel free to close the pull request if you disagree.


danrabbit:
Your .desktop file *does not* need to be made executable

PR #1767, opened 2017-06-29
add cache classes

btkostner:
This impliments some PHP cache classes for faster response times.

### Cache
The `Cache` class is a super simple with some helper functions.
The `CacheInterface` has some super basic functions so we can easily impliment other stores if needed.
The `FileStore` saves a new file for each key. It uses the OS temp directory.
Note: You will probably need to create the `/tmp/elementary/website` directory with permissions for cache to work.

### Slack
Because I didn't want to add this to the repo as dead code, I went ahead and implimented it on the `team.php` page. This lead to a larger PR than what I was expecting. Basiclly I put all of the slack code in `_backend/Service/Slack.php` for easier reading and better code splitting.

### End
As a result, we have an easy to use cache interface, and a much faster team page. Testing locally, the team page went from 1.2 seconds to 0.2 seconds. Luckely it's really simple to test if everything works. Just browse the team page :confetti_ball:


lewisgoddard:
Testing at https://beta.elementary.io/cache/team seems fine.

PR #1771, opened 2017-07-09
Insallation.json spanish typo

mochetts:
### Fixes
A typo in installation.js spanish translations.

### Changes Summary
`instalació` should be `instalación`


btkostner:
Please don't manually change translations in the json files. Those files are pulled directly from our weblate instance. If it is updated there (and it is), then it will be updated on the website.

mochetts:
Ok. Being thankful never hurts though.

PR #1772, opened 2017-07-11
Fixed a typo in Notifications (Making Preparations)

Tomires:
The folder's name should say 'build' instead of 'CMake'

### DIFF
--- a/docs/code/getting-started.md
+++ b/docs/code/getting-started.md
@@ -663,7 +663,7 @@ By now you've probably already seen the white notification bubbles that appear o
 ## Making Preparations {#making-preparations}
 1. Create a new folder inside of  "~/Projects" called "notifications-app"
 2. Create a file inside called ```notify-app.vala ```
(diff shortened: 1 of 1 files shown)


Philip-Scott:
I think cmake folder is correct, but maybe "re-create" isn't the right word. Because you do need the contents of this folder to be able to build, but you can just copy and paste your old one

Tomires:
screenshot from 2017-07-11 14 50 23

In that case, wouldn't it be better to remove the quotation marks (and in turn the rectangle)? I did not encounter a folder specifically called 'CMake' over the course of the tutorial (correct me if I am missing something obvious please).

I don't have any issues with the "re-create" term as the user needs to change a couple strings in the process (if they follow the tutorial's naming conventions).

danrabbit:
Yeah I think what this is is leftovers from when we used to have people put the cmake modules inside their projects instead of using the package. So this can go away and it can just say "Create a CMakelists.txt"

PR #1795, opened 2017-08-19
Move to new jsDelivr endpoints

lewisgoddard:
### Changes Summary

- Changed from old jsDelivr endpoints to new ones.

### To Test

- [ ] Test Chart
- [ ] Code Highlighting
- [x] Test Modals

Building at https://beta.elementary.io/jsdelivr-2/


btkostner:
So the code block doesn't seem to be an issue with us. Looking at the docs nothing important for us has changed. But jsdeliver is giving me a `Uncaught SyntaxError: Unexpected token (` when trying to load that file.

MartinKolarik:
@btkostner which file is giving you the error?

btkostner:
@MartinKolarik on loading of a doc page, looking at the console, `Uncaught SyntaxError: Unexpected token (` is thrown. Chrome says it's from loading highlightjs vala file.

MartinKolarik:
@btkostner do you mean https://beta.elementary.io/jsdelivr-2/ or some other page?

btkostner:
A docs page like `https://elementary.io/docs/code/reference#reference`, but our beta domain doesn't have the docs setup, so you will need to try it locally.

MartinKolarik:
Oh I see... [This file](https://cdn.jsdelivr.net/gh/isagalaev/highlight.js@9/src/languages/vala.js) does indeed contain a syntax error.

lewisgoddard:
Seems like that repository isn't built to host a distribution packages, which doesn't allow us to use the new jsDelivr format just yet. We should either ask them to carry that or just use the builds repo I've already change it to.

btkostner:
I'm down with using your repository for highlightjs, but once the issue is fixed we will wan't to use upstream again.

lewisgoddard:
Okay, seems like that's got it working for now. I'll open an issue to see if upstream are willing to work around it.

PR #1797, opened 2017-08-25
Developer docs: Use Gtk.Application throughout

danrabbit:
Seems like it would be more straightforward to use Gtk.App from the very beginning instead of changing things halfway through

This also moves the first compile to right after creating the application. The point of this is moving that “yatta” moment sooner and reducing the chances of messing up early on. This way the reader can test their changes in smaller increments 

I wrote this in the airport lobby, so please test it. I'm sure it has issues.


btkostner:
I summon the power of @elementary/desktop-developers

docs/code/getting-started.md
```diff
@@ -108,53 +108,69 @@ Apps on elementary OS are organized into standardized directories contained in y
...
 
-## Gtk.Window and Gtk.Button {#gtk-window-and-gtk-button}
+## Gtk.Application {#gtk-application}
 Now what you've been waiting for! We're going to create a window that contains a button. When pressed, the button will display the text "Hello World!" To do this, we're going to use a widget toolkit called GTK+ and the programming language Vala. Before we begin, we highly recommend that you do not copy and paste. Typing each section manually will help you to practice and remember. Let's get started:
 
 1. Create a new file in Scratch and save it as "gtk-hello.vala" inside your "src" folder
 
-2. First we must create the main function of our new GTK app. Type the following in your "gtk-hello.vala".
+2. First we create a special class called a `Gtk.Application`. `Gtk.Application` is a class that handles many important aspects of a Gtk+ app like app uniqueness and the application ID you need to identify your app to the notifications server. If you want some more details about `Gtk.Application`, [check out Valadoc](https://valadoc.org/gtk+-3.0/Gtk.Application). For now, type the following in your "gtk-hello.vala".
```
jeremypw: Mixed and repeated use of "app" and "application".

docs/code/getting-started.md
```diff
@@ -108,53 +108,69 @@ Apps on elementary OS are organized into standardized directories contained in y
...
+            }
         }
+        
+    Ready to test it out? Fire up your terminal and make sure you're in "~/Projects/gtk-hello/src". Then execute the following commands to compile and run your first Gtk+ app:
 
-3. Now, that we've initialized Gtk, we'll create a new window and declare a few properties of this window. You'll notice that most of these property names are pretty straightforward. Try to guess what each one does and we'll explain in a second. Type the following after the `Gtk.init` line, but before the last bracket:
+        ```bash
+        valac --pkg gtk+-3.0 gtk-hello.vala
+        ./gtk-hello
+        ```
```
jeremypw: I got warning "Gtk-WARNING **: gtk_widget_size_allocate(): attempt to allocate widget with width -13 and height 17
" and only a small window - is that intended?

docs/code/getting-started.md
```diff
@@ -108,53 +108,69 @@ Apps on elementary OS are organized into standardized directories contained in y
...
 
-        var window = new Gtk.Window ();
-        window.title = "Hello World!";
-        window.set_border_width (12);
-        window.set_position (Gtk.WindowPosition.CENTER);
-        window.set_default_size (350, 70);
-        window.destroy.connect (Gtk.main_quit);
+    Did it work? If so, congratulations! If not, read over your source code again and look for errors. Also check the output of your terminal. Usually there is helpful output that will help you track down your mistake.
```
jeremypw: Should specify what "work" means - what should the user see?

docs/code/getting-started.md
```diff
@@ -108,53 +108,69 @@ Apps on elementary OS are organized into standardized directories contained in y
...
 
-        Gtk.main ();
-        return 0;
+3. Now, that we've created a new `Gtk.Application` with a `Gtk.ApplicationWindow` called `main_window`, let's declare a couple properties of this window. You'll notice that most of these property names are pretty straightforward. Try to guess what each one does and we'll explain in a second. Add the following properties to your `main_window`:
```
jeremypw: Should specify exactly where to type these lines (within the activate function) rather than inferring it from the repeated line.

docs/code/getting-started.md
```diff
@@ -207,16 +223,16 @@ To create our first real app, we're going to need all the old stuff that we used
 
 3. Create "hello-again.vala" in the "src" folder.  This time we're going to prefix our file with a small legal header. More about legal stuff later. For now you can copy [the GPL header from our reference documentation](https://elementary.io/docs/code/reference#gpl-header). Be sure to assign the copyright to yourself at the top of the header and change the author to you at the bottom of the header.
 
-4. Now, let's create our main function, a `Gtk.Window`, and set the window's default properties. Refer back to the last chapter if you need a refresher.
+4. Now, let's create a `Gtk.Application`, a `Gtk.ApplicationWindow`, and set the window's default properties. Refer back to the last chapter if you need a refresher.
 
 5. For the sake of time let's just put a `Gtk.Label` instead of a `Gtk.Button`. We don't need to try to make the label do anything when you click it.
 
         var label = new Gtk.Label ("Hello Again World!");
```
jeremypw: For consistency should use more specific variable name for label.

docs/code/getting-started.md
```diff
@@ -584,11 +600,11 @@ Let’s create a Window with a vertical Grid that contains a Button and a Label:
     grid.add (button);
     grid.add (label);
```
jeremypw: More specific variable names?

docs/code/getting-started.md
```diff
@@ -584,11 +600,11 @@ Let’s create a Window with a vertical Grid that contains a Button and a Label:
...
 
-    window.add (grid);
+    main_window.add (grid);
```
jeremypw: Center the grid?

docs/code/getting-started.md
danrabbit: I didn't change this in my branch

docs/code/getting-started.md
danrabbit: This isn't a change that comes from my branch

docs/code/getting-started.md
```diff
@@ -108,53 +108,69 @@ Apps on elementary OS are organized into standardized directories contained in y
...
+        var main_window = new Gtk.ApplicationWindow (this);
+        main_window.title = "Hello World!";
+        main_window.set_default_size (350, 70);
 
-4. Now that we've defined a nice window, let's put a button inside of it. After our window stuff (but before `Gtk.main` line), leave a new line and then type the following:
+        main_window.show_all ();
+
+    The first line creates a new `Gtk.ApplicationWindow` called `main_window`. The second line sets the window title that you see at the top of the window. We also must give our window a default size so that is does not appear too small for the user to interact with it. Compile and run your application again and see how it's changed from the first time.
```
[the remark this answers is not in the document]
danrabbit: This paragraph describes what reader should see

> The second line sets the window title that you see at the top of the window. We also must give our window a default size so that is does not appear too small for the user to interact with it.

docs/code/getting-started.md
danrabbit: Outside the scope of this branch. I haven't touched the grid here

docs/code/getting-started.md
```diff
@@ -108,53 +108,63 @@ Apps on elementary OS are organized into standardized directories contained in y
...
+                main_window.default_height = 300;
+                main_window.default_width = 300;
+                main_window.title = "Hello World";
+                main_window.show_all ();
+            }
+
+            public static int main (string[] args) {
+                var app = new MyApp ();
+                return app.run (args);
+            }
```
benwaffle: Why not place `main()` outside the class? I think it looks cleaner
```Vala
int main (string[] args) {
    return new MyApp ().run (args);
}
```

Issue #1770, opened 2017-07-05
Add "clicked" guidance to code reference


cassidyjames:
GTK widgets respond to "clicked" which is a press + release; for example, this lets you drag a window from a button without triggering the button. We should consider mentioning this in our code reference as it seems to be a recurring (https://github.com/elementary/granite/issues/39, https://github.com/elementary/granite/issues/82) issue.

JackieCalapristi:
Is it alright if I try to take this one? It would be my first contribution to OS! :)

btkostner:
You don't have to ask permission, just comment on the issue so people don't also start working on it :wink:

JackieCalapristi:
Oh good to know! Ok, I'm working on this one! 😃

JackieCalapristi:
I've created a [pull request](https://github.com/elementary/website/pull/1798) that includes "clicked" guidance in reference.md!

PR #1798, opened 2017-08-25
Include GTK "Clicked" Considerations

JackieCalapristi:
### Changes Summary
Updated reference.md to include information about a recurring Gtk issue for future contributors to take into consideration.

Fixes [#1770](https://github.com/elementary/website/issues/1770)

This pull request is ready for review! 😃


docs/code/reference.md
```diff
@@ -256,3 +256,10 @@ Don’t forget to make your reports compelling. It’s up to you to sell the mer
...
+
+# Other Considerations {#other-considerations}
+
+Here's some other information to take into account before working on elementary applications.
+
+## Gtk Recurrent "Clicked" Issue
+Elementary applications utilize [Gtk](https://www.gtk.org/) (i.e. A GIMP toolkit for GUIs). Gtk widgets are intended to respond to click events that can be described as "press + release" instead of "press". There's been a recurrent issue (see [here](https://github.com/elementary/granite/issues/39) and [here](https://github.com/elementary/granite/issues/82)) in which elementary projects would activate on "press" which has since been patched.
```
btkostner: `elementary` is never capitalized

JackieCalapristi: Oh good to know! Should this be added to `reference.md` and/or `human-interface-guildlines.md`? It might be helpful to know not capitalize elementary even when it's at the beginning of a sentence.

JackieCalapristi:
I like the idea of putting it in a non-other section! Totally agree that "other" sections can get crowded.

docs/code/reference.md
lewisgoddard: This is documented at https://elementary.io/brand

I don't think it needs to make its way into the HIG as well.

JackieCalapristi: @lewisgoddard It's already in the HIG as [seen here](https://github.com/elementary/website/blob/62b751006961549634e4a0a15d52c5013af130fa/docs/human-interface-guidelines.md#notesexceptions-notes-exceptions)
```
Proper nouns should always be capitalized properly; this means that, for example, Google should always be shown as "Google," elementary should always be shown as "elementary," and MPEG should always be shown as "MPEG."
```
I was suggesting adding a caveat that said something like "elementary should always be shown as elementary, even at the beginning of sentences". But good call on referencing the lowercase mention in https://elementary.io/brand! I'll definitely reference that in the future.

btkostner: That would be a good idea, but that sentence is getting really wordy as it is. I would probably break it up to something like:

> Proper nouns should always be capitalized properly; this means that, for example, Google should always be shown as "Google," and MPEG should always be shown as "MPEG." One major exception is "elementary." It should follow the [brand guidelines](https://elementary.io/brand) and always appear lowercase, even at the start of a sentence.

ava-jacquelinec: Good call! I like your phrasing better.

PR #1800, opened 2017-08-30
Giving original logo color to .nav icons when hovered or focused

PAPERPANKS:
### Changes Summary

-  `.nav` icons get colored on hover or focused
-  removed `opacity` property from `.nav` icons when hovered 

### DIFF
--- a/_styles/main.css
+++ b/_styles/main.css
@@ -299,8 +299,8 @@ nav a:visited:hover {
 
 nav a[title="Facebook"]:focus,
 nav a[title="Facebook"]:hover {
-    color: #19227E;
-    fill: #1a237e;
+    color: #19227e;
+    fill: #19227e;
 }
 
 nav a[title="Google+"]:focus,


_styles/main.css
```diff
@@ -288,16 +288,70 @@ nav a:visited {
 }
 
 nav a:focus {
-    color: #08c;
-    fill: #08c;
```
btkostner: Please put these back in. We still want the left side links to change when in focus.

_styles/main.css
```diff
@@ -288,16 +288,70 @@ nav a:visited {
...
     text-decoration: none;
 }
 
-nav a:hover,
-nav a:visited:hover {
```
btkostner: Also keep this for the same reason as above.

_styles/main.css
```diff
@@ -288,16 +288,70 @@ nav a:visited {
...
+nav a:visited{
     opacity: 0.7;
 }
 
+/*color changes of .nav icon on focus*/
```
btkostner: Please remove this comment, it's pretty easy to tell what the code does.

_styles/main.css
```diff
@@ -288,16 +288,70 @@ nav a:visited {
...
+    color:#FF4500;
+    fill:#FF4500;
+}
+
+nav .right li a[title="StackExchange"]:focus{
+    color:#2196f3;
+    fill:#2196f3;
+}
+
+/*color changes of .nav icon on hover*/
```
btkostner: Also remove this comment

_styles/main.css
```diff
@@ -288,16 +288,70 @@ nav a:visited {
...
+
+nav .right li a[title="Facebook"]:hover{
```
btkostner: There is no need to specify all of these classes again. You should combine them with the above, so it reads like:
```
nav a[title="Facebook"]:focus,
nav a[title="Facebook"]:hover {
    color: #01579b;
    fill: #01579b;
}
```

_styles/main.css
```diff
@@ -288,16 +288,70 @@ nav a:visited {
...
 
-nav a:hover,
-nav a:visited:hover {
+nav a:visited{
     opacity: 0.7;
 }
 
+/*color changes of .nav icon on focus*/ 
+
+nav .right li a[title="Facebook"]:focus{
```
btkostner: This selector is over qualified. I would trim it down to `nav a[title="Facebook']:focus`

_styles/main.css
PAPERPANKS: Can't we have original logo color for icon when focused. If I do the same as mentioned all icon will have same color `#08c` and this can seem monotonous to users

btkostner: These won't effect the social logos on the right side because they get overwritten below, but this does effect the left side of the navbar. Without it, there is no feedback when a user hovers over "Store" or "Support."

lewisgoddard:
Congrats to @PAPERPANKS on their first GitHub Contribution! :tada: :balloon: :birthday:

PR #1801, opened 2017-09-01
Changing placeholder for amount input field

PAPERPANKS:
Making some changes in vocabulary of the website

### Changes Summary

- The value of `placeholder` 'custom' is changed to 'Pay As You Wish' giving user a more clear view.

- `<p class="small-label...../p>` of amount input field mentioning the dollar ($) symbol.


btkostner:
I'm honestly :-1: on this change. "Pay what you want" applies to that whole section, so putting it as the custom amount placeholder doesn't make sense. There is also already a title for that section that says "Pay What You Want."

Unless there is a solid reason why the current implementation doesn't work, or could be better, this section will stay the same.

PR #1808, opened 2017-09-03
Wi-Fi and explanation of connecting.

comradekingu:


comradekingu:
@btkostner Good point. I tried to rewrite it again following your advice.

Emphasis on these things:
- Much shorter, taking less time to read.
- Shorter sentences
- Less complex structure. reducing time to read further.
- "Wi-Fi" spelt correctly.
"If it detects there is not an" this should have been "no", but it is omitted, so one less error there. 
- Does away with explaining "captive portal", which opens a battle on a hill not worth fighting, even if a winning one. There is no need to understand, nor bring that understanding with you to grasp the final point, which is getting the idea, and closing the page.
- Much much easier to translate, meaning more localization coverage.

https://en.wikipedia.org/wiki/Captive_portal gives an idea of what languages have coverage. Which is to say maybe none, and the ones that do, have anglofied or directly adopted words. I had to make up a word, which isnt't a bad idea, but (at best) puts the burden of understanding a new word on the user.

comradekingu:
@danrabbit 
I think this approaches the level of involvement one would wish to acquire.

comradekingu:
@ danrabbit
OK, 3 words added.

danrabbit:
I think clarity and natural grammar is much more important to optimize for than number of letters. This reads very awkwardly

comradekingu:
An awkward read it may be, did you mean to suggest it isn't still leaps and bounds better than the original?
I could annotate my view of what that was, it appears to have escaped you.

danrabbit:
I’m not sure it’s better than the original no. I think it’s much less clear and harder to read

comradekingu:
You're connected! 
**Shouting at the user**

- The distinction between being connected and having a Internet connection still not established.

Your Internet connection appears to be working.
**Introducing doubt**

- This is almost what the user wanted to know.

You can safely close this window and continue using your device.

**Redundancy**
- "Safely" doesn't do anything, besides entertaining the idea that there is a danger afoot, which would warrant its use. The whole sentence is actually redundant. The user could knowingly have closed this message, knowing all there is to know, with the confidence instilled from communicating this clearly, but is instead compelled to read on.

Why is this window appearing?

**breaking the 4th wall**.
- Is there anything important enough about the above warranting communicating this in a manner which suggests it shouldn't be? Let's find out.

elementary OS automatically 

**Self-referencing** 
This forms a 
**garden sentence**
made worse by the placement of the word "automatically".

checks your Internet connection

- OK, it checks the internet connection, in an automatic fashion, then what?
 
 when you connect to a new WiFi network.
**spelling error**

- Well, of this the user is now informed, left to their own doubt as to why that is. What is clear here is that it has to be a Wi-Fi network, and that it has to be new for the check to happen. Right.

If it detects there is not an Internet connection
**spelling error**
 (i.e. if you are connecting to a captive portal at a coffee shop or other public location), 
**factual inconsistency**

- It doesn't have to be a public location, but the user is more concerned with why the automatic lookup is made, by the self-referencing OS, looking the spelling error, to question what there being a captive portal has to do with it all. Also, what is a captive portal?

this window will appear and display the login page.
**Breaking 4th wall**
- User still doesn't know what a captive portal is, and this is a self-reference to an evident description of something that has already happened.

Some networks can appear to be a captive portal at first, triggering this window,
**Repetition of the same thing**

- Still don't know what a captive portal is. Right, so the window appears, still evident.
It does though follow, that the window and the captive portal are linked, on a network that is public.

 then actually end up connecting. In those cases, you'll see this message and can safely close the window.
**Repetition of the same thing**

- The only thing the user knows about a captive portal, is that it triggers the message.

This is actually the third time the user has been informed about the appearance of the same evident thing. That it can be closed, is a second time coming, but at least now it can be established that nothing was actually wrong. It feels like a long time since the leisurely stroll into this garden sentence started, we have taken detours, been misguided, and arrive at next to nothing, to form a garden in which everything was spaghetti all along.

If you only ever use a captive portal network once, it would be easier to just comply with logging in as per requested.
----
I know what captive portal assistance is, still I had trouble understanding, and more yet in translating.
For the real test, use one of those translated, or non-translated as the case usually is with very complex strings, use mine and the result will still not favour the outcome, on someone that is a new user. Worst case that means non-translated for someone that doesn't speak English. Very likely it means badly translated if at all, and best case it is an English user, for whom the string is at least native.

Try that, report back, and tell me the new version is "much less clear and harder to read". I find it, unconvincing. Please present evidence like the above to the contrary, I am all ears.

comradekingu:
Current new strings:

Automatic connection to the Internet established

- Formal, impersonal tone used, as is convention with notifications.
OK, the user knows that the "Internet works", that it was automatically achieved. Cool beans, really only that the Internet worked was what there was to know. How that happened is now left to explain.

Gaining access 
- To the Internet, presumably
on this network may have required logging in, 
- The user is on a network, for which Internet was attained, automatically, and getting that all to work was predicated on logging in. Understanding all of this actually. Then again I wrote it. **4th wall needs looking into**

but it was possible to do without anyhow.

- (Still kind of formal, but more verbose.)
Right, so in the odd event that it wasn't needed, this is just some annoying pop-up. Well, still you are friendly elementary OS. Thank you for not getting me confused, nor wasting my time when i wanted to use the Internet in all its glory. Presumably that was my main intent. Yeees.

btkostner:
elementary does not use success dialogs because they are very redundent and usually offer no benefit to the user. This page is no exception. Sadly due to technical limitations we have to show it sometimes. Therefor we want to make sure the user understands why they are seeing it, and what this means to them.

Following your criticisms I would suggest do something like this.

```
<h1>You're connected.</h1>
<p>Your Internet connection is working as expected. You can close this window and continue using your device.</p>
<h2>Why is this window appearing?</h2>
<p>Your Internet connection is automatically checked when you connect to a new network. If it detects there is no connection, or if a login is required, this window will appear.</p>
<p>Some networks can appear to require a login, triggering this window, then actually end up connecting. In those cases, you'll see this message and can safely close the window.</p>
```

The first section explains what happened, and what to do next. For most users, this is all they care about. It's quick, easy to read, and to the point. If they want to use the internet, the can close it now. The second section is more for the currious type. It explains why they see this dialog, something that shouldn't happen normally in elementary, and reiterates that there is nothing wrong going on.

We would rather have more information about what is going on, than have too little and confuse the user about what is going on.

comradekingu:
There would no need to do tiered responses if the burden of reading the second part wouldnt exceed what could have been a first part that answered any and all valid questions.

It doesn't answer my criticism. What you did was use the very lowest fruit to polish what was. You can thus, reiterate the bulk of it, without those parts. I will do so with your words in mind.
This is not a success dialog, it is an informational tidbit. The distinction between which, is the reason why the old version, even polished, doesn't meet my criticism, nor does it comply to your explanation of it.

> h1>You're connected.</h1>

This doesn't explain what happened. You connect to a network, as per ususal, you can see it. Same way every time.

> <p>Your Internet connection is working as expected.

"As expected" is fully redundant. See how it all falls apart when you boil it down to the content it conveys? What users think of when you say "your internet connection" is their internet connection. That is not being used here. This is someone elses internet connection, being shared over a network the user is connected to.

> You can close this window and continue using your device.</p>

If this was the end of it, and it could be, then this whole sentence is redundant. Without the word "safely" it falls apart too. There is no implication of anything here. Regular messages don't communicate that they can be closed. For the same reason a "success" isn't anything over the ordinary. Two adverbs the world of dialogs that don't need them can do (better) without.

> <h2>Why is this window appearing?</h2>

The only thing curious about this, is why a message, after presenting itself as closable, goes on. Onto explaining that it is a message. If your user doesn't understand these two things, you have no user. Meaning for everyone else, or all users at it were. Time is wasted. There is no cost incurred to doing away with this, nor can the motive be questioned. It is fully redundant.

>  <p>Your Internet connection is automatically checked when you connect to a new network.

Why would it check the users internet connection, when this check is implemented for the users connection to the internet. That is at worst different. Why does the user need to know there is a check? This reads like annotated code.
Think about what kind of questions this raises. Checked for what? Read on to see where you get an answer, or why it has to do with being connected.

> If it detects there is no connection, or if a login is required, this window will appear.</p>

Yes, this reads like commented code.
No connection to what? Login to do what? And what was the connection checked for. Guess it was for a connection of sorts.

So we derive that there was no connection, because the window is there. The user was already told that it was all working. So what the actual?

> <p>Some networks can appear to require a login, triggering this window, then actually end up connecting.

Why is the user seeing this window _now_. And what does that mean?
Again, connect to what? Clearer this time, but still implied for no good reason.
So not on no connection, but on a subroutine of requiring a login, for when a network doesn't present one, but the ??? sees it as one anyway.

Remember the user doesn't know what captive portal assistance is.
This makes it seem like the OS doesn't know what it is doing. It makes a point of it. There is zero value to that. It is negative.

Furthermore, if there was no login required, and the window is still there. And the user doesn't know that isn't the reason why it is there now.

If so, what does it connect to, for what reason, and is it connected now? If it wasn't all working as said, and it wasn't to the contrary not working, as said, but was actually working "actually", for this third option.

There is no clarity as to why this dialog is there now. It is just inconclusive spaghetti. 

What happened? Is it a VPN? It could very well be. Because there is no good reason why all options aren't on the table now.

> In those cases, you'll see this message and can safely close the window.</p>

So the user can close the window, is required to log in, or not. Clear as mud.

Nothing about that explained anything of value over the new, and it certainly wasn't quick at doing so.

I get that this isn't supposed to happen. But it is the least gracious fail in the whole OS.

> have more

yes, a lot more.

> information about what is going on

No, none of that. Please argue to the contrary. Don't just say it does.

> than have too little

The is a false equivalency. If the amount of text that is extra, doesn't have any content, then it is too much, and no amount of that could help with the problem too little.
Less is more if more is baggage.

There would be no reason to reiterate if it was clear the first time. If there wasn't a reason created why the reiteration wouldn't be needed, it all falls apart. They are both redundant, because they are all the structure leaning against itself to not fall down.

>and confuse the user about what is going on

I dare you to present this to an actual user.
One that doesn't speak English, in English. How this is not a fail is beyond me.
One that speaks English. Then someone who speaks English or American English natively. This would exclude all of us.
A quantity of one is a representative sample-size here, because true positives don't exist in that amount. Developers excluded of course.

You would both learn something. I've seen great work in elementary OS, this is not on the show-reel in that respect. Dig in the heels when there is something to lose.

btkostner:
> There would no need to do tiered responses if the burden of reading the second part wouldnt exceed what could have been a first part that answered any and all valid questions.

By putting all of the information in the first part you lose clarity and easy reading. The first part explains what you can do now, and if the user doesn't care about the why, then they can just continue without reading it.

> This is not a success dialog, it is an informational tidbit. The distinction between which, is the reason why the old version, even polished, doesn't meet my criticism, nor does it comply to your explanation of it.

It is a success dialog. Everything is working. There is nothing you can do but click the close button. The point of the text is to provide a good user experience and explain why they are seeing the (hopefully) only success dialog in elementary OS. As such, it should offer a good middle ground between only reading `success` and offering a wikipedia page.

>> You're connected.

> This doesn't explain what happened. You connect to a network, as per ususal, you can see it. Same way every time.

It's the title. It doesn't need to explain everything.

I can see this being an argument as to why you should never see this dialog (which I agree), but sadly it's not possible currently. So a lot of this will be (to some) pretty obvious. This also is used to identify why this is showing up. Is it because I opened my music player, or because I tried connecting to a network? Unless you plugged in something the exact moment this showed up, it's going to be pretty obvious that `connected` means connected to the network. If there was doubt, the first sentence makes it even more clear.

> > Your Internet connection is working as expected.

> "As expected" is fully redundant. See how it all falls apart when you boil it down to the content it conveys?

Fair. `Your Internet connection is working` would also work.

> What users think of when you say "your Internet connection" is their Internet connection. That is not being used here. This is someone elses Internet connection, being shared over a network the user is connected to.

If you are at a coffee shop, and a window said `Your Internet`, I doubt the user would think the connection at their house. In English, most sentences assume ownership. If you loan a car, people don't say `the car currently in your possession.` It's `your car.` We could rephrase it as `The Internet connection you are currently connected to`, but that gets too wordy. If it's different in a translated language, then it should be translated in a way to make it seem normal.

>> You can close this window and continue using your device.

> Regular messages don't communicate that they can be closed. For the same reason a "success" isn't anything over the ordinary.

True, but this is the only success dialog in elementary. Other messages have clear actions like `Delete` or `Undo`. Notifications are a clear, widely used, situation. This is not. Therefor it should give a clear next step to the user.

>> Why is this window appearing?

> The only thing curious about this, is why a message, after presenting itself as closable, goes on. Onto explaining that it is a message. If your user doesn't understand these two things, you have no user. Meaning for everyone else, or all users at it were. Time is wasted. There is no cost incurred to doing away with this, nor can the motive be questioned. It is fully redundant.

Situation 1: We present it as closable at the end. The user reads all of the information and _then_ knows there is no danger in closing the window. User gets frustrated that we wasted their time.

Situation 2: We don't explain (very well) why the window shows. The user closes the window being confused why a dialog says everything is ok.

Situation 3: We give a simple explanation on what's going on, and that the user can close the window. If the user understands, or has seen this window before, they can close it right now (if not before). If not, they can continue reading and understand why this window window appears when trying to connect to a coffee shop network.

>> Your Internet connection is automatically checked when you connect to a new network.

> Why would it check the users Internet connection, when this check is implemented for the users connection to the Internet. That is at worst different. Why does the user need to know there is a check?
Think about what kind of questions this raises. Checked for what? Read on to see where you get an answer, or why it has to do with being connected.

I don't understand the first sentience. You are correct that the user doesn't need to know there is a check, but it's the reason this pages appear, and unavoidable if you want to explain.

`Your Internet connection is automatically checked for login pages (and pay walls?)`

Really, there is no way of explaining what it's checking without being 1) repetitive, or 2) technical.

`Your Internet connection is automatically checked for connection to the Internet`
`Your Internet connection is automatically checked for accessibility to <ip address>`

>> If it detects there is no connection, or if a login is required, this window will appear.

> Yes, this reads like commented code.

`true`

> No connection to what? Login to do what?

The network. The Internet. The whole thing this window is talking about, mentioned multiple times above.

> So we derive that there was no connection, because the window is there. The user was already told that it was all working. So what the actual?

Fair. That sentience needs rewording.

>> Some networks can appear to require a login, triggering this window, then actually end up connecting.

> Why is the user seeing this window now. And what does that mean?

Exactly what this sentience is talking about, and it means the first section of this window. `Your Internet connection is working as expected.`

> Again, connect to what? Clearer this time, but still implied for no good reason.

Again, the Internet. It's not implied, it's specifically stated multiple times.

> So not on no connection, but on a subroutine of requiring a login, for when a network doesn't present one, but the ??? sees it as one anyway.

`You can close this window and continue using your device.`

> There is no clarity as to why this dialog is there now. It is just inconclusive spaghetti.

Besides the whole section as to `Why is this window appearing?`.

> Nothing about that explained anything of value over the new, and it certainly wasn't quick at doing so.

There shouldn't be anything you are considering value to explain here. We aren't giving top secret, need to know, information about the connection. We have to show this window, and so we need to explain why the user is seeing it, and what to do next.

> than have too little
The is a false equivalency. If the amount of text that is extra, doesn't have any content, then it is too much, and no amount of that could help with the problem too little.
Less is more if more is baggage.

If the amount of text that is less, doesn't have any content, then it is too little.

> There would be no reason to reiterate if it was clear the first time. If there wasn't a reason created why the reiteration wouldn't be needed, it all falls apart. They are both redundant, because they are all the structure leaning against itself to not fall down.

There is perfect reason to reiterate. I could put a debug log message in the window, and only a very few percentage of the people seeing the window would understand. Reiterating helps people understand, and it prevents a huge TOS like wall of text where everyone needs to read to the end to understand what to do next.

Example 1:
```
You're connected!

Your Internet connection appears to be working. elementary OS
automatically checks your Internet connection when you connect
to a new WiFi network. If it detects there is not an Internet connection
(i.e. if you are connecting to a captive portal at a coffee shop or other
public location), this window will appear and display the login page.
Some networks can appear to be a captive portal at first, triggering
this window, then actually end up connecting. In those cases, you'll
see this message and can safely close the window. You can safely
close this window and continue using your device.
```

Example 2:
```
You're connected!

Your Internet connection appears to be working.
You can safely close this window and continue using your device.

Why is this window appearing?
elementary OS automatically checks your Internet connection when you
connect to a new WiFi network. If it detects there is not an Internet
connection (i.e. if you are connecting to a captive portal at a coffee
shop or other public location), this window will appear and display the
login page.Some networks can appear to be a captive portal at first,
triggering this window, then actually end up connecting. In those cases,
you'll see this message and can safely close the window.
```

---

> Automatic connection to the Internet established.

Ok. Why am I seeing this window this one time instead of the 500 times I've connected to the Internet?

> Your Internet connection appears to be working.

`appears` introduces doubt. We have connected to elementary.io, there is no doubt that the Internet is currently working.

From your previous comment:

Introducing doubt
This is almost what the user wanted to know. Does it work, or does it not?

> but it was possible to do without anyhow.

Weird English.

Overall the change has less words, yes, but it misses the "what to do next" step. Without it, the user has to read the whole window before realizing it has no next step and they should probably just close it and hope for the best. The current window has the first paragraph that ends with the next step. At that point there is no question what the user should do next, but if they need more info they can continue reading.

It's not just word count that makes things go fast. It's the time it takes for the user to be confident they know what to do next.

```
You're connected
You can safely close this window and continue using your device.
```

82 characters and the user knows what to do next. If they still have questions that can continue reading.

Lastly, something really picky that I doubt is mentioned anywhere. The title should be around 30 characters (usually not a full sentience). This keeps things short and helps with readability with that font.

comradekingu:
>It's the title. It doesn't need to explain the whole thing.

It being a title does not warrant wasting the time nor the space.
If you agree time is spent at the rate one reads, and understood given by the complexity of the content, then it follows that length and complexity are factors better kept to a minimum. If you go below hurting one with the other, then it isn't less than minimum.
Why would be a point of discussion, but against "you are connected", it starts at -3 words and one lineshift. "Internet connection now works" "Internet status: Connected" or similar is better, if an introduction is really needed.

If you remove "as expected", then you have truly said the same thing twice. Which to an ideal of "once, clearly", means wasting one iteration.

>By putting all of the information in the first part you lose clarity and easy reading.

Do argue as to why. Theoretically it is better than putting it multiple places, in folding and ever more thought-provoking levels.

> The first part 

Is longer and more convoluted than the other alone. It ends there.

>explains what you can do now, and if the user doesn't care about the why, then they can just continue without reading it.

This is a usability error. If you read all of it, and some users do, you will have confused a great deal of those. Meaning it makes less sense overall to people at large. It works as intended, for fewer people. At best time is wasted for those that read all.

>It is a success dialog.

OK. I would argue "success" alone is better, but that isn't on the table, and we agree there. What I do not understand is why "If it detects there is no connection, or if a login is required, this window will appear." makes any sense in a success dialog.
The user clicks close in the hope that it works, not in the confidence of knowing. That is worse.

> If you are at a coffee shop, and a window said Your Internet, I doubt the user would think the connection at their house.

You forgot, "for the first time". Mess up the possessive nouns, in languages where they differ, and that is what you have in some cases. Why would the user not think logging into their network isn't needed? The whole thing reads as undistinguishable from VPN.
Remember all of those "the operating system follows the machine"-people. This is what actual users are mislead to think. The difference between "your" and "_your_" is present in English.

>True, but this is the only success dialog in elementary. Other messages have clear actions like Delete or Undo. Notifications are a clear, widely used, situation. This is not. Therefore it should give a clear next step to the user.

Extremely good point. It makes no sense to X-close a dialog that is purely informational.
X closes, it does not continue. In every other desktop OS (?), the default behaviour is to throw up a button that says "OK". If the only default action is "OK", then we derive that what it says is of semantic value. I used windows XP back in the day, and have had the misfortune to later have taught people to use various variants of Windows in a job setting. The poweruser, me included, clicked "OK" faster than the message could be read, or, for the regular user, read the message, get scared and bring it in.

It being the only is good. However, any bad experience is the once-over. People remember, and they adapt. One face-plant loses users, for good. The user is already procrastinating thinking they are not up for the challenge of something new. The first amount of confidence is crucial. And this message will knock down any established confidence a good bit. At the wrong time, it is game over.

Situation 1: 2: and 3: You present the message as closable. That wastes time.
It wastes time once, and twice as much time twice.
If you close a message, it is gone, and you may have done something wrong. Users are terrified of that. They click the things they know. This is a new thing, in a new situation.

>> Why would it check the users Internet connection, when this check is implemented for the users connection to the Internet. 

I could have written "_users_ internet connections" or "user's". But i instead relied solely on the context of the second sentence, where I also forgot it. My bad.
My point was the user does not necessarily know the difference between those, and shouldn't be mislead to use poor judgement to decide which is which.
It becomes a multirelational thing very quickly, with many different written, and conflicting outcomes to confuse that.

> >Again, connect to what? Clearer this time, but still implied for no good reason.
>Again, the Internet. It's not implied, it's specifically stated multiple times.

This is a good example. Just because it made sense every time you knew it would, does not mean it conveys that information.

> checks

There is no real need to explain how something is done, when you can state the outcome and be done with it. I could have said "how _it_ does something", but i didn't, to prove a point. If the outcome is always the same, the premise is irrelevant to all but those that want to second guess it. That is the opposite of what one wants to have happen in a message like this.

The outcome of the check is a synopsis of what happened after the fact, not a integral explanation as to what constituted the check. Following this logic there is no need to state there was a check.

> If the amount of text that is less, doesn't have any content, then it is too little.

Here we agree. I take it you mean that to apply to the new version. Which suggests either you are not understanding some of it, or you think there is additional value in the current old version.

> wall of text

If you can put the actual content in not enough text to produce this effect, the issue isn't how it is broken up, or what the pieces or intended flow between them is.

>Automatic connection to the Internet established.
>> Ok. Why am I seeing this window this one time instead of the 500 times I've connected to the Internet?

Good point. It should be possible to get that sorted right away.

> Your Internet connection appears to be working.
>> appears introduces doubt. We have connected to elementary.io, there is no doubt that the Internet is currently working.

That string is actually in the original string, not the new one. I agree with my own point.

> but it was possible to do without anyhow.
>> Weird English.
It is quirky, not weird. To people whom English is a native language, this makes sense. I agree, as noted, that it fails for non-English users. Americans to some degree included in that.

`
You're connected
You can safely close this window and continue using your device.
`

> 82 characters and the user knows what to do next. If they still have questions that can continue reading.

82 characters and the user does not know:

1. What they are connected to. See earlier point about 500 times new thing.
2. What happened.

Having a second part that is for people that have more questions, when none of those are answered, is not really smooth-sailing. It is a crash-course waiting for novices. You can not expect people to stop reading because they can. This is a false assumption.

`
Why is this window appearing?`

This warrants reading.

There is a proposed question right there after you are safe to assume it can be closed. My first thought would be "How can i make this go away, maybe it will tell me". Amplified by the fact it hasn't carried any answers yet. Value - valuable. Not invaluable -valuable. (Which also makes sense, but isn't the point)

You came down two words, I will meet you half-way.

Will keep the 30 char limit in mind. How about:

```
Internet login skipped.
You can start browsing since elementary OS logged in for you.
In rare cases you may see it even if that was not needed.

[OK]
```
 3 lines  27 words 144 characters

"Internet login" has stopped meaning punching in PPPoE or similar, so lets just skip that mis-assumption. It is thus not a path to confusion for even novice users, since they don't know how to be confused about that. People who do, will also know what a captive portal is.

elementary OS self-references. OK, but taking credit where it is due, in a fashion that distinguishes it from any other party doing it, adds value in many ways.
Lets say your otherOS using friend also logs in on the same network concurrently, and doesn't see a message, if that becomes a topic of conversation, then i see a reason i see for having the last sentence.

PR #1809, opened 2017-09-03
adding underline to each li in nav-left on hover

PAPERPANKS:
### Changes Summary
- adding class `.nav-left` to `nav ul` in file `header.php`
- adding animated underline to `nav-left li a` on `:hover`

Reference taken from [here](http://tobiasahlin.com/blog/css-trick-animating-link-underlines/).


btkostner:
I'm going to give a no to this as it seems very out of place with the current style of the website, and doesn't provide any sort of feedback for smaller devices.

PAPERPANKS:
but there should be some visual change as user hover on these link.
can i play with opacity and color to achieve it

btkostner:
There was an opacity change to 0.7 on hover, and the color  should change on being clicked.

PR #1813, opened 2017-09-05
changing color of amount butttons

PAPERPANKS:
### Changes Summary

- color scheme of `amount button` ($5, $10, $25) change when clicked 

- on large screen

large screen

- on mobile devices

capture24


btkostner:
The white style of the inputs is a normal button style. The blue is a primary (`suggested-action`) button style. You shouldn't change them based on state.

PR #1843, opened 2017-10-27
Proposal for the main menu when viewed on mobile devices

jzunigarce:
Fixes #
I present a proposal for the design of the main menu the site to view the hamburger menu and avoid the horizontal scroll. It was tested in google chrome Version 61.0.3163.100 in mac.
### Changes Summary
- Change the menu display with horizontal scroll to hamburger menu
- 
- 

### DIFF
--- a/_scripts/pages/main.js
+++ b/_scripts/pages/main.js
@@ -6,6 +6,7 @@
 import analytics from '~/lib/analytics'
 
 import '~/external-links'
+import '~/menu-responsive'
 import '~/popover'
 import '~/smooth-scrolling'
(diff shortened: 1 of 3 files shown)


btkostner:
Thank you for taking the time to make a PR, but it looks like you are missing a JavaScript file, and the menu will unlikely change unless there is a very compelling reason to do so.

It was a @elementary/ux desision to do a horizontal scrolling section instead of a hamburger menu because of discoverability and ease of responsiveness. If there is a specific issue you want to address about the horizontal menu, please open up an issue so the UX team can talk about solutions.

PR #1880, opened 2018-01-20
Event scroll fix

### LINKED ISSUE #1331: Flipclock makes Webpage scroll to top
If you come to https://elementary.io/get-involved#desktop-development for example and hit the continue button  the url changes to https://elementary.io/get-involved#  and the webpage scrolls to top!

Firefox 48.0.1 on Mac OS 10.11.6

p-shannon:
Fixes #1331 

### Changes Summary

- Changes the continue to be a button instead of an anchor to increase semantic accuracy and dodge the smooth scroll event listener that listens to all clicked anchors with '#' in the href
- Apply a style to make the button look like a link.


btkostner:
I'll do an actual review of this when I get time (if nobody gets to it first), I just wanted to comment so nobody merges this yet.

p-shannon:
How would I go about disabling the event?

btkostner:
I'll just remove the added code. But i'll do it after testing and reviewing

p-shannon:
Thank you. Also sorry about messing up the code style again (as that is the reason the build failed), I'm still learning VIM so getting the tab spacing to play well with different values is a little painful. Will figure it out soon.

btkostner:
It's all good. It takes time to learn new editors

btkostner:
Sorry about the long review wait. Looking at the code, I wouldn't change the link to a button. Trying to reset all the styles is going to be a pain. The easier way would be to block the click event on the `a` tag. This is done with the [`Event.preventDefault();` method](https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault) and `return false` in the event handler.

p-shannon:
Thanks for the review, I thought that way would suffice but I thought a solution that also attempted to rectify semantic accuracy would be more suitable. I'll make the proposed corrections some time tonight.

btkostner:
I would normally agree that a button would be more semanticly accurate, but because of the default styles, it would take _a lot_ more work to use a button without making it a pain to maintain.

p-shannon:
Alright, did it the way you recommended! Also disabled the event and re-enabled autoplay. Please review at your own leisure. Thank you for your patience!

PR #1886, opened 2018-01-31
Full project name in debian/control example

peteruithoven:
Prevents people from getting common "Warnings about Debian control file" issue when submitting app. 


btkostner:
Thanks for the PR!

PR #1896, opened 2018-02-09
Add notifications to desktop integration

cassidyjames:
And a few other small cleanup things in that section.

### DIFF
--- a/docs/human-interface-guidelines.md
+++ b/docs/human-interface-guidelines.md
@@ -391,7 +391,7 @@ Users can enable Do Not Disturb mode from Notification Center or System Settings
 
 #### Notification Settings {#notification-settings}
 
-Notification bubbles, sounds, and appearance in Notification Center can each be toggled on or off on a per-app basis from the system notifications settings.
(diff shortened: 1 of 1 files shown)


docs/human-interface-guidelines.md
```diff
@@ -342,22 +344,54 @@ A badge shows a count of actionable items managed by your app. Its purpose is to
...
+1. [Farewell to the Notification Area](http://design.canonical.com/2010/04/notification-area/) by Matthew Paul Thomas
+2. [Status Icons and GNOME](https://blogs.gnome.org/aday/2017/08/31/status-icons-and-gnome/) by Allan Day
+
+## Notifications {#notifications}
+
+Notifications play a sound and are displayed as bubbles just below the system indicators. They briefly appear on screen where they can be selected to open the relevant app or manually dismissed by hitting the X icon. After a short time, they automatically slide away. Missed notifications can be seen in and cleared from the Notification Center indicator.
+
+### Sounds {#notification-sounds}
+
+Notifications play a system sound by default, but app developers are able to set an appropriate app-specific sound for users to be able to more quickly recognize the source of the notification.
```
danrabbit: It's probably worth mentioning here that developers should use the notifications API so that users can disable sounds if they choose to and not play sounds manually

cassidyjames: Ah yeah, that's what I was trying to imply but I'll be more specific.

PR #1930, opened 2018-03-23
Update code-of-conduct.json

nath5394:
Fixes mistakes in french

### Changes Summary

- correct a few mistakes in french


btkostner:
We do not accept translation fixes in this repo because they will get overwritten the next time someone translates something. Please use our [weblate instance](https://l10n.elementary.io/projects/website/code-of-conduct/fr/) instead.

nath5394:
Ok, i'll do like this 👍

PR #1948, opened 2018-05-04
Fix texts of alt element in support.php

ryonakano:
I think some texts of alt element in support.php may be wrong.

### Changes Summary

- Changed the alt element of AppCenter from "Calendar" to "AppCenter" in support.php
- Changed the alt element of Mail from "Geary" to "Mail" in support.php


ryonakano:
Thank you for your quick response!

PR #1956, opened 2018-05-13
Update section 5

cassidyjames:
Some updates regarding avoiding conflict.


danrabbit:
I'm not a big fan of the `Basically, “If you don't have anything nice to say, don't say anything at all.”` part. I don't think we want to encourage people to be quiet and sit on things they aren't happy about. We just want grievances to be expressed in a way that's more "Us vs the problem" than "You vs me"

cassidyjames:
@danrabbit good call, removed. I was trying to be cute and just referring to the sentence immediately before, but it didn't work. :P

PR #1958, opened 2018-05-13
Clarify labels of buttons/sections in "Get involved" page

waldyrious:
The choice of labels for some of the ways to get involved with the project were a bit ambiguous:

- "Support" is a very generic term, and can mean anything from financial contributions to helpdesk work. Changed to the more explicit "User Support".
- "Web" and "Desktop" are vague and unclear terms when not accompanied by the "development" suffix. Changed to "Website" and "Development", which IMO makes both meanings clearer in this context.

### Screenshots

Before:

screenshot-2018-5-13 get involved with elementary os 2

----

After:

screenshot-2018-5-13 get involved with elementary os

-----

Alternative (using one-word labels for all buttons):

screenshot-2018-5-13 get involved with elementary os 1

### DIFF
--- a/get-involved.php
+++ b/get-involved.php
@@ -29,9 +29,9 @@
         <div class="whole" id="sections-menu">
             <a class="button flat suggested-action" href="#funding">Funding</a>
             <a class="button flat suggested-action" href="#translations">Translations</a>
-            <a class="button flat suggested-action" href="#support">Support</a>
-            <a class="button flat suggested-action" href="#web-development">Web</a>
-            <a class="button flat suggested-action" href="#desktop-development">Desktop</a>
+            <a class="button flat suggested-action" href="#user-support">User support</a>
+            <a class="button flat suggested-action" href="#website">Website</a>
+            <a class="button flat suggested-action" href="#development">Development</a>
             <a class="button flat suggested-action" href="#design">Design</a>
         </div>
     </div>
@@ -101,7 +101,7 @@
     </div>
 </section>
 
-<section id="support" class="grey">
+<section id="user-support" class="grey">
     <div class="grid">
         <div class="two-thirds">
             <h2>Support</h2>
@@ -128,7 +128,7 @@
     </div>
 </section>
 
-<section id="web-development">
+<section id="website">
     <div class="web-browser">
         <div id="toolbar">
             <img src="images/get-involved/browser-left.svg">
@@ -144,7 +144,7 @@
     </div>
 </section>
 
-<section id="desktop-development" class="grey">
+<section id="development" class="grey">
     <div class="grid">
         <div class="two-thirds">
             <h2>Desktop Development</h2>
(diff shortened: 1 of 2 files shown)


get-involved.php
```diff
@@ -29,9 +29,9 @@
         <div class="whole" id="sections-menu">
             <a class="button flat suggested-action" href="#funding">Funding</a>
             <a class="button flat suggested-action" href="#translations">Translations</a>
-            <a class="button flat suggested-action" href="#support">Support</a>
-            <a class="button flat suggested-action" href="#web-development">Web</a>
-            <a class="button flat suggested-action" href="#desktop-development">Desktop</a>
+            <a class="button flat suggested-action" href="#user-support">User support</a>
+            <a class="button flat suggested-action" href="#website">Website</a>
+            <a class="button flat suggested-action" href="#development">Development</a>
```
danrabbit: I don't think changing "Desktop" to "Development" makes things more clear. Web development is also development.

get-involved.php
```diff
@@ -144,7 +144,7 @@
     </div>
 </section>
 
-<section id="desktop-development" class="grey">
```
danrabbit: I don't think we should change the IDs here. This is not user-facing and this change makes this ID more ambiguous than it was before

get-involved.php
```diff
@@ -29,9 +29,9 @@
         <div class="whole" id="sections-menu">
             <a class="button flat suggested-action" href="#funding">Funding</a>
             <a class="button flat suggested-action" href="#translations">Translations</a>
-            <a class="button flat suggested-action" href="#support">Support</a>
-            <a class="button flat suggested-action" href="#web-development">Web</a>
-            <a class="button flat suggested-action" href="#desktop-development">Desktop</a>
+            <a class="button flat suggested-action" href="#user-support">User support</a>
```
danrabbit: I'm not really sure I agree here because the nav link is also labeled "Support". If you go to Apple, Microsoft, Dell, Google, etc it's always just called "Support".

get-involved.php
waldyrious: Yes, I'm definitely don't mean to question that :) The ambiguity is in omitting "development" in both. Being explicit with "Web development" and "Desktop development" would resolve the ambiguity, but I fear it would be too verbose.

My perception is that an unqualified "development" is generally understood as "desktop development" by default, especially in the context of an operating system. I don't think that relying on that to provide a concise (and hopefully sufficiently clear) label to the button inherently conveys the notion that we don't consider web development to be development.

That said, we can always do something else, as long as it resolves the ambiguity of the current labels. Would something like "Coding" or "Programming" be better?

get-involved.php
waldyrious: Indeed. I just wanted to keep them in sync with the actual visible labels, as those small details can add brittleness over time. But yeah, it *is* a small detail, so no objection to revert this.

waldyrious:
@lewisgoddard was there a decision against the changes proposed here? The discussions above were still pending responses from @danrabbit.

PR #1968, opened 2018-06-01
Spelling: Versality treaty, other fixes

comradekingu:


docs/learning-the-basics.md
```diff
@@ -102,25 +102,25 @@ The panel, dock, and workspaces appear only on the primary display. Any other at
 
 ## Installing Apps {#installing-apps}
 
-elementary OS comes bundled with **AppCenter**, an app store for free apps. Installing a new app from AppCenter is easy:
+elementary OS comes bundled with **AppCenter**, an app store for apps free of charge. Installing a new app from AppCenter is easy:
```
cassidyjames: We should probably update this section to cover pay-what-you-want; not all apps are considered free.

docs/learning-the-basics.md
```diff
@@ -102,25 +102,25 @@ The panel, dock, and workspaces appear only on the primary display. Any other at
...
 
 1. Open the **AppCenter** app.
 2. Search in the top-right, or browse.
 3. Click **Install** next to the app you want to install.
 
 You may be asked for your password prior to installing an app.
 
-_Note: Some software may not be available from AppCenter. While we don't recommend downloading software from the general Internet, apps that are compatible with Ubuntu 16.04 LTS should work on elementary OS Loki._
+_Note: Some software may not be available from AppCenter. While downloading software from the general Internet is not recommended, apps compatible with Ubuntu 16.04 LTS should work on elementary OS Loki._
```
cassidyjames: I don't like the more passive voice here.

docs/learning-the-basics.md
```diff
@@ -6,75 +6,75 @@ As with every operating system there are some things that you may not be familia
...
 
 #### Indicators {#indicators}
-On the right side of the panel are icons called Indicators. These tell you of the current status of your session, i.e. your network connections, battery power, sound input and output, notifications, etc. Selecting an indicator exposes more information and related actions.
+On the right side of the panel are icons called indicators. These tell you of the current status of your session, i.e. your network connections, battery power, sound input and output, notifications, etc. Selecting an indicator exposes more information and related actions.
 
 A middle click or three finger tap on an indicator will present you with the following quick actions:
 
 * Sound: Toggle the mute state of the primary output device
-* Bluetooth: Toggle the Enabled/Disabled the primary bluetooth device
+* Bluetooth: Toggle on or off the primary Bluetooth device
```
cassidyjames: While we're here, I think it would be more concise and effectively as accurate to say, "Toggle on or off". Most devices are only going to have one Bluetooth radio, so specifying "the primary" feels overly precise

docs/learning-the-basics.md
danrabbit: This probably was meant to mean free as in freedom, not free as in beer. We could change it to Open Source

comradekingu: I would use the more precise "libre".

docs/learning-the-basics.md
comradekingu: It begs the question of who "we" are.

comradekingu:
@cassidyjames That is fine, but the use of big versal program names isn't consistent either.
Imo, they should be quoted, to avoid problems with the likes of

https://github.com/elementary/files/blob/481c5814a2bd69b81c048a6c4bf119590ada2902/src/View/directory_view_popup.ui#L170

"Folders Before Files"

PR #1969, opened 2018-06-12
Fix #1688: Change a alt tag text 'Text Editor' to 'Code'

### LINKED ISSUE #1688: Need to change the alt tags for the images on the website
We have unmatched app names in website with weblate (website/support)
* Geary
* Epiphany (Used in untranslated)
* Text Editor

They should be changed to 
* Mail (with using translated name, e.g. 메일)
* Epiphany (e.g. 에피퍼니)
* Scratch (e.g. 스크래치)

ryonakano:
Fixes #1688 

### Changes Summary

- Changed the alt element of Scratch from "Text Editor" to "Scratch" in support.php


tintou:
We rebranded Scratch to Code btw

ryonakano:
@tintou So you mean it is better to change to "Code"?

cassidyjames:
@ryonakano yes

ryonakano:
@cassidyjames I understand. Should I change in other pages too?

support.php
```diff
@@ -65,8 +65,8 @@
     </a>
 
     <a class="app" href="https://elementaryos.stackexchange.com/questions/tagged/scratch" target="_blank" rel="noopener">
-        <img width="64" height="64" src="images/icons/apps/64/accessories-text-editor.svg" alt="Text Editor"/>
-        <span>Scratch</span>
+        <img width="64" height="64" src="images/icons/apps/64/accessories-text-editor.svg" alt="Code"/>
```
ryonakano: And also, I couldn't find Code's icon. Do you know where it is?

cassidyjames: Hm, I believe it's shipped in the Code repo. I'm not sure what the best way to include it here is.

cassidyjames:
>Should we change in other pages too?
 Yeah, but that should probably be a different branch.

support.php
ryonakano: I see.
How about adding the icon in the Code repo to elementary Icons repo which links to website repo?

ryonakano:
> Yeah, but that should probably be a different branch.

I would like to work on this work, too. I created new PR #1974. I would appreciate it if you would give me some instructions and advice.

ryonakano:
I'm going to close this because I think this work is able to do in #1974.
Thank you for advising me a lot! And if possible, I would be grateful if you could give me some instructions and advice in #1974.

PR #2004, opened 2018-07-23
Increase download timeout (again)

danrabbit:
It seems like we're still having customers that are having trouble with this redirect race condition business. Increase the timeout even more to try to avoid this


btkostner:
At this point I think we should just omit it until we can get a better fix.

cassidyjames:
Yeah, moving the download logic to the thank-you page makes sense to me if we can do that as a permanent fix.

danrabbit:
@btkostner Yeah I kinda feel like maybe we should just kill the redirect for now

btkostner:
Same

PR #2029, opened 2018-08-17
Reference: Update GPL version in header example

### LINKED ISSUE #1888: Licenses in Getting Started doc
Following the getting started I ended up with 2 versions of the GPL license:

- [debian/copyright](https://github.com/elementary/debian-template/blob/master/debian/copyright): GPL-3.0+
- [src/* (GPL header)](https://elementary.io/docs/code/reference#gpl-header): GPL-2.0+ (I'm no license expert, maybe this is more a GPL generic license)
- [COPYING](http://www.gnu.org/licenses/gpl-3.0.txt): GPL-2.0+

Following conventions I saw in other apps I also added the license to:
- meson.build
- appdata.xml

Shouldn't we switch all licenses to GPL-3.0+? Looking through the apps on the AppCenter that's what most people use. If GPL-2.0 is preferred we should probably alter the [debian/copyright](https://github.com/elementary/debian-template/blob/master/debian/copyright) file.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/54544916-licenses-in-getting-started-doc?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github).
</bountysource-plugin>

ryonakano:
Fixes #1888

### Changes Summary

- Update the version of GPL header example that is in the reference page from 2 to 3


danrabbit:
GPL 2 and 3 are actually different licenses. I can't remember all the differences between them, but I think there was a reason that this is GPL 2 and not 3. Maybe someone like @Conan-Kudo can sum up the differences

Conan-Kudo:
GPLv3 adds additional protections:
* Requires grant to allow using modifications on target systems (anti-TiVoization)
* Explicit software patent grant
* Explicit compatibility with ASL 2.0 and AGPLv3.

However, GPLv3 is incompatible with GPLv2 because it imposes a restriction.

ryonakano:
@Conan-Kudo Thank you for the information.
Thus, legally it is possible to upgrade this header example because it says "either version 2 of the License, or (at your option) any later version" (= GPLv2+), isn't it?

Conan-Kudo:
Correct.

PR #2046, opened 2018-08-30
Fix markdown headlines in HIG for Github preview

AtosNicoS:
### Changes Summary
- On Github the markdown preview now renders Headlines correct.
- You should test if the doc rending for the website still gives the same output.


btkostner:
Thanks for the PR!

PR #2049, opened 2018-09-05
Spelling: StackExchange → Stack Exchange

comradekingu:
As per, https://stackexchange.com and https://en.wikipedia.org/wiki/Stack_Exchange


comradekingu:
Spotted by @Wompo  on https://weblate.elementary.io/translate/website/support/fi/?checksum=91c35a71d7561b48 :)

ryonakano:
Just a note, if we update "StackExchange" to "Stack Exchange", I think we should also update:

* The one in the reference list of https://github.com/elementary/website/blob/master/docs/human-interface-guidelines.md#preference-dialogs-preference-dialogs

* The one in the reference list of https://github.com/elementary/website/blob/master/docs/human-interface-guidelines.md#tooltips-text-buttons-tooltips

and this one:

https://github.com/elementary/website/blob/6a5522cd5f9d202a69e77e20e586ac0fdcfcad7e/support.php#L98

comradekingu:
@ryonakano Done :)

PR #1944, opened 2018-04-24
Homepage redesign & button shadows

jwinnie:
A miscellaneous collection of style tweaks.

Makes the AppCenter description on the homepage more like the AppCenter description on Houston; it adds highlights of purple and more bold branding.
appcenter

I am an amateur web developer, so please point it out if I made any mistakes.

This pull request IS ready for review.

PR #2056, opened 2018-09-26
Homepage design pass

cassidyjames:
Based on #1944 with a bit of a reduced diff.

### Changes Summary

- Purple style for AppCenter section
- New AppCenter copy
- Sections moved around a bit
- Tweaked lead copy


danrabbit:
Couple minor comments:

I wonder if the links in the appcenter section should become buttons instead of changing the `a` color to contrast better against the background

We'll need to update the back icon SVG (or switch that to a FontAwesome icon) in the apps section

cassidyjames:
@danrabbit both addressed!

PR #2060, opened 2018-10-01
Update get involved page to palette colors

### LINKED ISSUE #1654: Use elementary color palette for Get Involved page
Buttons and such still use the solarized color palette, we should use our own.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/44174741-use-elementary-color-palette-for-get-involved-page?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github).
</bountysource-plugin>

micahilbery:
Fixes #1654

### Changes Summary

- Update Get Involved page's buttons to use palette colors
- Add focus state for top buttons
- Updated SVGs to also use palette colors


micahilbery:
I tried to at least keep the same contrast if not better.

cassidyjames:
Does what it says. I wonder if we should try this approach while we're changing the colors and looking at contrast: https://twitter.com/steveschoger/status/1045371124274483200?s=21

micahilbery:
Like the 100 color weight with dark text on top? That could be cool. For the focus states would we want to go up a weight to like the 300? I feel like that could work for everything except maybe the slate color at the end but I guess we could easily go to like silver for it.

cassidyjames:
Screenshot from @micahilbery 

get-involved-light

PR #2073, opened 2018-10-11
OS Development tips

peteruithoven:
Fixes #
Lack of general tips for development on elementary OS itself.

### Changes Summary
-  Adds a very crude os-dev page with tips.

Is this something we want? Is this the place to do it?


jwinnie:
Addition from @donadigo:
"Yes, I actually wanted to write some kind of guide for people that do that
Your escape from this is one of the TTY sessions Linux provides
So if your gala crashes you can get to another terminal by hitting e.g: Ctrl+Alt+F1
You login there and you're in
Then if gala didn't crash but hang, you just kill it with killall gala
Now you need to start gala inside the graphical session
Unfortunately DISPLAY=:0 gala doesn't work
So you spawn a terminal app in the graphical session and it will be automatically focused because there is no WM: DISPLAY=:0 io.elementary.terminal
Then you switch to the graphical session by hitting Ctrl+Alt+F7
And launch gala in the newly opened terminal window: (gala -r &) &
Gala should start and you can close this terminal
If your gala can't start, you have to either reinstall it (sudo apt install --reinstall gala) or fix the error in your code with an editor like vi, both in the TTY session"

jeremypw:
Suggest adding a section on the basic use of gdb to obtain a backtrace and maybe touch on `set env G_DEBUG='fatal-criticals'`

Maybe also mention io.elementary.vala-lint

peteruithoven:
Thanks @jwinnie, I've included the tip from @donadigo on restarting Gala. 
@jeremypw While I agree `gdb` or `G_DEBUG='fatal-criticals'` are super interesting I'm not knowledgeable enough to write those sections. PR's are more than welcome. 

I'm going to try adding something on the vala linter.

btkostner:
I also want someone from @elementary/desktop-developers to review this before we merge as well.

docs/code/os-dev.md
```diff
@@ -0,0 +1,56 @@
+# elementary OS Development
+Want to contribute code to elementary OS itself? Here are some tips.
+
+## Vala linting
+To make it easier to follow the [elementary Code-Style guidelines](https://elementary.io/docs/code/reference#code-style) you can use [Vala-lint](https://github.com/elementary/vala-lint).
+
+## GtkInspector
+Use the [GtkInspector](https://wiki.gnome.org/Projects/GTK+/Inspector) to make development easier.
```
cassidyjames: This should probably cover how to get it (install 'libgtk-3-dev`, which is included in `elementary-sdk` and then set `org.gtk.Settings.Debug enable-inspector-keybinding` to `true`) instead of just linking to a GNOME Wiki.

docs/code/os-dev.md
```diff
@@ -0,0 +1,56 @@
...
+1. Start wingpanel by running: `wingpanel`
+
+To restore normal behavior simply enable "Use default value" again. *Cerbere* will notice this and start to monitor *Wingpanel* again.
+
+## Gala
+*Gala* is the window manager of elementary OS. When it crashes or freezes it can be difficult to get going again. Here's how to do it:
+1. Go to one of the virtual consoles by pressing: <kbd>Ctrl</kbd> + <kbd>Alt</kbd> + <kbd>F1</kbd>.
+2. Login.
+3. If *Gala* didn't crash but froze, you can kill it by running `killall gala`.
+4. You unfortunatly can't start *Gala* from here, so we start a *Terminal* by running: `DISPLAY=:0 io.elementary.terminal`.
```
cassidyjames: I'm fairly sure running `DISPLAY=:0 gala --replace` works fine in TTY? I've certainly done it before.

peteruithoven:
@btkostner this "mdr markdown error", I don't see this when I run it locally. Also not when using `npm run lint`. How can I reproduce it?

docs/code/os-dev.md
peteruithoven: I didn't want to duplicate their documentation, but sure, I'll add it.

docs/code/os-dev.md
santileortiz: Using `gala --replace -d :0` stopped working some time ago. Mutter now shows the error `Window manager warning: Unsupported session type`.

I just tried, and it can be solved by using `XDG_SESSION_TYPE=x11 gala --replace -d :0`

btkostner:
@peteruithoven It's a PHP offset error. It doesn't prevent the page from loading, but it shows up in the console, so I want to fix it.

jeremypw:
I am by no means and expert on gdb so I was envisaging only the simplest commands to produce a backtrace (that is all I use) and a link to a good tutorial on the web.

docs/code/os-dev.md
peteruithoven: Interesting! I just tried that after killing gala manually and that does seem to work. It's also way easier. I'll update the docs.

peteruithoven:
I'd love for someone who has actually used `gdb` to do a separate PR. I don't think it's a reason to hold back this PR. 
I've implemented the other suggestions.

I'll wait for more information on the "mdr markdown error".

docs/code/os-dev.md
cassidyjames: @peteruithoven their documentation can change and doesn't necessarily imply elementary OS, whereas we know exactly the steps required for our platform (like installing the SDK).

peteruithoven: Good point

jeremypw:
@peteruithoven : That's fine. There is no need to hold this PR up.

peteruithoven:
@RMcNeely thanks for the review. @cassidyjames thanks for the improvements.

docs/code/os-dev.md
```diff
@@ -30,7 +30,7 @@ First, make sure you have the `elementary-sdk` installed. Then enable the Inspec
 gsettings set org.gtk.Settings.Debug enable-inspector-keybinding true
 ```
 
-Focus your app, then launch the Inspector by pressing <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>I</kbd> to inspect the widget beaneath your cursoe, or <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>D</kbd> to open the inspector without a widget selected.
+Focus your app, then launch the Inspector by pressing <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>I</kbd> to inspect the widget beneath your cursor, or <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>D</kbd> to open the inspector without a widget selected.
```
cassidyjames: lol I don't know how that happened, thanks. :sweat_smile:

peteruithoven:
In a future PR I'd also like to include D-Feed, for debugging / experimenting with Dbus interfaces.

cassidyjames:
@btkostner is that build error still happening here? Otherwise this looks good to me.

PR #2085, opened 2018-10-18
Fix IP override

lewisgoddard:
I've figured out why we were temporarily sending all traffic to `ams3`. The old `$ip = '127.0.0.1'` was always `true` because it was setting the variable, not a comparison `==` like it should be. Additionally, leaving `!$ip` would mean all undetectable IPs would be sent to the same region.

### Changes Summary

- Fix comparison. `==` not `=`
- Allow appending `?ip_override` to any URL to override IP detection for local network testing.
- Don't override undetectable, as it would produce false shipping estimates and lock region instead of balancing.


btkostner:
This fixes some issues, but all that code was commented out so I assume it doesn't fix the issue in production right now

lewisgoddard:
Based on the analytics, production right now is balancing properly. Odd fringe things, especially in public networks like what cassidy described, were always a thing. This is definitely what was causing issues up to three hours ago https://github.com/elementary/website/commit/7b38f8d23692116d7f40dd985c0ec16cbe24ca4b

lewisgoddard:
https://elementary.io/api/config

```
{
    "release": {
        "title": "Juno",
        "version": "5.0"
    },
    "keys": {
        "sentry": "https:\/\/5dc4884b382c42b3a0d6440797af3454@sentry.io\/101325",
        "stripe": "pk_live_ZtcG6SqsVtMbTuJARKujZ0lE"
    },
    "user": {
        "ip": "92.26.49.66",
        "region": "ams3",
        "timecode": "MTUzOTg4MTc2MQ==",
        "trackme": true
    }
}
```

---

https://elementary.io/api/config?ip_override

```
{
    "release": {
        "title": "Juno",
        "version": "5.0"
    },
    "keys": {
        "sentry": "https:\/\/5dc4884b382c42b3a0d6440797af3454@sentry.io\/101325",
        "stripe": "pk_live_ZtcG6SqsVtMbTuJARKujZ0lE"
    },
    "user": {
        "ip": "78.148.208.61",
        "region": "ams3",
        "timecode": "MTUzOTg4MTc2Ng==",
        "trackme": true
    }
}

PR #2088, opened 2018-10-19
Use correct item in Dynamic Quicklists example

oskarkunik:
'item1' should be 'item2' in two lines


btkostner:
Thanks for the PR!

PR #2095, opened 2018-10-21
Improve gulp/npm build process

jwinnie:
### Changes Summary

- Add `package-lock.json` (*autogenerated by NPM*)
- `gulp` now runs Webpack and `gulp watch` now watches for JavaScript files as well
- `gulp watch` now runs the PHP server
- Fixed NPM error with gulp versioning that was causing builds to fail


gulpfile.babel.js
```diff
@@ -208,4 +233,9 @@ gulp.task('watch', gulp.series('default', function watch () {
     gulp.watch('_images/**/*.svg', gulp.series('svg'))
 
     gulp.watch('_styles/**/*.css', gulp.series('styles'))
+    gulp.watch('_scripts/**/*.js', gulp.series('scripts'))
```
btkostner: In a perfect world this would run webpack in watch mode so it's faster, but our script builds do not take that long so I'm ok with it.

btkostner:
Thanks for updating this. It's been a while since anyone has looked through that.

PR #2102, opened 2018-10-28
Add alt tag text for images on the homepage

samu101108:
### Changes Summary
- Added some alternative text missing on images and icons


index.php
```diff
@@ -421,7 +421,7 @@
             </div>
         </section>
         <section class="cta">
-            <img src="images/icons/places/128/distributor-logo.svg">
+            <img src="images/icons/places/128/distributor-logo.svg" alt="Elementary OS logo">
```
danrabbit: "elementary" should be written lowercase:
```suggestion
            <img src="images/icons/places/128/distributor-logo.svg" alt="elementary OS logo">
```

index.php
```diff
@@ -307,7 +307,7 @@
                         <?php include('images/pantheon/actions/system-search.svg'); ?>
                         Applications
                     </div>
-                    <div id="slingshot-arrow"><img src='images/slingshot/arrow.svg'></div>
+                    <div id="slingshot-arrow"><img src='images/slingshot/arrow.svg' alt="An arrow that points to Applications name"></div>
```
danrabbit: This doesn't really make sense to add an alt tag for since it's a part of a larger graphic, not a independent graphic in itself. It's essentially a border image

index.php
```diff
@@ -92,7 +92,7 @@
             <div class="app-display app-display--overflow">
                 <img class="app-display__image" src="images/screenshots/appcenter.png" srcset="images/screenshots/appcenter@2x.png 2x" alt="elementary OS AppCenter home page"/>
                 <div class="app-display__description">
-                    <img src="images/icons/apps/128/system-software-install.svg" />
+                    <img src="images/icons/apps/128/system-software-install.svg" alt="elementary OS AppCenter icon"/>
```
danrabbit: When disambiguating application names, the convention is "elementary Foo" rather than "elementary OS Foo" 
```suggestion
                    <img src="images/icons/apps/128/system-software-install.svg" alt="elementary AppCenter icon"/>
```

index.php
samu101108: Yes you're right! Decorative images shouldn't have any descripition, but it still needs and alt attribute, even if it is blank, so the code should be:
`<div id="slingshot-arrow"></div>`

Here are some info about using alt in decorative images [](https://webaim.org/techniques/alttext/#decorative)

index.php
```diff
@@ -307,7 +307,7 @@
                         <?php include('images/pantheon/actions/system-search.svg'); ?>
                         Applications
                     </div>
-                    <div id="slingshot-arrow"><img src='images/slingshot/arrow.svg'></div>
+                    <div id="slingshot-arrow"><img src='images/slingshot/arrow.svg' alt=""></div>
```
btkostner: Remove the alt completely if you are not setting text (which you shouldn't here)

lewisgoddard: This is not correct, many screen readers will read the URL if no alt tag is set. Decorative images should have empty alt tags.

PR #2144, opened 2019-01-12
Added marathi option in the list in L10.php file 

josprachi:
Fixes #

### Changes Summary
added 'mr => मराठी' into L10.php
- 
- 
- 


_backend/Lib/L10n.php
```diff
@@ -116,6 +116,7 @@ public static function languageDirectory($lang)
         'ko' => '한국어',
         'lt' => 'Lietuvių kalba',
         'ms' => 'bahasa Melayu',
+        'mr' => 'मराठी'
```
ryonakano: I suppose you'll need to add `,` at the end of this line like other lines.

NathanBnm: Yes, because otherwise it would break the code.

PR #2150, opened 2019-01-20
Update link to l10n Slack invite

axilleas:
The link on https://elementary.io/docs/translation-guide was
different (also expired) than the one on https://elementary.io/get-involved.


btkostner:
Thank you for the PR!

PR #2152, opened 2019-01-22
Spelling: FastMail, Cloudflare

comradekingu:
Related note: Is "behavioural elements" precise? Should it not be the "behaviour of elements"?
Also, "disable" could be "turn off", and information could be info. All in the interest of brevity and easy of understanding.

### DIFF
--- a/privacy.php
+++ b/privacy.php
@@ -16,7 +16,7 @@
                 <h2>elementary OS</h2>
                 <p>We do not collect any data from elementary OS. Your files, settings, and all other personal data remain on-device unless you explicitly share them with a third-party app or service.</p>
                 <h3>Online Accounts</h3>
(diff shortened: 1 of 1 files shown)


btkostner:
behavioural elements = behaviour of elements
disable is more of a computer term

Honestly all of those changes could be made, but it would still mean the same thing. For sake of time and change, I think we should just let those two changes be as they are now.

comradekingu:
@btkostner I left them out since these were clear.
I would argue that if what could be considered a computer term can be replaced equally with something that isn't, inroads on friendliness will have been made.
The reason in my head to use "disable", is if it is possible to enable it, but then not have it turned on. Meaning therefore that the opposite status "enabled", does not necessitate action, but rather capability.

PR #2156, opened 2019-01-27
Update translations to support title attribute on inputs + Update layout

### LINKED ISSUE #1685: Let's translate "Use PayPal"
For some reason, "Use PayPal" string on get-involved page is not available for translation.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/44712675-let-s-translate-use-paypal?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #2154: Missing translations
Some translations are still missing although they have been recently updated:

- In the footer "Press" and "Privacy" are probably not translatable so we have to add them to be translatable, here is a screenshot (French) (See #2162)

image

- The "Store" page has been translated to French but still appearing in English for the moment. **Why is the store in the exclusions if it can be translated?**

NathanBnm:
#### Update translations to support title attribute on inputs

I added *"PayPal - The safer, easier way to pay online!"* which is the displayed title when hover on the *"Use PayPal"* button the possibility to be translated. Fixes #1685.

- Update `L10n.php`,
- Update `get-involved.json` file in all languages to include the new string,
- Add French translation to `fr/get-involved.json`

#### Update layout

This should partially fix #2154.

- Update `layout.json` to update footer translations including last strings like *"OEMs"*,
- Delete default English translations,
- Update *"Copyright © 2017 elementary LLC."* to *"Copyright © 2019 elementary Inc."*

Sorry about both of those things at once. I'll be more careful next time.


NathanBnm:
Ready for review (I just made some fixes)

btkostner:
@NathanBnm 	did you make any manual changes besides adding `title`? I assume you used the translation script for updating all the translation files

NathanBnm:
> @NathanBnm did you make any manual changes besides adding `title`? I assume you used the translation script for updating all the translation files

@btkostner just look at my Changelog above, there is all the details. I updated `L10n.php` to make the title attribute translatable. Then I updated all the `get-involved.json` files by adding the corresponding string. Finally I also updated the `layout.json` files which were out of date.

Uh I made the changes manually 'cause I don't know how to use the script, and there is no doc about that. And concerning `layout.json` files I had to be careful of the existing translations to not erase them.

Today there was just one conflict which I resolved. I think you can merge this without any issue.

btkostner:
@NathanBnm 	In that case, I'm going to make this two different comments in master. It's easier to manage. Also, the translation scripts are _very_ picky about encoding and what not, so after I'm done with that change I'm just going to run the script to extract translation strings. Sorry those scripts are not documented. They really should be.

btkostner:
Fixed in:

https://github.com/elementary/website/commit/ceb965ff36b3c9560b497f5a65aa0ca1d69eaf8f
https://github.com/elementary/website/commit/b1ffc8aa002458ddd438963e2da2f1603805cf86

NathanBnm:
@btkostner thanks for merging this 👍 And yes this very should be documented!

btkostner:
@NathanBnm 	I noticed that the translation script does not extract any of the template files and they cause breakage when trying to include them on their own. I'll open a separate bug about it.

NathanBnm:
@btkostner there is no way to update this manually ? Because with the script some existing translations may be deleted. I had to be carefum about that when I made the changes.

btkostner:
The script uses the same `key` generation that the website uses when translating. So if the JSON `key` changes, it's because it needs to, otherwise that translation will never be used. If I remember correctly, the `value` will only change if the `key` changes or if the `value` is the same as the `key` (untranslated).

We don't accept translations in the form of PRs due to weblate git issues that sometimes occur if we do. We also try not to update translation files by hand because it can be error prone and the way we generate and escape can be weird.

If there is an error with the way we extract the strings and update the templates with the `php` script, I would consider it a bug and would want to fix that instead of trying to manually edit the files.

NathanBnm:
Alright!

PR #2157, opened 2019-01-29
Add Workflow section to homepage

cassidyjames:
Based on feedback from the Forbes challenge and users. This adds a new Workflow section to the homepage to cover handy workflow and productivity features that users did not know about before using elementary OS. The idea is that once they learned of these features, the OS became more compelling to them. We should use that as a selling point to attract potential users.

image

Exact location of the section, copy, and illustrations are all open to improvement. I'm generally happy with the direction of the content, though.


danrabbit:
I feel like the screenshots are kind of same-y and I wonder if we should use icons or illustrations here instead

cassidyjames:
@danrabbit I agree they're same-y. I think it could be _really_ slick if they're animated, and maybe we switch out the wallpaper. One of the intents was not just to talk about these features, but show them. A sort of symbolic illustration could also achieve that. But I don't think I just want a little symbolic icon like the Privacy section if we can do something that shows a bit more.

index.php
```diff
@@ -127,6 +127,50 @@
                 <a class="read-more" href="https://usn.ubuntu.com/releases/ubuntu-18.04-lts/">Security Notices</a>
             </div>
         </section>
+        <section id="workflow" class="grey">
+            <div class="grid">
+                <div class="two-thirds">
+                    <h2>Get work done. Or play.</h2>
+                    <h4>elementary OS is designed to keep you productive and focused by getting out of your way with Multitasking View, Picture-in-Picture, Do Not Disturb, and more. And by staying out of your way, elementary OS is great when you're watching videos or playing games, too.</h4>
```
danrabbit: I feel like mentioning "elementary OS" each time is a bit repetitive since this whole page is in the context of elementary OS. Maybe we could do something like:

```suggestion
                    <h4>Stay productive and focused with Multitasking View, Picture-in-Picture, Do Not Disturb, and more. Or keep work out of sight when watching videos or playing games.</h4>
```

Or in general just something more concise. Otherwise each section is saying "elementary OS is designed to" at the beginning of each paragraph

index.php
```diff
@@ -127,6 +127,50 @@
...
+                </div>
+            </div>
+            <div class="grid">
+                <div class="third">
+                    <figure class="multitasking">
+                        <div class="workspace"></div>
+                    </figure>
+                    <h3>Multitasking View</h3>
+                    <p>Powerful workspaces help you organize your work by task. Keep work and play separate, but just one tap away.</p>
```
danrabbit: Using the word "Powerful" is a pet peeve. What makes them "powerful"?

```suggestion
                    <p>Workspaces help organize your work by task. Keep work and play separate, but just one tap away.</p>
```

index.php
```diff
@@ -127,6 +127,50 @@
...
+                </div>
+                <div class="third">
+                    <figure class="pip">
+                        <div class="workspace">
+                            <img class="window" src="images/screenshots/videos.png" srcset="images/screenshots/videos@2x.png 2x" alt="Videos screenshot" />
+                        </div>
+                    </figure>
+                    <h3>Picture-in-Picture</h3>
+                    <p>Keep tabs on one thing while working on another. Whether you’re watching a movie, game, or terminal process, Picture-in-Picture keeps it within view.</p>
```
danrabbit:
```suggestion
                    <p>Whether you’re watching a movie, game, or terminal process, Picture-in-Picture helps keep tabs on one thing while working on another. </p>
```

index.php
```diff
@@ -127,6 +127,50 @@
                 <a class="read-more" href="https://usn.ubuntu.com/releases/ubuntu-18.04-lts/">Security Notices</a>
             </div>
         </section>
+        <section id="workflow" class="grey">
+            <div class="grid">
+                <div class="two-thirds">
+                    <h2>Get work done. Or play.</h2>
```
danrabbit: Other sections are title case :)

```suggestion
                    <h2>Get Work Done. Or Play.</h2>
```

Issue #1518, opened 2017-01-01
Add/refresh Press Quotes


cassidyjames:
Our press quotes are pretty old... we should see if we can get some fresher quotes to add/replace some that are there.

cassidyjames:
Not sure if these are big enough names to unseat any of the current quotes, or should be added:

# [TechRepublic](http://www.techrepublic.com/article/open-source-predictions-for-2017/)
 > very user-friendly, beautifully designed, and makes working with Linux a complete no-brainer on every level

# [Network World](http://www.networkworld.com/article/3130760/open-source-tools/elementary-os-04-review-and-interview-with-the-founder.html)
> astoundingly approachable for, well, anyone.

cassidyjames:
# [Linux.com](https://www.linux.com/learn/elementary-os-loki-has-arrived)
> one of the most elegant and well-designed Linux desktops on the planet.

nickheppleston:
Happy to work on this one, however not too sure where you obtain the SVG images from for the corporate logos?

lewisgoddard:
I think we had to make some of those, if not look for press kits for those sites and desaturate.

cassidyjames:
# [MakeUseOf](http://www.makeuseof.com/tag/time-try-something-new-elementary-os-loki/)

Not sure what of this to trim down to, but there's some good snippets:

> The elementary OS team sweats the details. 

>  the software all fits together. Learn how to use one program, and you know how to use the next.

> elementary OS takes seconds to understand... quick for newcomers to grasp. But this is also what I consider the biggest draw for experienced Linux users too. 

> The lack of distractions leaves more time to focus on whatever I’m doing, whether that’s browsing the web or writing a novel. elementary OS’s simple design helps me get more done.

vikaskyadav:
Check out these links too. #2 and #3 are the latest ones.
1) https://fosspost.org/2016/09/11/complete-review-on-elementary-os-0-4-loki/
2) http://www.linuxandubuntu.com/home/elementary-os-04-loki-review-an-easy-to-use-linux-distribution
3)http://www.everydaylinuxuser.com/2017/01/an-everyday-linux-user-review-of.html

btkostner:
Dropping some ideas here:

- On page, load pick 3 random quotes from a list
- Possibly use language or geo data to load better quotes. Maybe even include quotes in other languages.
- Inline svg images for faster load times (needs some more investigation)

PR #2163, opened 2019-02-21
Add Ars and Forbes to press section

cassidyjames:
Adding a couple more press quotes to the site. Not sure about the Forbes one (it's so much longer and required bracketed text for context), and not sure about that Ars logo (it looks optically so much smaller).

image

While I'm here, I fixed the tweet length allowance in the Twitter js since it was messing up the Forbes quote. 

Would probably be considered a fix for #1518.


cassidyjames:
Can be previewed at https://beta.elementary.io/press/

Issue #696, opened 2015-06-12
Link bug reporting guide from "get involved" page


nicekiwi:
The bug report guide is current hidden in Developer > Code. Shouldn't it also be next to the link to open bug reports in Desktop development on the Get Involed page?

danrabbit:
Hm I'm actually not sure about this. The button there is more about linking to existing reports as a list of work items. I've begun making sure that we have a link to the guide on various launchpad pages where it says "Bug reporting guidelines".

lewisgoddard:
Maybe a new link is the way to go for this. "How to Report Bugs" > "Report a Bug"

lewisgoddard:
I think `/support` should include a link to "How to Report a Bug"

danrabbit:
I'm very very -1 on putting bug reporting in "Support". Reporting issues is not a form of support. There is no guarantee an issue will be addressed in a timely manner.

lewisgoddard:
Support is still going to be the place people go when they experience a bug though, and we can help them report bugs better.

PR #2166, opened 2019-02-23
Extending the Getting involved > Desktop development section

peteruithoven:
Extends the Getting involved > Desktop development section.

Added a area on reporting issues (fixing #696). I think this is an okay spot because it is a very low threshold way of contributing to desktop development. 

Also added a section on fixing issues / improving features / adding features to the OS. This would be, as far as I know, the first place we'd be linking to the new [elementary OS Development page](https://elementary.io/docs/code/os-dev).

I think the third way that people can improve our desktop ecosystem is by creating new apps, might be a bit of stretch, but I do think this aspect deserves attention on our Getting involved page.

I'm not a great writing so I' sure you'll have feedback.


RMcNeely:
Since each of your headings is only one sentence maybe it would make more sense to combine them into two. You could have "Find and Fixing Issues" and "Create Apps and Features". I think if you were to split up Fixing issues you may be able to write more for each block. 

Also I think you probably want to line the buttons up on below the blocks. Right now "Publish on Appcenter" button looks to be starting about 50% of the way up on the other two.

peteruithoven:
I've separated finding and fixing issues for a couple of reasons:
- They have separate read more links
- They talk to different people where finding and reporting issues is way more approachable. (It's how I started contributing)
- I'd also love to extend the Report issues part to invite people to help organize issues, the "issue gardening" aspect.

I've created 3 parts party because of the available resources. 
1. For reporting issues we have [Reporting bugs](https://elementary.io/docs/code/reference#reporting-bugs).
2. For fixing issues and adding features to elementary itself we have [elementary OS Development](https://elementary.io/docs/code/os-dev) which focuses on working with existing aspects of elementary OS, like how to work on the Wingpanel. In a lot of cases you don't have to know a lot about things like Vala or Meson at all to contribute.
3. For creating apps from scratch there is the [getting started](https://elementary.io/docs/code/getting-started). 

> I think if you were to split up Fixing issues you may be able to write more for each block.

Why though? Making things more specific / focused usually means less text?

> Also I think you probably want to line the buttons up on below the blocks. Right now "Publish on Appcenter" button looks to be starting about 50% of the way up on the other two.

I'd prefer doing that in a separate, more style focused PR.

get-involved.php
```diff
@@ -195,10 +195,27 @@
         <div class="whole">
             <div class="actions">
                 <a class="button flat suggested-action" href="https://github.com/elementary" target="_blank" rel="noopener">Browse Our Code</a>
+            </div>
+        </div>
+        <div class="third">
+            <h3>Report issues</h3>
+            <p>Help out by tracking down issues and reporting those or by clarifying or cleaning up existing issues. <a class="read-more" href="/docs/code/reference#reporting-bugs" target="_blank" rel="noopener">Learn More</a></p>
```
jeremypw: Suggest improved clarity/grammar:

`Help out by tracking down issues and reporting them, or by clarifying and cleaning up existing issues.`

peteruithoven: Thanks

PR #2176, opened 2019-03-16
Do not allow access to router script in production.

vaibhavpandeyvpz:
Upon visiting `https://elementary.io/router`, currently (20190316) it shows:

This page isn’t working
**elementary.io** is currently unable to handle this request.
HTTP ERROR 500

If the script is only meant for use while development, then better to forbid access to it on production servers.


router.php
```diff
@@ -6,6 +6,12 @@
  * your local PHP development server. NOT intended for production whatsoever.
  */
 
+if (php_sapi_name() !== 'cli-server') {
```
lewisgoddard: We use `if ( getenv('PHPENV') !== 'production' ) {` in `_templates/alert.php`, maybe that would be better here?

vaibhavpandeyvpz: I do not prefer directly dealing with `getenv` as changing those values require `sudo` access to the server. Since `php` 5.4, `cli-server` is present as an `sapi` and should be preferred over custom env vars. Though we must not mix 2 ways to accomplish same thing in 1 project, I hereby amended that to use (opposite of) proposed condition.

btkostner:
Thank you for the PR!

PR #2178, opened 2019-03-26
Use more illustrative style

danrabbit:
Thoughts?


cassidyjames:
I get the idea, but I feel like if you don't already know what you're looking at, it's just some weird gray rectangles animating around.

Issue #2149, opened 2019-01-19
Add gdb docs


peteruithoven:
Add some introduction to using `gdb` to:
https://elementary.io/docs/code/os-dev

Also see: https://github.com/elementary/website/pull/2073#issuecomment-435891909

PR #2198, opened 2019-05-22
Inspecting crashes intro

peteruithoven:
Fixes: #2149

### Changes Summary
- A ultra short intro into using `gdb` to inspect crashes


doublehourglass:
Good general description. Would add that it's a general purpose external debugger and more options can be found in manpages. As it's not only a crash inspector.

PR #2213, opened 2019-06-07
Disable sentry on non-production deployments.

lewisgoddard:

### DIFF
--- a/_backend/log-echo.php
+++ b/_backend/log-echo.php
@@ -3,7 +3,11 @@
 require_once __DIR__ . '/bootstrap.php';
 
 // Setup sentry error logging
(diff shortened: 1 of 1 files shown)


_backend/log-echo.php
```diff
@@ -3,15 +3,19 @@
 require_once __DIR__ . '/bootstrap.php';
 
 // Setup sentry error logging
-if (isset($config['sentry_key']) && $config['sentry_key'] !== false) {
+if (
+    getenv('PHPENV') == 'production' &&
```
btkostner: I'm like 99% sure that all the `beta.elementary.io` sites are also in production to avoid weird errors while testing.

PR #2217, opened 2019-06-09
Update Font Awesome Pro to version 5.9.0

### LINKED ISSUE #2151: Update Slack Icon
Slack just reviewed their interface and icon as explained in this article : https://slackhq.com/say-hello-new-logo
Maybe the old icon should be replaced by the new one ?

lewisgoddard:
Fixes #2151

### DIFF
--- a/_templates/header.php
+++ b/_templates/header.php
@@ -76,7 +76,7 @@ $l10n->begin_html_translation();
         <link rel="stylesheet" type="text/css" media="all" href="https://fonts.googleapis.com/css?family=Raleway:800|Open+Sans:300,400,600,700,300italic,400italic|Roboto+Mono">
         <?php } ?>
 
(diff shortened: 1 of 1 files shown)


lewisgoddard:
openssl dgst -sha384 -binary all.css | openssl base64 -A
    PzE/Pz8/I3c/P3ZiPws/P0gGKj8/P38/Pz9PPy56Oj8/Pyo7Pz8/JWRWP3NhPz8vDQo=

NathanBnm:
Is the integrity key correct? Maybe it's still the same issue.

NathanBnm:
@btkostner @lewisgoddard apparently FontAwesome reviewed their way to import fonts. Now you have to include an script with their new icon kit system. Just check on their website for further info. This will permit to manage your icons directly from your account and always have the last version.

I've just used it for a personal project and it's working fine. You just have to add a script like that:

`<script src="https://kit.fontawesome.com/<number>.js"></script>`

Then you'll have your FontAwesome kit loaded.

lewisgoddard:
Okay, we 100% need access to the account with the Pro subscription. They now use a link that is unique and carries settings within it.

btkostner:
Looks like the migration path to the new JS kit way would require us to make a list of all the icons we use on the site and probably change the way we use them. I think we should do this as a stop gap first.

PR #2229, opened 2019-07-14
Align display on hero section

andreasfelix:
Maybe it is just me, but for me it looks like the display of the laptop in the hero section is not perfectly aligned.

### Changes Summary

- align display of laptop
- add subtle border-radius similar to the real OS

**Before**
elementarty_hero_before

**After**
elementarty_hero_after


cassidyjames:
Thanks, looks great!

PR #2230, opened 2019-07-15
Payment trust

### LINKED ISSUE #2124: Improve payment flow trust
Something I hear from users is that it's not clear they're using Stripe for the payment, or they're not sure what card types are accepted, or they're not comfortable entering their payment info on the site. I could see how this wouldn't be the most clear that Stripe is processing it and that they're not just handing over payment data directly to elementary:

screenshot from 2018-12-04 12 30 06

Something we could do to combat those fears and help them realize they can trust the site is to surface the payment processor earlier, i.e. include a Stripe logo and possibly common card type icons below the payment options.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/66997320-improve-payment-flow-trust?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github).
</bountysource-plugin>

cassidyjames:
Fixes #2124 by adding accepted card logos and copy about Stripe below the payment options. Open to input.

image


cassidyjames:
Actually I think I like our color payment icons better, so we should use these

image

cassidyjames:
Using our color icons:

image

lorenzocafaro:
Looks good to me! I'd maybe only show the most used cards and keep the Stripe link grayscale since it's a "secondary" link a bit like the social icons on top. But it's just minor nitpicking :)

cards

PR #2260, opened 2019-10-03
Add Dependabot config

larouxn:
### Changes Summary
Adds a Dependabot config. I can see we already have Dependabot running (https://github.com/elementary/website/pull/2254) but this will allow us to customize it a bit more. For example, with a config we can set default reviewers (btw, who should those be?), add tags, adjust frequency, and so on.


danrabbit:
I'm not sure I should be a reviewer for website dependency stuff. Maybe @btkostner is a better choice?

lewisgoddard:
I'm guessing we should get a general update done first, plus #2224 and #2265, then merge this. We can always adjust later.

Issue #2130, opened 2018-12-13
LiveCD Credentials


kyle-miller:
### Expected behavior

Default LiveCD credentials are easily available online.

### Actual behavior

I was not able to find the LiveCD credentials anywhere online via your elementaryos website or a search.

### Steps to reproduce the behavior

1. Start a LiveCD session
2. Log out

### System information

- **Operating system**: elementary OS
- **Operating system version**: 5.0-stable.20181016
- **Browser**: Epiphany
- **Browser version**: bundled version
- **Enabled Browser Extensions**: bundled only

cassidyjames:
I believe the user name is `elementary` and the password is blank.

aydin41k:
@cassidyjames  Where would you like this to be included?

cassidyjames:
Probably in the instructions at https://elementary.io/docs/installation#booting-from-the-installation-medium

aydin41k:
Should I also add a line to the **_lang** files?

cassidyjames:
@aydin41k no, those will be updated automatically from Weblate.

PR #2266, opened 2019-10-09
Added LiveCD creds to installation.md (#2130)

aydin41k:
Fixes #

### Changes Summary

- 
- 
- 


docs/installation.md
```diff
@@ -75,8 +75,11 @@ In order to start the installation process, you must boot your computer from the
 * Press <kbd>F12</kbd> (or the appropriate key) and select the install drive&mdash;usually "USB-HDD" or something containing the word "USB", but the wording may vary. If you choose the incorrect drive, your computer will likely continue to boot as normal. Just restart your computer and pick a different drive in that menu.
 * Shortly after selecting the appropriate boot drive, you should be presented with the elementary OS splash screen. You may now follow the on-screen instructions which will guide you through the rest of the process.
 
-</div>
+## Default LiveCD login credentials
+
+* Enter <kbd>elementary</kbd> as the username and leave the password field blank in order to log into the LiveCD instance
 
+</div>
```
lewisgoddard: I'm pretty sure the Live CD should never lock, or at least isn't intended to, although I have had several reports of it doing so. As such, I cannot confirm that these are the right credentials.

Secondly, I believe we would want this outside of the source-os specific switcher rather than just being in the Windows area.

jeremypw:
@danrabbit Has no recent activity. Does this need closing or converting to draft? I do not have the necessary rights.

PR #2280, opened 2019-10-20
Update docs with logging ref, fixes #2247

### LINKED ISSUE #2247: [Docs → Reference] Explain log levels and when to use them
i.e. when to use `debug ()`, `warning ()`, and `critical ()`, as mentioned in https://github.com/elementary/code/pull/705#issuecomment-529104283

xevenheaven:
Fixes #2247 

### Changes Summary

- Added logging information and reference
- Based on vala logging

This pull request **is** ready for review.

### DIFF
--- a/docs/code/reference.md
+++ b/docs/code/reference.md
@@ -349,7 +349,7 @@ There are various logging methods to use and here are some sample references for
 
 ## Critical {#critical}
 
-Critical log level is used when there is a severe application failure that causes premature termination and should be investigated immediately.
+Critical log level is used when there is a severe application failure that should be investigated immediately.
 
 ```vala
 public static int main (string[] args) {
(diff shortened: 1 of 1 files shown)


peteruithoven:
Not necessarily a bad thing but there is some overlap with: https://elementary.io/docs/code/os-dev#debug-logs

xevenheaven:
Ooooh I didn't know of the existence of that docs.

After looking through, it seems to me that [docs/code/os-dev](https://elementary.io/docs/code/os-dev#debug-logs) are catered to peeps who want to contribute to elementaryOS.

While having it on [docs/code/reference](https://elementary.io/docs/code/reference#reference) is more for peeps who is using elementaryOS to build their own apps.

I could combine them to one place, but I think separate might be better. Any thoughts?

docs/code/reference.md
```diff
@@ -343,6 +343,95 @@ public override bool configure_event (Gdk.EventConfigure event) {
 }
 ```
 
+# Logging {#logging}
+
+There are various logging methods to use and here are some sample references for you.
+
+## Critical {#critical}
+
+Critical log level is used when there is a severe application failure that causes premature termination and should be investigated immediately.
```
donadigo: I think `that causes premature termination and should be investigated immediately.` belongs more to the `error` than to `critical`. Critical is just a print in the console while error actually exits.

xevenheaven: Updated!

Issue #1663, opened 2017-04-23
Move markdown parser to upstream composer version


btkostner:
Currently we have a local copy of a markdown parser we use for the docs. We should instead use an upstream version downloaded with composer.

PR #2285, opened 2019-10-25
#1663 Replace Parsedown with composer package

mathus13:
#1663 Replace local Parsedown and ParsedownExtra with upstream versions via composer


lewisgoddard:
Also see https://github.com/elementary/website/blob/master/docs/_mdr/settings.php#L43 and https://github.com/elementary/website/blob/master/docs/_mdr/index.php#L76

mathus13:
Sorry, missed that manual inclusion.

PR #2301, opened 2019-12-04
Try to be more explicit about 401s

cassidyjames:
i.e. for the Distrowatches and other bad actors linking directly to (expiring) ISO links.

### DIFF
--- a/410.php
+++ b/410.php
@@ -16,7 +16,7 @@
     </div>
     <div class="column alert">
         <h3>Whoops! That link has expired.</h3>
-        <p>If you’re trying to download elementary OS, please head back to <a href="/">the home page</a> for the most up-to-date version. If you were brought here from a third-party website, it is possible they were attempting to bypass the payment prompt or installation directions. Please inform them that download links are temporary, per-user, and load-balanced—the correct way to link to the download is to link to <a href="https://elementary.io">elementary.io</a>.</p>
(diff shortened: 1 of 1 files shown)


410.php
```diff
@@ -16,7 +16,8 @@
     </div>
     <div class="column alert">
         <h3>Whoops! That link has expired.</h3>
-        <p>If you're trying to download elementary OS, please head back to the home page. If you were looking for something else, feel free to <a href="https://github.com/elementary/website/issues/new">file an issue on our GitHub</a>.</p>
+        <p>If you’re trying to download elementary OS, please head back to <a href="/">the home page</a> for the most up-to-date version. If you were brought here from a third-party website, it is possible they were attempting to bypass the payment prompt or installation directions. Please inform them that download links are temporary, per-user, and load-balanced—the correct way to link to the download is to link to <a href="https://elementary.io">elementary.io</a>.</p>
```
btkostner: > per-user, and load-balanced—the correct way

That reads weird to me

cassidyjames:
```suggestion
        <p>If you’re trying to download elementary OS, please head back to <a href="/">the home page</a> for the most up-to-date version. If you were brought here from a third-party website, it is possible they were attempting to bypass the payment prompt or installation directions. Please inform them that download links are temporary, per-user, and load-balanced. The correct way to link to the download is to link to <a href="https://elementary.io">elementary.io</a>.</p>
```

PR #2315, opened 2020-02-06
Bring back the indiegogo toast

danrabbit:


_scripts/pages/main.js
```diff
@@ -15,3 +16,9 @@ import '~/twitter-links'
 // Send some analytic information on every page load
 ga('send', 'pageview')
 ga('send', 'event', 'Language', 'Pageload', document.documentElement.lang)
+
+jQuery.then(($) => {
```
btkostner: The `.toast` or the `.overlay`? This is pretty easy to use regular JS.

danrabbit: I just copied from the old JS. I’d imagine we could honestly just use CSS for this too right?

PR #2333, opened 2020-04-02
Update Community Slack link

danrabbit:


get-involved.php
```diff
@@ -110,7 +110,7 @@
             <p>Both elementary OS and our website are openly translated using an online platform called Weblate. <a href="/docs/translation-guide" class="read-more">Learn More</a></p>
 
             <div class="actions">
-                <a class="button flat" href="https://join.slack.com/t/ele-l10n/shared_invite/enQtNDkyOTA1MTEzMTQxLWQ1N2YyNDkwYTQwMjA4NWI1MTA5Mzk3MTQyNmMzNmMwODgyYWRjOWQwZDFkZjY5MzVjMTlmMDBlZGJlYmNiNWE" target="_blank" rel="noopener">Join Chat</a>
+                <a class="button flat" href="https://join.slack.com/t/elementarycommunity/shared_invite/zt-d8cswzw1-03M9gXN~e9cOWTpymnNKcw" target="_blank" rel="noopener">Join Chat</a>
```
cassidyjames, line 113: Is this intentionally changing from the l10n Slack to the Community one?

lewisgoddard:
Do we no longer have a separate localization slack, or did you not mean to update ele-l10n to elementarycommunity?

get-involved.php
danrabbit: Oops, no. That was a mistake

PR #2352, opened 2020-04-23
Add shipping notices throughout store, switch to alert style

### LINKED ISSUE #2328: Add notice to store about COVID-19 delays
Printful is likely to have fulfilment delays due to stay-at-home orders and reduced production. https://www.printful.com/faq/products/465-covid-19-updates-for-printful-customers

cassidyjames:
Improvement to #2349 to fix #2328. Previous link was intended for merchants, not end-users (but was better than nothing). This follows some of Printful's guidance about messaging and adds an alert to the main store page and the cart, plus a paragraph to the shipping section and below item descriptions.

The description is added to the config so we can update it in one place, and I did it as an event so we can more easily change the dates it's affected and end it. If it makes more sense, we can also just check for the existence of the description instead of using an event since we don't know the end date.

~Requires back-end changes to add the notice to the config and the event.~

### DIFF
--- a/_backend/config.example.php
+++ b/_backend/config.example.php
@@ -41,4 +41,5 @@ return array(
     'twitter_access_token'    => 'test_atoken',
     'twitter_access_secret'   => 'test_asecret',
 
+    'covid_estimate' => 'Current estimates are 13–16 business days for apparel shipping from the US, 4–7 business days for apparel shipping from the EU, 6–9 business days for mugs shipping from the US, and 2–4 business days for mugs shipping from the EU.',
 );
--- a/store/index.php
+++ b/store/index.php
@@ -20,7 +20,7 @@
     );
 
     include $template['header'];
-    // include $template['alert'];
+    include $template['alert'];
 
     $products = \Store\Product\get_products();
 
--- a/store/index.php
+++ b/store/index.php
@@ -20,7 +20,7 @@
     );
 
     include $template['header'];
-    include $template['alert'];
+    // include $template['alert'];
 
     $products = \Store\Product\get_products();
 
@@ -49,7 +49,7 @@
             </div>
             <div class="icon-text">
                 <h3>You are missing API keys</h3>
-                <p>You are viewing a developmental version of the store without configuring api keys. This will lead to false positives and incorrect errors. Please set your keys to testing configuration.</p>
+                <p>You are viewing a development version of the store without configuring API keys. This will lead to false positives and incorrect errors. Please set your keys to testing configuration.</p>
             </div>
         </div>
     </div>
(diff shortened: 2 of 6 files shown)


lewisgoddard:
@btkostner Can you confirm we no longer need to add these notices to infrastructure, just here is fine since we changed the release process?

btkostner:
@lewisgoddard

store/index.php
```diff
@@ -20,7 +20,7 @@
     );
 
     include $template['header'];
-    include $template['alert'];
+    // include $template['alert'];
```
cassidyjames, line 23: Oops

```suggestion
    include $template['alert'];
```

store/index.php
```diff
@@ -49,7 +49,7 @@
             </div>
             <div class="icon-text">
                 <h3>You are missing API keys</h3>
-                <p>You are viewing a developmental version of the store without configuring api keys. This will lead to false positives and incorrect errors. Please set your keys to testing configuration.</p>
+                <p>You are viewing a development version of the store without configuring API keys. This will lead to false positives and incorrect errors. Please set your keys to testing configuration.</p>
```
cassidyjames, line 52: While I'm here, cleaning up some copy. This is only shown for local development anyway, but it was bugging me. :sweat:

store/index.php
```diff
@@ -61,11 +61,20 @@
 <section class="grid">
     <div class="two-thirds">
         <h2>Support Development. Get Swag. Win Win.</h2>
-        <p>Every purchase goes towards developing elementary OS, its apps, and its services. We're a small <a href="/team">team</a>, mostly volunteer, working constantly to make elementary better. Every little bit of help is one step closer to hiring another full-time developer.</p>
-        <p>Due to COVID-19 some orders may experience delays. Please visit Printful's <a href="https://www.printful.com/covid-19" target="_blank" rel="noopener" title="Covid-19 updates for Printful customers">fullfillment estimate page</a> for more details.</p>
+        <p>Every purchase goes towards developing elementary OS, its apps, and its services. We're a small <a href="/team">team</a>, mostly volunteer, working constantly to make elementary better—your support helps make elementary OS more sustainable.</p>
```
cassidyjames, line 64: Removed the COVID paragraph here since we moved to an alert. At the same time, iterate on this copy just a tiny bit.

store/index.php
```diff
@@ -146,7 +159,12 @@
 <section class="grid">
     <div class="two-thirds">
         <h2>Worldwide Shipping</h2>
-        <p>We now ship all around the world! Place your order and choose from a number of shipping methods to fit your needs. Orders are made on-demand typically within 2–7 days.</p>
+        <p>We ship all around the world! Place your order and choose from a number of shipping methods to fit your needs. Orders are made on-demand typically within 2–7 days.</p>
```
cassidyjames, line 162: We don't need to say we "now" ship all around the world, because we have for a while now.

_backend/config.example.php
```diff
@@ -41,4 +41,5 @@
     'twitter_access_token'    => 'test_atoken',
     'twitter_access_secret'   => 'test_asecret',
 
+    'covid_estimate' => 'Current estimates are 13–16 business days for apparel shipping from the US, 4–7 business days for apparel shipping from the EU, 6–9 business days for mugs shipping from the US, and 2–4 business days for mugs shipping from the EU.',
```
cassidyjames, line 44: This includes the latest relevant estimates from https://www.printful.com/covid-19. I expect we may need to iterate on these from time-to-time as estimates change.

PR #2364, opened 2020-05-05
Create update_translations_from_source.yml

lewisgoddard:


.github/workflows/update_translations_from_source.yml
```diff
@@ -0,0 +1,34 @@
+name: Update Translations from Source
+
+on:
+  push:
```
btkostner, line 4: We should probably just do this as cron. Otherwise this will run when translations update and get pushed to master creating a possible loop.

.github/workflows/update_translations_from_source.yml
```diff
@@ -0,0 +1,34 @@
...
+    - name: Setup GitHub Checkout
+      uses: actions/checkout@v2
+    - name: Setup PHP Action
+      uses: shivammathur/setup-php@1.8.7
+      with:
+        php-version: '7.4'
+    - name:  Update Translations from Source
+      run: |
+        php _backend/Console/Translations.php
+        echo "Setting up git credentials..."
```
btkostner, line 28: I'd probably split this to two steps. One to run the PHP script, and another to push the files.

.github/workflows/update_translations_from_source.yml
```diff
@@ -0,0 +1,37 @@
...
+  push:
+    branches:
+      - master
+      - auto-update-translations
+  schedule:
+    # * is a special character in YAML so you have to quote this string
+    - cron:  '0 0 * * *'
+
+jobs:
+  build:
```
btkostner, line 13: also change this to `translate`

btkostner:
The current action error is because we also need to setup npm and run the install and build scripts.

.github/workflows/update_translations_from_source.yml
lewisgoddard: It only commits and pushes when there's a change, so it shouldn't do.

.github/workflows/update_translations_from_source.yml
lewisgoddard: ??

PR #2370, opened 2020-05-06
Automate update of Composer libraries

lewisgoddard:


btkostner:
I think dependabot works with composer. I think we should go with that option instead of this

lewisgoddard:
@btkostner I think that only looks for security advisories, this is more liberal.

btkostner:
Updated to use dependabot instead

PR #2372, opened 2020-05-06
Automate updates of Submodules

lewisgoddard:

### DIFF
--- a/.github/workflows/update_submodules.yml
+++ b/.github/workflows/update_submodules.yml
@@ -1,0 +1,47 @@
+---
+
+name: Update Submodules
+
+on:
+  schedule:
+    - cron:  "0 0 * * *"
+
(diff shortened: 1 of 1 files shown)


btkostner:
For these, we should just put them all in a single `update.yaml` file with a single cron job. It will clean up the actions tab in GitHub because there will only be one thing instead of an `update submodules` and `update translations` etc.

lewisgoddard:
@btkostner That seems far more likely to break, a single error would halt the whole process.

btkostner:
@lewisgoddard no it wouldn't. It only stops things that depend on it. If we had translations and submodules on the same pipeline, they would run at the same time.

lewisgoddard:
That still seems like an uncomfortable mix of disciplines.

PR #2419, opened 2020-05-25
Add Devices & Retailers to Store

### LINKED ISSUE #2412: Add retailers to store page
Following #2411

cassidyjames:
Fixes #2412. Currently limiting to retailers we've recently talked to and arranged some sort of referral or other understanding of contribution.


cassidyjames:
Limiting this to Laptops With Linux since they've directly communicated with us and provided a referral link right away. Other retailers can be added on a case-by-case basis.

PR #2428, opened 2020-06-03
Store: Add Star Labs

cassidyjames:
We're not listed on their site yet, but should be next week. Opening here so it's faster to merge when needed.


cassidyjames:
We're live on their site, now. This is ready to go.

PR #2457, opened 2020-06-23
Add Plausible Goals

lewisgoddard:
### Plausible Change to Tracking

- Remove error tracking because Plausible does not support event metadata.
- Simplify Payment tracking to Skipped/Initiated/Complete/Failed.
- Remove language tracking because plausible does not support event metadata.
- Simplify Store tracking to exclude variables.

### Goals to Add (27)

- Error 410: Download Link Expired
- Hera 5.1.5 Payment: Skipped/Initiated/Complete/Failed
- Hera 5.1.5 Download from OS: Other/Android/iOS/Windows/macOS/Linux/Unknown
- Hera 5.1.5 Download from Region: nyc3/sfo1/ams3/fra1/sgp1
- Hera 5.1.5 Download Method: HTTP/Magnet
- Juno 0.4.1 Download of Previous Version
- Store: Visit
- Store: View Product
- Store: Cart Visit
- Store: Country Change
- Store: Checkout Visit
- Store: Payment Processed
- Store: Order Completion

### Notes

Plausible doesn't support event metadata. Would be a lot cooler if it did. https://github.com/plausible/analytics/issues/134

This branch does not remove Google Analytics, `plausible-only` will do that.


cassidyjames:
@lewisgoddard I do still think I want some measurable overlap between using Google Analytics and Plausible, just for a sanity check. Would it be possible to split this into two PRs? One to add Plausible Goals, and the other to remove Analytics?

PR #2460, opened 2020-06-24
Remove Google Analytics

lewisgoddard:
### Changes Summary

- Remove Google Analytics


_scripts/pages/main.js
```diff
@@ -3,20 +3,15 @@
  * Loads all of the site wide snippets
  */
 
-/* global ga */
+/* global */
```
btkostner, line 6:
```suggestion
```

lewisgoddard:
I have removed references to Google Analytics and added a Do-Not-Track exclusion to Plausible being loaded. We used to honour that before so I assume we want to now, or do we think the additional privacy should be enough?

cassidyjames:
Considering Do Not Track has mostly been abandoned by browser vendors, I don't know that it's necessary—especially since we're not sharing stats with third parties, either.

lewisgoddard:
Okay, I have reverted that and removed the section from the privacy page mentioning it.

danrabbit:
Should we add something to the privacy section about how to opt-out of plausible or at least a TL;DR of their privacy stance and a link to either their homepage or our public dashboard?

lewisgoddard:
@danrabbit Possibly, you can see where the equivalent GA stuff was removed above. is there an opt-out for Plausible.io?

danrabbit:
I guess it would just be blocking stats.elementary.io or something right?

btkostner:
Browsers still use `do not track`. We should just check that for loading the plausable script.

lewisgoddard:
W3C disbanded the group developing the standard in January 2019.

Apple is dropping DNT from Safari and Google are expected to too, which will probably filter down to Edge/Opera.

Sites like Twitter and Microsoft are just starting to ignore it.

cassidyjames:
(Further, Plausible emphatically _does not track users_. See: https://plausible.io/data-policy)

privacy.php
```diff
@@ -24,21 +24,13 @@
...
-                <h3>Do Not Track</h3>
-                <p>This site honors the do-not-track setting found in most modern browsers and disables Google Analytics when it is present. We cannot automatically remove other tracking methods without severely impacting the use of the site. Should you wish to manually manage these tracking methods, reference the section on disabling cookies.</p>
                 <h3>Cookies</h3>
                 <p><strong>You can choose to disable or selectively turn off any cookies or third-party cookies in your browser settings.</strong></p>
                 <p>This site uses cookies for incremental improvements. You may find the services function without them but at a reduced usability. For example, the site will not remember if you have previously paid for elementary OS; by default you will be asked to pay again.</p>
                 <h5 data-l10n-off>Cloudflare</h5>
                 <p>Stores cookies to log behavioral elements and analyze potential threats. For more information, see the <a class="read-more" target="_blank" rel="noopener" href="https://www.cloudflare.com/security-policy">Cloudflare Privacy &amp; Security Policy</a></p>
```
cassidyjames, line 33: For transparency, we should add something like this:

```suggestion
                <p>Stores cookies to log behavioral elements and analyze potential threats. For more information, see the <a class="read-more" target="_blank" rel="noopener" href="https://www.cloudflare.com/security-policy">Cloudflare Privacy &amp; Security Policy</a></p>
                <h5 data-l10n-off>Plausible</h5>
                <p>We use the open source Plausible Analytics routed through our <code>stats.elementary.io</code> domain to count website visits, downloads, etc. No personal data—not even an IP address or browser user agent—is stored. <a class="read-more" target="_blank" rel="noopener" href="https://plausible.io/data-policy">Plausible Data Policy</a></p>
```

danrabbit: Yeah I like this. Should we try to fit in a link to the dashboard here too or naw?

cassidyjames: Oh yeah good call. Updated.

cassidyjames:
Okay, I added a "Plausible Analytics" section to the elementary.io section of the privacy policy. I kept it out of the cookies section since it doesn't actually use cookies at all.

If we're happy with the data we're getting from Plausible, I'd be happy to get this merged.

lewisgoddard:
This will cause errors if we don't clear Cloudflare on merge.

PR #2461, opened 2020-06-24
Fix #2459: Disable Plausible for local development and beta subdomain

### LINKED ISSUE #2459: Disable Plausible for local development and beta subdomain
We probably don't want those stats being grouped in with the others.

lewisgoddard:
Fixes #2459

### Changes Summary

- Stop loading Plausible unless it's a production location.
- Keep the array snippet to avoid quoting out tracking everywhere.
- You can still check if they work by putting `plausible` in console.

### DIFF
--- a/_templates/header.php
+++ b/_templates/header.php
@@ -87,7 +87,7 @@ $l10n->begin_html_translation();
         <script src="scripts/common.js"></script>
         <script src="scripts/main.js" async></script>
         
-        <?php if ( getenv('PHPENV') == 'production' ) { ?>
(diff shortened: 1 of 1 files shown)


_templates/header.php
```diff
@@ -86,7 +86,10 @@
         <?php if (!isset($scriptless) || $scriptless === false) { ?>
         <script src="scripts/common.js"></script>
         <script src="scripts/main.js" async></script>
+        
+        <?php if ( getenv('PHPENV') == 'production' ) { ?>
```
btkostner, line 90: I think this is also `production` for the beta domain because they are the same nginx and php instance

lewisgoddard:
Does work.

PR #2475, opened 2020-07-06
Make the text lighter and easier to read

yarons:
The text was very dark and I couldn't read it so I changed the text color to be brighter.

### DIFF
--- a/_styles/solarized_dark_bash.css
+++ b/_styles/solarized_dark_bash.css
@@ -4,7 +4,7 @@
 
 .language-bash.hljs {
     background: #002b36;
-    color: #839496;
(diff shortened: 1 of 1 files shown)


lewisgoddard:
I would have thought that this was pulled from the application solarized palette and that it was checked for contrast there. Calling @elementary/ux

yarons:
I think that it would be better to switch to another theme.

danrabbit:
Yeah this was pulled from solarized. We've since switched colors on desktop to a more neutral palette. So the new values are:
```
background: #2e2e2e
color: #a5a5a5
```

which should pass WCAG AA

PR #2491, opened 2020-07-23
Add Dockerfile

btkostner:
This adds a `Dockerfile` that will build and host the website.

This moves the actual PHP running from `nginx` to `apache` mainly because there is no _official_ php nginx image, and I would rather just use `apache` instead of making our own. Easier to update. Easier to run. etc. etc.

Things to do:
- [ ] Port nginx config to `.htaccess`
- [ ] Setup configuration to load from environment variables
- [ ] Test
- [ ] Deploy GitHub action to publish the image to DockerHub

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -1,0 +1,27 @@
+<Files ~ "\/(.|_)">
+    Deny from all
+</Files>
+
+<Files ~ "^\/node_modules\/.*">
+    Deny from all
+</Files>
+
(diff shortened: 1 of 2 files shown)


StylusEater:
@btkostner is this still in progress?

btkostner:
I have not touched this since the initial commit, so it's pretty stale

StylusEater:
@btkostner Thank you for confirming.

davidmhewitt:
Superseded by #3160

PR #2518, opened 2020-09-17
Switch to Inter, Add Slimbook

### LINKED ISSUE #2464: Remove Google Fonts
##Prerequisites
- [x] I have searched open and closed issues for duplicates.

## Feature
**Is your feature request related to a problem? Please describe.**
Google has a lot of privacy concerns, and there are plenty of alternatives specifically to this subject, I know that you already use FontAwesome, but you also keep using Gfonts.

N/A

**Describe the solution you'd like**
To stop using completely the service and use a libre alternative instead.

**Existing work**
A lot of websites do not use Gfonts, I think there are too many to list them, and it's kind of irrelevant.

**Describe alternatives you've considered**
https://alternativeto.net/software/google-web-fonts/?license=opensource

**Additional context**
N/A

cassidyjames:
This ended up being bigger than expected, but I needed Inter in order to do the typography I wanted in the store. :sweat_smile: But in the end, it's not actually that huge of a diff, just a handful of little things plus the 150+ line Inter web font CSS (self-hosted instead of on Google Fonts), plus the ~100-line license.

Screenshot_2020-09-16 Store ⋅ elementary

- [x] Add Slimbook to Devices section of Store
- [x] Revamp Devices section design
- [x] Remove Google Fonts, fixes #2464
- [x] Remove embedded Raleway
- [x] Switch to Inter
- [x] Probably embed Inter so we can rely on its style/weights
- [x] Clean up any pages from fallout from switching to Inter
  - [x] Home
  - [x] Support
  - [x] Thank You/Previous
  - [x] Privacy
  - [x] Team
  - [x] Press
  - [x] Open Source

Once nice thing is that this gets our page heading styles much more consistent without bad semantics like using `<h4>`s like we previously did.


cassidyjames:
As a note, you can test that the embedded Inter is working by running this on elementary OS 5 without Inter installed, or probably easier: visit DEVMACHINEIP:8000 on another device like a phone, tablet, etc. on the same network.

_styles/main.css
```diff
@@ -97,6 +82,10 @@ h1 .logotype-svg {
     }
 }
 
+h1 + p {
+    font-size: 1.25em;
+}
```
danrabbit, line 87: I'm not sure I'm a fan of this. Maybe something a little more subtle like:
```suggestion
h1 + p {
    font-size: 1.15em;
}
```

_styles/main.css
```diff
@@ -72,9 +58,8 @@ a:focus {
 }
 
 h1 {
-    font-family: "Raleway";
     font-size: 32px;
-    font-weight: 100;
+    font-weight: 700;
```
danrabbit, line 62: This looks dope

danrabbit (changes requested):
This is mostly awesome. Great work!

I'm not sure if it's outside the scope of this branch, but it looks like we're being pretty inconsistent with headers. For example with the headers on these sub-sections. Maybe we should make sure to always use h2 for these half and third sections and h1 for the main header:

Screenshot from 2020-09-17 10 21 11@2x

Screenshot from 2020-09-17 10 21 44@2x

Screenshot from 2020-09-17 10 22 22@2x

and I think we should be more consistent with headers in sections. For example we could do:

Screenshot from 2020-09-17 10 23 32@2x

_styles/main.css
danrabbit: Alternatively (or maybe additionally) I feel like we should decrease line height a bit here. As in:

Screenshot from 2020-09-17 10 29 42@2x

cassidyjames:
@danrabbit I think I want to do another pass later where we clean up consistency with headings and whatnot for sure. I can pull the line-height down a bit on the `h1 + p` though, I agree. I bumped up the font size to match what we did on some pages, but kept the line-height of the smaller text on other pages. :sweat_smile:

cassidyjames:
@danrabbit thoughts on pulling the `h1 + p` line-height down to this?

Screenshot from 2020-09-17 11 45 17@2x

Screenshot from 2020-09-17 11 45 27@2x

Screenshot from 2020-09-17 11 45 37@2x

danrabbit:
@cassidyjames that looks solid!

danrabbit (approved):
Looks good. We can come back into another branch to clean up heading sizes

PR #2520, opened 2020-09-17
Open Source: Revise copy

cassidyjames:
Right now, the top copy just talks about being built _on top_ of open source. In reality, I feel like there are three things to convey:

- We're built _on_ open source
- Everything _we build_ is open source
- We collaborate _within_ open source

These three things are unique compared to major OSes; Apple and Microsoft can say they build _on_ open source, but their products are not typically open themselves. Google can say they build on open source and make open source things, but they don't really collaborate within the ecosystem.

Screenshot from 2020-09-17 11 38 18@2x

(Was built on #2518, but broken up for easier review/merging… so ignore all those commits that have now been merged!)


danrabbit (approved):
Great distinctions 👍

PR #2522, opened 2020-09-17
Move Open Source copy to Home

danrabbit:
I feel like this makes more compelling arguments than we currently have on the home page for security and why open source is important.

I also trimmed down some of the copy in these supporting paragraphs. And I left out the "Go fast" since I don't feel like it's particularly compelling. We don't have any metrics other than just "cause we say so" and no real supporting arguments about what makes things faster from a design or tech perspective

I also moved this down near the keyboard shortcut stuff because it feels more technical to me and related to that persona and to keep the layouts from being too samey in a row

## BEFORE
Screenshot from 2020-09-17 12 26 43@2x

## AFTER

Screenshot from 2020-09-17 12 22 03@2x


cassidyjames:
@danrabbit I like the content, but I feel like those half-sections are too close to the section above. I think if you close the grid and open a new one there, it adds some padding?

index.php
```diff
@@ -466,9 +450,26 @@
...
-        <section id="shortcuts">
+        <section class="grid">
+            <div class="two-thirds">
+                <h1>Everything We Do is Open&nbsp;Source</h1>
+                <p>Our platform itself is entirely open source, and it’s built upon a strong foundation of Free &amp; Open Source software. Plus, we actively collaborate within the ecosystem to improve it for everyone.</p>
+                <a class="read-more" href="/open-source">Explore Our Stack</a>
+            </div>
+            <div class="half">
+                <h2>Secure &amp; Privacy-respecting</h2>
+                <p>When source code is available to audit, anyone — a security researcher, a concerned user, or an OEM shipping the OS on their hardware — can verify that the software is secure and not collecting or leaking personal information.</p>
```
cassidyjames, line 461: No spaces around em-dashes

```suggestion
                <p>When source code is available to audit, anyone— a security researcher, a concerned user, or an OEM shipping the OS on their hardware —can verify that the software is secure and not collecting or leaking personal information.</p>
```

cassidyjames (approved):
:shipit:

PR #2523, opened 2020-09-17
Store: update COVID notices; move devices up top

### LINKED ISSUE #2519: Retire COVID-19 delays from store
Sounds like things are back to normal for the products we ship. I'm comfortable retiring at least the big alert, if not all COVID-related notices at this time.

https://www.printful.com/covid-19

cassidyjames:
I opted to keep the more contextual notices about COVID-related delays there, but made the copy a bit more optimistic than before. For context, Printful says production is back on schedule but I know I've seen continued delays with shipping. So better to note that to customers.

Fixes #2519

### DIFF
--- a/_backend/config.example.php
+++ b/_backend/config.example.php
@@ -41,5 +41,5 @@ return array(
     'twitter_access_token'    => 'test_atoken',
     'twitter_access_secret'   => 'test_asecret',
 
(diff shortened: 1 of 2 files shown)


danrabbit (approved):
:rocket:

PR #2524, opened 2020-09-17
Home: move the press to the botom

danrabbit:
Alessandro suggested we de-emphasize this section and yeah I think it looks nice at the bottom:
Screenshot from 2020-09-17 13 43 28@2x


cassidyjames (approved):
:+1:

PR #2525, opened 2020-09-17
Opt out of centered grids

cassidyjames:
So it turns out a lot of our centered text woes is due to the grid using `text-align: center` and then its children being `display: inline-block`. This PR:

- Stops trying to be so clever with centering based on only-children and non-text-related wrapper classes.
- Adds a new `.grid.flex` class that opts out of that in favor of using flexbox to center things. As a result, we get much nicer left-aligned text while still having the contents centered within the grid.

Another option I considered was swapping `.grid` to work this way by default, but that broke pretty much all the things and would end up resulting in rewriting half the stylesheet and DOM at the same time. This lets us opt into the better layout over time.

### DIFF
--- a/_styles/main.css
+++ b/_styles/main.css
@@ -564,6 +564,11 @@ footer ul li a:focus {
         vertical-align: top;
         width: calc(66% - 48px);
     }
+
+    .grid *:only-child p,
+    .grid .two-thirds h1 + p {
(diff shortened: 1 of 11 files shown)


danrabbit:
Closing because it's old and conflicts a bunch

PR #2526, opened 2020-09-17
CSS: remove most of column

danrabbit:
afaict the only place column is still used at all is the privacy page

### DIFF
--- a/_styles/main.css
+++ b/_styles/main.css
@@ -462,36 +462,10 @@ footer ul li a:focus {
     vertical-align: top;
 }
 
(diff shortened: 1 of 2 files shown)


RMcNeely:
Is the goal to remove the just the descendent styles from column?

danrabbit:
Just trying to get rid of some old styles so we have one shitty outdated grid system instead of 2 lol

RMcNeely (approved):
I can't find any more uses of these classes. It looks good to me :+1:

Issue #2534, opened 2020-09-28
Please remove the redundant Spanish localizations and leave only “es”


fitojb:
With the time I have it’s impossible for me to complete all **six** (!) of es, es_419, es_AR, es_MX, es_PE and es_US.

In Spanish technical translations, country locales aren’t crucial or needed, unlike the situation of Brazilian vs. European Portuguese, where there are numerous grammar and vocabulary differences that are hard to reconcile.

fitojb:
This was requested at the translators Slack back in 2018: https://app.slack.com/client/T52UMUPAN/C53NW9DH9

fitojb:
@elementary

btkostner:
The problem isn't that the folder exists, it's that our site will use any language in that folder instead of the hard coded list we have. @lewisgoddard I think we should change that. Only the languages that we hard code should be translatable to. Everything else just exists until it's added to that list.

The reason I don't want to just delete those folders is that weblate may add them back, and if I remember correctly, we would need to disable the languages globally for that not to happen. I don't want to touch the desktop stuff.

fitojb:
@btkostner Yeah, the country locales need to be deleted separately on Weblate so they aren’t added back to the repo.

lewisgoddard:
I think we should try and delete these and see what happens without translation source files.

PR #2548, opened 2020-10-01
only recognize languages hard set in our list

btkostner:
This _should_ make it so the website only translates to languages in our hard set list


btkostner:
Fixes #2534

lewisgoddard (approved):

PR #2581, opened 2020-10-28
Plausible Metadata

lewisgoddard:
### Changes Summary

- Move detectedOS to widget
- Add metadata to downloads
- Collate payments more sensibly

```
plausible('Downloads', {
  meta: {
    Region: config.user.region,
    Method: downloadMethod,
    OS: detectedOS(),
    Version: config.release.version
  }
})
```

Thanks to plausible/analytics#381


_scripts/previous.js
```diff
@@ -8,13 +8,26 @@
...
         // ACTION: .download-http.click: Track download over HTTP
         $('.download-link').click(function () {
-            plausible('Download of Previous Version')
-            plausible('Download of ' + config.previous.title + ' ' + config.previous.version)
+            if ($(this).hasClass('http')) {
+                var downloadMethod = 'HTTP'
+            }
+            if ($(this).hasClass('magnet')) {
+                var downloadMethod = 'Magnet'
+            }
```
btkostner:
```suggestion
			let downloadMethod = 'Unknown'

            if ($(this).hasClass('http')) {
                downloadMethod = 'HTTP'
            } else if ($(this).hasClass('magnet')) {
                downloadMethod = 'Magnet'
            }
```

_scripts/pages/download.js
```diff
@@ -115,50 +116,44 @@ Promise.all([config, jQuery, Payment, modal]).then(([config, $, Payment]) => {
 
             // Free download
             if (Number.isNaN(paymentAmount) || paymentAmount < paymentMinimum) {
-                plausible('Payment: Skipped')
+                plausible('Payment', {
+                    meta: {
+                        Input: paymentAmount,
```
btkostner, line 121: I'm not sure we want Input here. We are going to end up with a list of mostly one events for every invalid value. I think `Amount` is good enough.

btkostner (changes requested):
One thing and then it should be good

_scripts/pages/download.js
lewisgoddard: I wanted to get the same info for all 3 payment results (Skip, Pay, Fail) to see what inputs were causing which action.

btkostner: Mhmm, alright. We will probably have a lot of noise though

btkostner (approved):

_scripts/previous.js
```diff
@@ -8,13 +8,27 @@
...
-            plausible('Download of Previous Version')
-            plausible('Download of ' + config.previous.title + ' ' + config.previous.version)
+            let downloadMethod = 'Unknown'
+            if ($(this).hasClass('magnet')) {
+                downloadMethod = 'Magnet'
+            }
+            if ($(this).hasClass('http')) {
+                downloadMethod = 'HTTP'
+            }
+            plausible('Downloads', {
```
cassidyjames, line 24: `Download` makes more sense here and is what we have in Plausible dashboard.

PR #2586, opened 2020-10-31
Track additional data in stores

lewisgoddard:
This adds back tracking we lost because Plausible didn't support it for a while.

### DIFF
--- a/_scripts/pages/store/cart.js
+++ b/_scripts/pages/store/cart.js
@@ -101,11 +101,19 @@ Promise.all([jQuery]).then(([$]) => {
          *
          * @return {void}
          */
-        const updateAddressForm => {
+        const updateAddressForm = (notify = true) => {
             const form = $('form[action$="checkout"]')
             const value = $('select[name="country"]', form).val()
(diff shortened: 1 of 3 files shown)


_scripts/pages/store/cart.js
```diff
@@ -107,7 +107,13 @@ Promise.all([jQuery]).then(([$]) => {
             const $state = $('select[name="state"]', form)
             const $statelabel = $('label[for="state"]', form)
 
-            if (notify) plausible('Store: Country Change') // value
+            if (notify) {
+                plausible('Store: Country Change', {
+                    props: {
+                        Country: value.toString()
```
btkostner, line 113: Is value not already a string? More so, wouldn't it make more sense to merge this with the payment process event so we just get one event of "Purchase" with amount and country?

lewisgoddard: We don't currently have the country as a field on the checkout page, which I assume is why it is being tracked here.

btkostner:
Store is removed from website

PR #2612, opened 2020-12-01
CONTRIBUTING: elementary OS 6 directions

cassidyjames:
Since I always have to figure it out from scratch. It looks like newer than 12.x might have issues building the site, so I recommend 12.x specifically.


btkostner (approved):

cassidyjames:
@btkostner any idea why this is failing CI?

btkostner:
Nope. I'd force push / blank push / something to get it to rebuild. Might have just been a broken master at the time.

PR #2613, opened 2020-12-02
Store: Add support for partner devices; add Star LabTop Mk IV

cassidyjames:
This adds a section up top for partner devices, and adds the Star LabTop Mk IV as the first featured device. Ideally over time this will expand to a small selection of partner devices.

- [ ] Use local image
- [ ] Decide if we like the "Learn More" link where it is

**Do not merge** until I say so, as we have to actually ensure the LapTop Mk IV achieves its partner status first. :smile:

### DIFF
--- a/_styles/store.css
+++ b/_styles/store.css
@@ -1,7 +1,7 @@
-/*************************************
-* Copyright 2016 elementary LLC.     *
-* This file is part of elementary.io *
-*************************************/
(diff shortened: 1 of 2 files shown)


cassidyjames:
@danrabbit said he likes the Learn More link here. We can always iterate on it in the future if we like something else more.

PR #2673, opened 2021-03-18
Add JS alert that is removed with JS

### LINKED ISSUE #2672: Add JS requirement banner
## Problem

If you disable JS, you're unable to download the OS (or at least pay for it afaik).

## Proposal

We add a "Portions of this site require JavaScript" banner that is then hidden with JS.

## Prior Art

Super common on the web, but System76 has a decent version. We could probably be a _little_ more subtle though.

Screenshot from 2021-03-18 10-09-21

cassidyjames:
Fixes #2672 

### Changes Summary

- Adds a JS alert just below the closing `</body>` tag
- Styles it with some simple warning styling
- Immediately removes it with inline JS

Screenshot from 2021-03-18 12-37-36

### DIFF
--- a/_templates/footer.php
+++ b/_templates/footer.php
@@ -38,10 +38,6 @@
                 <li><a href="<?php echo $sitewide['root'].'open-source'; ?>">Open Source</a></li>
             </ul>
         </footer>
-        <div id="js-alert">
-            <p><strong>JavaScript is required</strong> for parts of this site, like downloading elementary OS and some interactive components.</p>
-        </div>
-        <script>document.getElementById("js-alert").remove ();</script>
         <?php
             include $template['legacy'];
             $l10n->set_domain('layout');
--- a/_templates/footer.php
+++ b/_templates/footer.php
@@ -38,6 +38,10 @@
                 <li><a href="<?php echo $sitewide['root'].'open-source'; ?>">Open Source</a></li>
             </ul>
         </footer>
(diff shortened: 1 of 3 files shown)


_templates/footer.php
```diff
@@ -38,6 +38,10 @@
                 <li><a href="<?php echo $sitewide['root'].'open-source'; ?>">Open Source</a></li>
             </ul>
         </footer>
+        <div id="js-alert">
+            <p><strong>JavaScript is required</strong> for parts of this site, like downloading elementary OS and some interactive components.</p>
+        </div>
+        <script>document.getElementById("js-alert").remove ();</script>
```
btkostner, line 44: I'd just put this in a `<noscript>`. Because it's a different script block, even if the regular JS errors out, this will still run, so it's no different than `noscript`.

btkostner:
Thoughts about putting this under the main header? I'd like to be consistent on this with our other sites.

_templates/footer.php
cassidyjames: Derp, somehow I always forget about `<noscript>`

cassidyjames:
@btkostner yeah we could. It was simpler to work within the existing layout by, well, not working within it, lol. But I can look at that.

cassidyjames:
Screenshot from 2021-03-18 12-48-50

btkostner (approved):

PR #2686, opened 2021-04-05
Brand: add default logomark, tweak copy

cassidyjames:
Adds the new logo and tweaks the copy to be a little clearer. Also swaps the order to prioritize the logomark over the logotype.


danrabbit (approved):
Great copy changes. Nice work

cassidyjames:
@btkostner it looks like this isn't deployed; I tried purging the cache for the page itself and the whole site in Cloudflare, but I'm still seeing the old version: https://elementary.io/brand

PR #2734, opened 2021-06-03
Add edw banner

### LINKED ISSUE #2733: edw banner
We should do an edw promo banner that links to edw.elementary.io like we've done in the past for Indiegogo campaigns.

cassidyjames:
Fixes #2733. @danrabbit might want to poke at the styles. :smile: 

Screenshot from 2021-06-03 17-15-04


danrabbit:
That's pretty sick. Seems a little low contrast on some parts. Maybe a box shadow or semibold weight could help that out a bit?

danrabbit (approved):

PR #2785, opened 2021-08-04
Add toast for Tuesday

cassidyjames:
Screenshot from 2021-08-04 16-37-01


btkostner:
Worth remembering that this causes hundreds of browsers to refresh our homepage at the same time, basically DDOSing ourselves, sooooooo we should probably have #web on standby in case stuff crashes.

danrabbit:
I guess if we're not using the countdown anymore we can dump all its styles and JS and whatever else it uses?

cassidyjames:
@danrabbit yeah will do.

danrabbit (approved):
L E T S   G O :shipit:

PR #2786, opened 2021-08-04
Odin toast

cassidyjames:
Alternative to #2785. Uses a toast like we've done for IGG and edw instead of the flip clock countdown.

### DIFF
--- a/_templates/event.php
+++ b/_templates/event.php
@@ -9,7 +9,7 @@ require_once __DIR__ . '/../_backend/event.php';
 
 $l10n->set_domain('layout');
 
-const EVENT_NAME = 'odin 6.0 release';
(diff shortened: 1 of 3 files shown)


_templates/event.php
```diff
@@ -9,7 +9,7 @@
 
 $l10n->set_domain('layout');
 
-const EVENT_NAME = 'odin 6.0 release';
+const EVENT_NAME = 'juno 5.0 release';
```
danrabbit, line 12: This seems like an oops

cassidyjames: Nope, I was explicitly setting it back to the value before this PR. This const controls whether or not the flip clock appears, and doesn't have anything to do with the toast. :wink:

PR #2787, opened 2021-08-06
Use local time and format for toast

cassidyjames:
A nice little enhancement instead of showing UTC everywhere.

- [x] Change time based on location
- [x] Change time format based on location
- [x] Change day format based on location

### DIFF
--- a/_templates/event.php
+++ b/_templates/event.php
@@ -20,10 +20,10 @@ if (event_active(EVENT_NAME)) { ?>
   </div>
   <script>
     const utcDate = new Date("<?php echo date(DATE_ISO8601, date_timestamp_get($event_expires[EVENT_NAME][1])) ?>");
-    let localDay = new Intl.DateTimeFormat('<?php echo $page['lang']; ?>', {weekday: 'long'}).format(utcDate);
-    let localTime = new Intl.DateTimeFormat('<?php echo $page['lang']; ?>', {hour: 'numeric', minute: 'numeric'}).format(utcDate);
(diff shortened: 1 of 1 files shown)


_templates/event.php
```diff
@@ -9,13 +9,22 @@
...
+if (event_active(EVENT_NAME)) { ?>
   <div class="overlay">
     <div class="overlay__content toast">
       <div class="toast__close"><i class="fas fa-times"></i></div>
-      <span class="toast__text"><strong>Something new is coming.</strong> Check back Tuesday after 1600 UTC.</span>
+      <span class="toast__text"><strong>Something new is coming.</strong> Check back <time id="odin-day">Tues</time> after <time id="odin-time">1600 UTC</time>.</span>
     </div>
   </div>
+  <script>
+    const utcDate = new Date("<?php echo date(DATE_ISO8601, date_timestamp_get($event_expires[EVENT_NAME][1])) ?>");
```
RMcNeely, line 22: Is there any benefit to winning the Date in PHP as opposed just making it a JavaScript date?

cassidyjames: I mean we already have the date in PHP so I was just trying to be consistent, especially in case we were to change it (as unlikely as that is).

RMcNeely (approved):

btkostner (approved):
Assuming you tested this and it works, I'm ok with the merge. `DateTimeFormat` doesn't support old IE, but that shouldn't be a problem for us.

PR #2790, opened 2021-08-09
Fix translators slack link 

lewisgoddard:
This will only work for the next 200 signups, but it does work.


keycommander (approved):
Looks neat. Excited to get a look into elementary os.

PR #2793, opened 2021-08-10
Fix language codes

lewisgoddard:


_backend/Lib/L10n.php
```diff
@@ -123,7 +123,6 @@ public static function languageDirectory($lang)
         'mr' => 'मराठी',
         'nb' => 'Bokmål',
         'nl' => 'Nederlands',
-        'pa' => 'ਪੰਜਾਬੀ',
```
cassidyjames, line 126: Do we need to add Punjabi to Weblate, then?

lewisgoddard: I assume we either had Punjabi and removed it or it was never added. Maybe? I am not sure, but it sure isn't here.

cassidyjames: @lewisgoddard there was a PR to add it here and a translator who wanted to translate it in Weblate. I'm not sure what needs to be done to make it available.

ryonakano: Yeah @elSolus requested and I made a PR for adding Punjabi here. But not sure how to add translation files for Punjabi because adding on weblate does not seem to be possible.

PR #2800, opened 2021-08-11
Landing page reworked

comradekingu:
Came to look for the blogpost. "CalDav" should be "CalDAV".

### DIFF
--- a/index.php
+++ b/index.php
@@ -98,7 +98,7 @@
       <div class="grid">
         <div class="two-thirds">
           <h2>What’s New in elementary OS 6 Odin</h2>
-          <p>The biggest update to the platform yet with multi-touch, dark style, app sandboxing, and an all-new installer. OS 6 empowers you to be in control and express yourself, continues to innovate with new features, and is both easier to get and more inclusive than ever.</p>
+          <p>The biggest update to date brings multi-touch, a perceptually dark style, app sandboxing, and an all-new installer. Odin grants visionary insight in the persuit of knowledge, with innovative features making the answers to riddles available to more people.</p>
           <a href="https://blog.elementary.io/elementary-os-6-odin-released/" target="_blank" rel="noopener" class="read-more">Read the Announcement</a>
         </div>
       </div>
@@ -109,7 +109,7 @@
         <div class="app-display__description">
           <img src="images/icons/apps/128/system-software-install.svg" alt="elementary AppCenter icon"/>
           <h2>Get it on <strong>AppCenter</strong></h2>
-          <p>Get free and paid apps on AppCenter, the open, pay-what-you-can app store for indie developers. Each app has been reviewed and curated by elementary to ensure a native, privacy-respecting, and secure experience.</p>
+          <p>Get and buy apps on AppCenter, the open, pay-what-you-can app store for indie developers. Each app is reviewed and curated by elementary to ensure an experience of ease, security, and privacy.</p>
           <div class="buttons">
             <a href="https://appcenter.elementary.io" target="_blank" rel="noopener" class="button flat">Discover AppCenter Apps</a>
             <a href="https://blog.elementary.io/tags#appcenter-spotlight" target="_blank" rel="noopener" class="button flat">Read AppCenter Spotlight</a>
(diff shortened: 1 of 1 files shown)


index.php
```diff
@@ -98,7 +98,7 @@
       <div class="grid">
         <div class="two-thirds">
           <h2>What’s New in elementary OS 6 Odin</h2>
-          <p>The biggest update to the platform yet with multi-touch, dark style, app sandboxing, and an all-new installer. OS 6 empowers you to be in control and express yourself, continues to innovate with new features, and is both easier to get and more inclusive than ever.</p>
+          <p>The biggest update to date brings multi-touch, a perceptually dark style, app sandboxing, and an all-new installer. Odin grants visionary insight in the persuit of knowledge, with innovative features making the answers to riddles available to more people.</p>
```
comradekingu, line 101: Norseman approved :)

index.php
```diff
@@ -122,7 +122,7 @@
       <div class="grid">
         <div class="two-thirds">
           <h1>Get Work Done. Or Play.</h1>
-          <p>Stay productive and focused with Multitasking View, Picture-in-Picture, Do Not Disturb, and more. Or keep work out of sight when watching videos or playing games.</p>
+          <p>Stay productive and focused with Multitasking View, Picture-in-Picture, Do Not Disturb, and more. Or keep work out of mind when watching videos or playing games.</p>
```
comradekingu, line 125: This is a subtle but mindful accessibility tweak.

index.php
```diff
@@ -421,7 +421,7 @@
       <img alt="Download elementary OS icon" src="images/icons/apps/48/system-os-installer.svg">
       <div class="content-area">
         <p class="primary">Choose a Download</p>
-        <p>Download from a localized server or by magnet link. For help and more info, see the <a class="read-more" href="docs/installation" target="_blank" rel="noopener">installation guide</a></p>
```
comradekingu, line 424: Maybe some don't see what was done, but the blind man can spin.

index.php
cassidyjames: These multitasking features are literally visual features that keep things out of sight, though.

index.php
cassidyjames: Eh, this changes language away from what's used in the blog post and changes the meaning.

cassidyjames (changes requested):
These are all opinionated changes that don't really jive with what we wanted to say on the home page.

index.php
```diff
@@ -350,15 +350,15 @@
         <img class="app-display__image" src="images/screenshots/shortcut-overlay.png" width="885" height="656" alt="elementary OS Keyboard Shortcuts"/>
         <div class="app-display__description">
           <h1>User Friendly. And Keyboard Friendly.</h1>
-          <p>elementary OS is designed to be easy to understand and pick up as a new user. But that doesn’t mean it’s dumbed down; powerful, customizable keyboard shortcuts ensure you’ll stay productive whether it’s your first day or your thousandth.</p>
+          <p>elementary OS is designed to be easy to understand and pick up as a new user. But that doesn’t mean it’s dumbed down; powerful, customizable keyboard shortcuts ensure you stay productive on your first or thousandth day.</p>
```
comradekingu, line 353:
```suggestion
          <p>elementary OS is designed to be easy to understand and pick up as a new user. The one-armed man can fight, and we offer a honed axe. Powerful, customizable keyboard shortcuts within reach to stay productive on your first or thousandth day.</p>
```
Another nod to the sagas.

cassidyjames:
Blog post updated, by the way, thanks!

comradekingu:
@cassidyjames The critique I have seen more than once is that the message is too weak, which is a style not complemented by fluff and empty words.
It can't truly be "accessibility" if "sight" and "see" is invoked as actionable. Good angle though, props for doing something there.
I did add some comments, but I will gladly point out all the repetition, bad SEO, and lackluster concepts I got rid of.
Will explain my deeper thinking if that is needed too.
Interesting and captivating is better, emotions are better. That is kept and I tried to make the most of it.

index.php
comradekingu: @cassidyjames I couldn't get through the blogpost while grasping content, some of which I know already.
If the concepts aren't there or stick, the info is not accessible or even retained.
The blogpost is an OK meeting, not something I would want to share with people.
It needs some editing and bite.

index.php
comradekingu: @cassidyjames Right, but keeping things out of sight is easy. The whole experience of also having no notifications, getting the blue-light dimmed and so on is less stress. You can identify with getting work out of mind.
Getting work out of sight is just a cheap trick that neither grants peace nor allows rest.
If I can use my background in media in advertising for something that isn't cynical, this is it. It sells.

PR #2841, opened 2021-09-22
Installation: increase recommended disk size

### LINKED ISSUE #2840: Increase minimum recommended disk size
Apparently [OS 6 is not a great experience on 16 GB disks](https://old.reddit.com/r/elementaryos/comments/psw30s/slim_down_eos6_install/) like those in Chromebooks. We should probably bump the recommendation up to something like 64 GB to be more realistic. Especially because these aren't "hard minimums" but are just our lowest recommendations.

cassidyjames:
Fixes #2840

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -14,7 +14,7 @@ While we don't have a strict set of minimum system requirements, we recommend at
 
 * Recent Intel i3 or comparable dual-core 64-bit processor
 * 4 GB of system memory (RAM)
-* Solid state drive (SSD) with 64 GB of free space
+* Solid state drive (SSD) with at least 32 GB of free space
 * Internet access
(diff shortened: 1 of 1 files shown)


davidmhewitt:
I mean if the installed image is close to 15GB, I'd be tempted to day 32GB or 40GB is fine as it leaves you with >50% capacity to store your stuff and other apps.

Like I wouldn't be mad if I had a PBP with a 32GB eMMC or SD card and didn't think I could store full length movies on it.

lewisgoddard (approved):

cassidyjames:
@davidmhewitt yeah that's fair, I wouldn't be completely opposed to 32 GB. But even still, I wouldn't _recommend_ a 32 GB drive for someone to have a good time. So I'm kinda torn.

PR #2848, opened 2021-09-30
Update links to Slack

danrabbit:
Fix broken slack invite links

Update a link on Get Involved while we're here


btkostner:
Give me a bit, I think I solved this with cloudflare workers instead of having to update this code all the time.

PR #2859, opened 2021-10-05
Config: Bump to 6.0.2 ISO

cassidyjames:
Filesize is the same, but new filename, magnet, and SHA256. Wait for go-ahead from @lewisgoddard.


lewisgoddard (approved):
Distributed.

PR #2878, opened 2021-10-30
Small Update in footer

itsMeShon:
Fixes #

### Footer Update

The footer was sam as the color of the body before so i just changed the footer color to light gray color now it looks little bit diffrent from body color.

i think it may help you 😊


itsMeShon:
is there any problem with this or anything wrong?

PR #2923, opened 2022-01-02
Adjust copy. Issue #2894

### LINKED ISSUE #2894: Mission statement on website is (likely unintentionally) misleading
### What Happened?

Hello!

While this may seem nitpick initially, this minor change in wording can help resolve (what I'm sure is unintentional/accidentally) misleading information on the use of elementaryOS.

On the website, the opening mission statement is as follows: "The thoughtful, capable, and ethical _replacement_ for Windows and macOS". This is nice and all but it is, apparently, incorrect as confirmed by contributors working on elementaryOS.

According to both Files maintainer jeremypw, and Files contributor Marukesu, they have confirmed with me multiple times that elementaryOS should absolutely _not_ be considered a replacement to either Windows or Mac (or even other Linux distros) and instead is actually an "alternative". They had confirmed it was a mistake on my part for believing that it could be a replacement OS, of which I got that contradicting information directly from elementaryOS's website.

As such, simply changing the wording to "The thoughtful, capable, and ethical _alternative_ for Windows and macOS" would be more accurate and less misleading. I agree with them on their word choice, and it would have me less likely to believe I could use it as a replacement OS and would instead use it alongside another.

Just want to help pass that along so nobody else is mistaken by it.

Thanks!

### Steps to Reproduce

N/A

### Expected Behavior

N/A

### OS Version

6.x (Odin)

### Software Version

Latest release (I have run all updates)

### Log Output

_No response_

### Hardware Info

_No response_

StylusEater:
Fixes #2894

### Changes Summary

- Adjusted copy per suggestion.

### DIFF
--- a/_backend/preload.php
+++ b/_backend/preload.php
@@ -13,7 +13,7 @@ date_default_timezone_set('UTC');
 
 $sitewide['title'] = 'elementary';
 $sitewide['author'] = 'elementary, Inc.';
-$sitewide['description'] = 'The thoughtful, capable, and ethical replacement for Windows and macOS';
+$sitewide['description'] = 'The thoughtful, capable, and ethical alternative to Windows and macOS';
 $sitewide['image'] = 'https://elementary.io/images/preview.png';
 $sitewide['theme-color'] = '#3689e6';
 


StylusEater:
@cassidyjames what's the process to get eyes on PRs for new contributors? There seem to be a few PRs in the backlog.

cassidyjames:
@StylusEater opening a PR notifies those working on the related project, directly pinging contributors is not necessary. It's also not typically a good idea to invest time in a solution when the initial issue hasn't been agreed upon; in the case of #2894, the copy is intentional and the proposal here reverts to copy we intentionally replaced after talking with people out in the world about elementary OS. :)

PR #2943, opened 2022-01-20
Support: Update with GitHub Discussions links

### LINKED ISSUE #2630: Redesign the Support page for GitHub Discussions
## Problem

Currently the support page is built around guiding users towards StackExchange. It sounds like we want to use GitHub Discussions instead since that's one less login, it's easier to convert issues into discussions or reference discussions in issues etc.

## Proposal

We need to remove the references to StackExchange from the support page and at a minimum replace the existing links with GitHub Discussions links. We also need to remove the StackExchange link from the header

### LINKED ISSUE #2942: Update support page copy for clarity
Update website copy to clarify that it is community support, not employee/full-time/etc

cassidyjames:
- Fixes #2630
- [x] Removes StackExchange link from nav
- Fixes #2942
- Adds a section explicitly to mention Payment & Download Support
- Renames "Epiphany" to "Web" to match the OS
- Adds "Calculator" since it was missing
- Fixes alphabetization
- Fixes whitespace

support-gh

Other apps (and desktop components) are missing, but I figured we'd get this change to existing links in first before adding more apps/components (especially since that will require pulling in more icons and whatnot).

View without whitespace changes.

### DIFF
--- a/_templates/header.php
+++ b/_templates/header.php
@@ -127,7 +127,6 @@ $l10n->begin_html_translation();
                     <li><a href="https://www.facebook.com/elementaryos" target="_blank" rel="noopener" data-l10n-off title="Facebook"><i class="fab fa-facebook-f"></i></a></li>
                     <li><a href="https://mastodon.social/@elementary" target="_blank" rel="noopener me" data-l10n-off title="Mastodon"><i class="fab fa-mastodon"></i></a></li>
                     <li><a href="https://www.reddit.com/r/elementaryos" target="_blank" rel="noopener" data-l10n-off title="Reddit"><i class="fab fa-reddit"></i></a></li>
-                    <li><a href="https://elementaryos.stackexchange.com" target="_blank" rel="noopener" data-l10n-off title="Stack Exchange"><i class="fab fa-stack-exchange"></i></a></li>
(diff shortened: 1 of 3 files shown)


_templates/header.php
```diff
@@ -127,7 +127,6 @@
                     <li><a href="https://www.facebook.com/elementaryos" target="_blank" rel="noopener" data-l10n-off title="Facebook"><i class="fab fa-facebook-f"></i></a></li>
                     <li><a href="https://mastodon.social/@elementary" target="_blank" rel="noopener me" data-l10n-off title="Mastodon"><i class="fab fa-mastodon"></i></a></li>
                     <li><a href="https://www.reddit.com/r/elementaryos" target="_blank" rel="noopener" data-l10n-off title="Reddit"><i class="fab fa-reddit"></i></a></li>
-                    <li><a href="https://elementaryos.stackexchange.com" target="_blank" rel="noopener" data-l10n-off title="Stack Exchange"><i class="fab fa-stack-exchange"></i></a></li>
```
btkostner, line 130: This change will need to be mirrored to any other sites that use this

cassidyjames: Issues filed and linked below; feel free to file any others I missed!

PR #3067, opened 2022-09-06
Fixed a few alignment issues in the "get involved" section

jesuslencina:
I noticed that the alignment of a few elements in this section looked a bit off (at least to me), so I attempted to make them look better. 

Here's an example:

Before:
image

After:
image

This is my first pull request ever. Please let me know if there's anything else to check.

Thanks in advance

### DIFF
--- a/_styles/get-involved.css
+++ b/_styles/get-involved.css
@@ -15,6 +15,10 @@ section {
   text-align: left;
 }
 
+.grid .third p{
+  text-align: center;
+}
+
 @media only screen and (min-width: 640px) and (max-width: 900px) {
   .grid .third {
     width: 66%;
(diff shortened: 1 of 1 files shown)


dar5hak:
Congrats on your first PR, Jesús!

I see where you are coming from, but while centred text might look more symmetrical, it also affects readability.

Unlike with left-aligned text, a reader does not have a consistent starting point to return to after finishing a line. They have to make a semi-conscious visual effort to find the starting point.

In this case, it is just a few lines, so this is debatable. But I'd rather left-align the heading and the button with the body text instead of the other way around.

(This is just my opinion; please don't change anything unless a maintainer says so.)

PR #3077, opened 2022-09-29
Added hovering effect color to youtube icon and mastodon icon present…

shreyas2711:
… on right navbar

Fixes #

### Changes Summary
Nothing much added just added hovering effect(when hovered color changes) effect to Youtube icon and mastodon icon since other icons also had them so felt it would look better if I added for the other two. 
- 
- 
- 

This pull request [ is] ready for review.


RMcNeely:
Hey @shreyas2711 thanks for the PR! Changes look mostly good but there's some extra whitespace that's failing linting. If you can remove that we should be able to get this change merged in.

danirabbit (approved):
Went ahead and fixed the linting errors. Thanks for your contribution!

PR #3096, opened 2022-10-29
Fix translations build

lewisgoddard:
To avoid the NPM problem with eslint, just use `--production`


davidmhewitt:
I believe this would also be solved with #3101

PR #3098, opened 2022-11-05
Bump eslint development dep to latest

davidmhewitt:
Fixes an issue trying to get a development environment setup.

Shouldn't affect anything since this is just a dev dependency, but I've checked that this still works on Node 12.


danirabbit (approved):
CI says :rocket:

PR #3100, opened 2022-11-06
Migrate to Stripe Checkout for OS payments

### LINKED ISSUE #998: Create customers with Stripe
Stripe lets you create a customer as well as a transaction when someone is checking out. If possible, we should be creating a customer or potentially bypassing the payment if it is a return Stripe customer.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/30174408-create-customers-with-stripe?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #3097: Migrate to the latest stripe checkout
### Problem

We're still using the legacy Stripe checkout which presents as a modal on our website and doesn't support newer secure payment methods

### Proposal

Migrate to the new stripe checkout which supports the latest payment methods: https://stripe.com/docs/payments/checkout/migration

### Prior Art (Optional)

_No response_

davidmhewitt:
Fixes #3097 
Fixes #998 

Test site (with Stripe in test mode) at https://beta.elementary.io/stripe-checkout/


RMcNeely:
Well now I feel a little bad telling Derek in #3097 that I wanted to see this as a modal still but the new page looks good! How much of the page is under our control for the checkout?

davidmhewitt:
It's not hugely customizable, mostly just colours, logos and fonts (they do have Inter as an option though! 😆 ):
https://stripe.com/docs/payments/checkout/customization?locale=en-GB

If we went the modal route, it's a lot more customizable, but a lot more complex in terms of receiving the payment callback, and would also require someone with some design chops to fit it into the existing page. It would require somewhat of a redesign because I think the modal would need to accommodate a bigger form with more variable options (Google/Apple Pay etc)

RMcNeely:
Mostly I told him that because I didn't feel comfortable yet making the design decision to navigate to a separate checkout page but it seems that this is what Stripe would recommend.

danirabbit:
Just gonna throw my +1 down here for using the separate checkout page as recommended by Stripe

danirabbit (approved):

PR #2924, opened 2022-01-02
Move to Mailchimp. Issue #2817

StylusEater:
Fixes #2817

### Changes Summary

- Adjusted composer.json for new requirement.
- Adjusted composer.json to remove mandrill.
- Updated config.example.php with new key for mailchimp.
- Adjusted to new mailchimp API.

This pull request is ready for review.

PR #3106, opened 2022-11-07
Switch away from deprecated mandrill API

### LINKED ISSUE #2817: mandrill/mandrill is deprecated and should be replaced with mailchimp/transactional
#### mandrill/mandrill
This package is abandoned and no longer maintained. The author suggests using the mailchimp/transactional package instead.

- https://packagist.org/packages/mandrill/mandrill
- https://packagist.org/packages/mailchimp/transactional

davidmhewitt:
Fixes #2817 
Closes #2924 

Tested and working, still sends emails. The changes to the `$res` loop/error handling have been tested and verified correct.


_backend/email/os-payment.php
```diff
@@ -8,17 +8,19 @@
 require_once __DIR__ . '/../../_backend/preload.php';
 require_once __DIR__ . '/../../_backend/config.loader.php';
 
-$mandrill = new Mandrill($config['mandrill_key']);
+$mailchimp = new MailchimpTransactional\ApiClient();
+$mailchimp->setApiKey($config['mandrill_key']);
```
RMcNeely, line 12: Is there a reason why you did not create a new key here or just rename the existing key? Its might be hard to remember in the future why the MailChimp API is consuming the Mandril key

_backend/email/os-payment.php
```diff
@@ -81,11 +83,15 @@ function email_os_payment (\Stripe\PaymentIntent $intent) {
...
+    $res = $mailchimp->messages->sendTemplate([
+        "template_name" => "os-purchase",
+        "template_content" => $req,
+        "message" => $message
+    ]);
 
     foreach ($res as $mail) {
-        if (isset($mail['reject_reason']) && $mail['reject_reason'] !== '') {
-            throw new Exception($mail['reject_reason']);
+        if (isset($mail->reject_reason) && $mail->reject_reason !== '') {
```
RMcNeely, line 93: I don't really know PHP outside of trying to use it on this repo but is there a reason for changing the notation here? I understand that they are both forms of object notation.

RMcNeely (commented):
Seems straightforward enough, right from the README in Packagist, two small questions.

lewisgoddard (approved):

_backend/email/os-payment.php
lewisgoddard: PHP has a history of objects and arrays (they are kind of the same thing in the language), but typically you would use `$object->function()` and `$object->key` but `$array['key']`. This is probably just to be consistent so you know what you're working with.

lewisgoddard:
This is going to require a manual build on the server, maybe update a config not kept in this repo when it goes live.

PR #3112, opened 2022-11-07
Homepage: remove payment-trust section

danirabbit:
Removes this:

Screenshot from 2022-11-07 15 23 02

Since the new checkout already has this information:

Screenshot from 2022-11-07 15 23 24


davidmhewitt (changes requested):
You should remove the presumably now unused icons from gulpfile.babel.js

danirabbit:
@davidmhewitt nice catch, thanks!

davidmhewitt (approved):

PR #3113, opened 2022-11-08
CI: Update translations dependencies

davidmhewitt:
Translation update action has been broken for a while for various reasons.

Latest failures seem to be because of something an action is doing that GitHub considers a security risk and has disabled. So let's move to the latest versions of these actions and see if we get any further.


danirabbit (approved):
Let's go!

PR #3118, opened 2022-11-09
Bump PHP/Node/Actions versions in workflows

davidmhewitt:
We're running these Node and PHP versions on the server now, so lets update CI to match


danirabbit (approved):
CI approves :)

PR #2305, opened 2019-12-13
Adds chunk hashing for cache busting to Webpack

RMcNeely:
Fixes #2093 

### Changes Summary

- Adds cache busting for JS built in Webpack
- Adds cache busting for CSS via adding a time string to CSS files
- Updates Webpack to v4
  - Uses Terser Webpack Plugin as new default method for minification

This pull request [ is ] ready for review.

PR #3126, opened 2022-11-15
Add cache busting to build process

### LINKED ISSUE #2093: Browser caching on Chrome breaks website
Looks like Chrome is caching the CSS/JS and it is not being updated when the website is updated. I would recommend cache fingerprinting  (use `main.0fklh924sd834.css` instead of `main.css`).

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/65035183-browser-caching-on-chrome-breaks-website?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F10548672&utm_medium=issues&utm_source=github).
</bountysource-plugin>

davidmhewitt:
Fixes #2093 
Closes #2305 

Adds a hash to our generated CSS and JS filenames that allows us to bust any CDN and Browser caches on updates meaning there will be less brokenness when we update the site.

Also adds back some CSS minification via `cssnano` that I seem to have inadvertently broken in a previous commit.

### DIFF
--- a/gulpfile.mjs
+++ b/gulpfile.mjs
@@ -6,12 +6,15 @@
 import gulp from 'gulp'
 import cache from 'gulp-cached'
 import changed from 'gulp-changed'
+import rev from 'gulp-rev'
+import rename from 'gulp-rename'
 import { spawn } from 'child_process'
 
 import imagemin from 'gulp-imagemin'
 
 import postcss from 'gulp-postcss'
 import postcssPresetEnv from 'postcss-preset-env'
+import cssnano from 'cssnano'
 
 import webpack from 'webpack'
 import webpackConfig from './webpack.config.babel.js'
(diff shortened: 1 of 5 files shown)


gulpfile.mjs
```diff
@@ -184,8 +187,15 @@ gulp.task('styles', () => {
     return gulp.src(src, { base })
     .pipe(changed(dest))
     .pipe(postcss([
-        postcssPresetEnv()
+        postcssPresetEnv(),
+        cssnano()
     ]))
+    .pipe(rev())
+    .pipe(gulp.dest(dest))
+    .pipe(rename({
```
RMcNeely, line 195: Is this rename just for the manifest file?

RMcNeely: I suppose I have two questions. First, does `rev` need to happen after `postcss`? In the README instructions for `gulp-rev` they show it as running directly after the src is declared. Second is line 194 actually needed? You typically only want to have one destination at the very end of your stream. It almost looks like that is why you are needing to rename the file.

RMcNeely: Ok so rereading the docs on `gulp-rev` looks like they do not support streaming which means you do have to move the files then create the manifest. That said we can pass in the base path of `dest` as an option to the rev.manifest function then we wouldn't need the gulp-rename dependency.

If you would like I can push a commit with these changes.

davidmhewitt: Thanks for looking into this.

I had some real issues getting the content of the manifest to be `styles/whatever.css` instead of just `whatever.css`, and this rename was the only way I could get it to work.

Absolutely feel free to improve it if you can though.

Thank you!

RMcNeely (approved):

PR #3160, opened 2023-01-23
Added Dockerfile with nginx and php

Claudio-code:
### Changes Summary

-  Added docker configuration to start project in container
- Added nginx in container


dev/Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM php:7.4-fpm
```
davidmhewitt, line 1: Can we use PHP 8.1 as this is what we're running on the web server?

dev/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+
+RUN curl -sL https://deb.nodesource.com/setup_16.x | bash -
```
davidmhewitt, line 3: Can we use node 18 as this is what we're running on the web server?

.github/CONTRIBUTING.md
```diff
@@ -2,113 +2,14 @@
 
 ## Web Server
 
-### Simple PHP Server
+### Docker Web Server with Nginx and PHP
 
 For a simple local development environment running on PHP, you will need:
-
```
davidmhewitt, line 8: I think it's worth keeping the manual installation instructions around in case people want to do a setup without docker. Can we move them to a "manual setup" section, and have the Docker instructions under a new "recommended setup" section?

davidmhewitt: That said, after reviewing the instructions here, it looks like they need updating a bit :sweat_smile:

davidmhewitt (changes requested):
Hey, thanks for working on this!

I haven't tested this or looked at it in any detail yet, but I have a few things that I think are worth changing if possible.

dev/Dockerfile
Claudio-code: Yes we can, I uploaded the version

dev/Dockerfile
Claudio-code: Yes we can, I uploaded the version

.github/CONTRIBUTING.md
Claudio-code: Cool I added these two sessions and update the recommended versions of php and nodejs

dev/Dockerfile
```diff
@@ -0,0 +1,33 @@
+FROM php:8.1-fpm
+
+RUN curl -sL https://deb.nodesource.com/setup_18.x | bash -
+
+RUN apt-get update && apt-get upgrade -y && apt-get install -y git libicu-dev nodejs \
+    nginx libpng-dev libzip-dev libwebp-dev unzip zip libxml2-dev libjpeg62-turbo-dev zlib1g-dev \
+    libfreetype6 libfreetype6-dev
+
+RUN docker-php-ext-install opcache \
```
davidmhewitt, line 9: Do we really need the opcache stuff? It seems like added complication and build time in this Dockerfile.

I understand it can be useful for performance, but it seems pretty performant without this to me, and this is just for a development environment anyway.

dev/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+    && docker-php-ext-configure gd --with-webp=/usr/include/webp --with-jpeg=/usr/include --with-freetype=/usr/include/freetype2/ \
```
davidmhewitt, line 10: Have you found something that needs the GD extension?

davidmhewitt (changes requested):
Hey, thanks for the quick changes.

I mostly just have some questions about whether extensions are needed this time. Maybe you've found a reason for these extensions that I haven't :laughing:

dev/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+    && docker-php-ext-install -j$(nproc) gd \
+    && docker-php-ext-install -j$(nproc) zip \
+    && docker-php-ext-configure intl \
+    && docker-php-ext-install intl
+
+RUN curl -sS https://getcomposer.org/installer | php -- --install-dir=/usr/local/bin --filename=composer
```
RMcNeely, line 16: Is there not a Composer image that we could pull from and use as the base?

dev/entrypoint.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+cd _backend
+composer up
+cd ../
+npm i -g npm
```
RMcNeely, line 6: I don't think you should do a global install here of node modules. You should be fine doing a plain `npm install`

.github/CONTRIBUTING.md
```diff
@@ -1,6 +1,20 @@
 # Contributing
 
-## Web Server
+## Recommended setup
+
+### Docker Web Server with Nginx and PHP
```
RMcNeely, line 5:
```suggestion
### Docker Web Server with Nginx, PHP and Node
```

dev/Dockerfile
Claudio-code: I tested it now and it's really not necessary, it was removed

dev/Dockerfile
Claudio-code: I tested it now and it's really not necessary, it was removed

dev/Dockerfile
Claudio-code: you're right there's a way to extract composer from an image i had never done that with composer, really cool i swore it would break but it worked thanks for the tip

dev/entrypoint.sh
Claudio-code: didn't really need that global install, it's already been removed

RMcNeely:
Right now I can build the container but it looks like there's an error in your branch where `npm build` is not working correctly. I'll check and see if this error is happening in elementary/master.

For what it's worth I do see permissions errors with the way that NPM is running but my understanding is that this is how the server has node installed as well. @davidmhewitt can you confirm that?

RMcNeely:
ok well now I can't reproduce the issue that I was having in the first place! I've been testing this out on MacOS and aside from some strange errors I got the first time spinning this up its been working pretty well for me. I've deleted and recreated the container a few times so I'm not really sure what my problem was. 

If David says the configs are a close match to whats on the server then this is pretty good.

Claudio-code:
In my tests I didn't find this problem but I can create a virtual machine with mac os to try to reproduce this error too, but running on linux fedora and ubuntu systems didn't give a problem.

davidmhewitt (approved):
This works well and is representative of the setup on the production web servers.

Thanks for your work on this @Claudio-code ! It should hopefully be easier for more people to contribute now!

PR #3163, opened 2023-01-24
Dependency cleanup

davidmhewitt:
### Changes Summary

- Add `ext-intl` and `ext-json` (which the website code depends on for translations) to composer so that it can warn us if the platform we're trying to install on doesn't have the right extensions.
- Update the Node and PHP versions in the contributing guide
- Remove some no longer used dependencies from the contributing guide. These seem to have been used in the past for average payments and charts
- Remove some dead "chart" code that used to generate burndown charts from launchpad issues on the get involved page


.github/CONTRIBUTING.md
```diff
@@ -12,22 +12,20 @@ For a simple local development environment running on PHP, you will need:
...
-  * `php-sqlite3`
-  * `php-xml`
   * `composer`
 * [Node.js](https://nodejs.org/) and `npm`
   * packages installed with `npm ci`
 
 #### GNU/Linux Based Operating Systems
 
-First, install Node.js (12.x recommended) from [Nodesource](https://github.com/nodesource/distributions/blob/master/README.md#debian-and-ubuntu-based-distributions). 
+First, install Node.js (18.x recommended) from [Nodesource](https://github.com/nodesource/distributions/blob/master/README.md#debian-and-ubuntu-based-distributions).
```
RMcNeely, line 21: 🎉

PR #3171, opened 2023-02-07
Support: update icons, add other Q&A link

danirabbit:
* Update icons on the support page
* Add more components
* Fix broken calculator link
* Make sure we link directly to Q&A
* add a link for org-wide Q&A so we can catch discussions that aren't easy to categorize or aren't represented by the apps listed
* Update Get Involved links

Screenshot from 2023-02-08 08 58 30


get-involved.php
```diff
@@ -109,9 +109,9 @@
         </div>
         <div class="half">
             <h3>Question &amp; Answer</h3>
-            <p>Stack Exchange is a Q&amp;A website built around finding the best answers to common questions. Anyone can create an account to start asking and answering.</p>
-            <a class="button flat" href="https://elementaryos.stackexchange.com/questions?sort=newest">New Questions</a>
-            <a class="button flat suggested-action" href="https://elementaryos.stackexchange.com/unanswered">Unanswered Questions</a>
+            <p>We use GitHub Discussions for community Q&A. Anyone can create an account to start asking and answering.</p>
+            <a class="button flat" href="https://github.com/orgs/elementary/discussions/categories/q-a?discussions_q=category:Q&A+sort:date_created">New Questions</a>
```
RMcNeely, line 113: I'm testing both of these links and it looks like they are not filtering something correctly when they open discussions. For each of these, I see a screen that says there are no discussions. I know that there ARE discussions for elementary so I'm not sure if thats what we want.

RMcNeely:
One other very tiny thing, in your screenshot you have underlined the text on the support page for "Other Questions & Help" but it is not underlined in code.

get-involved.php
danirabbit: Ah I figured it out. The ampersand ha

danirabbit:
> One other very tiny thing, in your screenshot you have underlined the text on the support page for "Other Questions & Help" but it is not underlined in code.

Yeah that's not intentional on my end. I'm guessing it's a keyboard focus style? I didn't touch anything with styles there

RMcNeely:
> 

Cool, I think it looks much better the way it is now. Just checking!

RMcNeely (approved):

PR #3176, opened 2023-02-16
stylelint: Use recommended rules

davidmhewitt:
The stylelint configuration file we were using was having very little effect since most of the options in there have long since been deprecated.

I propose we switch to extending a recommended configuration so it will stay up to date with best practices.

As you can see from the changes I've had to make to the CSS, the new config caught some minor issues.

I've disabled the `no-descending-specificity` rule because it had 100+ false positives that I couldn't be bothered to sort through at this stage.

This also has an added bonus of formatting the CI output for GitHub so you get nice inline lint issues in PRs.


_styles/docs.css
```diff
@@ -401,10 +402,6 @@ div.icons-container > span:nth-child(6) { width: 128px; }
   }
 }
 
-div.icons-container.small-screen {
-  display: none;
```
davidmhewitt, line 405: This has been removed because there's a duplicate rule further up on line 374

.github/workflows/ci.yml
```diff
@@ -36,7 +36,7 @@ jobs:
         run: npm ci
 
       - name: Lint
-        run: ./_tests/css.sh
+        run: npx stylelint "_styles/**/*.css" -f github
```
RMcNeely, line 39: What is the reason for having the formatter specified here? Does it not work if you just allow the default setting?

RMcNeely:
How do you feel about npm script instead of a folder full of shell files?

.github/workflows/ci.yml
davidmhewitt: This is the console output format of the lint errors.

This one is machine readable by GitHub actions and allows GitHub to show errors inline against specific lines during the PR process.

We wouldn't want to make this the default as it's much better to have the nice pretty coloured human readable output for local development.

davidmhewitt:
I think the shell scripts are all just one liners that you can run with `npm lint` anyway, which seems a little strange.

I'd be up for cleaning it up and unifying it a little if you wanted to open another PR!

RMcNeely (approved):

PR #3177, opened 2023-02-16
Use cache-busted versions of scripts on docs pages

davidmhewitt:
In #3126 we added cache busting by appending checksums to our script and css filenames.

This works for most pages, but I've noticed some of the markdown pages (i.e. the installation guide) include scripts inside the markdown, so we need to add some code to replace these filenames from the manifest.

Nothing is broken on the server here currently as the only javascript file without the checksum is still present on the server, but this may break in the future if we don't do this.

To test:

* Run site locally, go to http://localhost:8000/docs/installation 
* See that the `installation.js` file cannot be loaded
* Run this branch, see that the correct script filename is substituted and loaded


RMcNeely:
How is the Markdown getting rendered here? It seems like an antipattern to embed the script tag inside of the Markdown file.

davidmhewitt:
It's this PHP file parsing the markdown and rendering it out as HTML.

The markdown files sometimes have inline script tags, see:
https://github.com/elementary/website/blob/2b3d591e35692a255e61461000343a2b45e91985/docs/installation.md?plain=1#L216

Obviously that script file is specific to the installation page, so it wouldn't be relevant to include it on other markdown pages.

Open to better ways of doing this, as this does feel a bit hacky.

RMcNeely:
I guess what I'm wondering is how the PHP backend is rendering that specific page as Markdown. The issue is that the files where this happens would all need their own individual templating or something like that to load any scripts that we want embedded on the page. Then again you are already creating custom logic for each script so whats a template to render the correct Markdown at that point.

RMcNeely (approved):
This appears to be a weird one off situation where we've embedded a ton of logic inside of the markdown file. It's probably worth revisiting this page as a whole since but not worrying about this specific code funk at the moment.

Issue #3179, opened 2023-02-17
Allow verification of the OS download via the website itself


4jNsY6fCVqZv:
### Expected behavior

I wonder if it would help more non-technical users to actually verify their downloads instead of skipping this step due to complexity (using terminal application).

There is a beautiful working example at the Tails project website.

https://tails.boum.org/install/linux/index.en.html
(see here step 3 "Verify your download")

### Actual behavior

At the moment, the installation guide still advises to check through the terminal.
look at https://elementary.io/docs/installation/#verify-your-download

Do you have any feelings and thoughts about this?

davidmhewitt:
That's very cool, I like that idea a lot! 

It's a little slow since it's running in the browser under JavaScript, but definitely way less of a technical barrier.

I guess if someone was to propose a design of the most user friendly way to fit this into the elementary download flow, I'd be willing to have a go at implementing it.

JBanks:
From https://github.com/elementary/website/pull/3189#issuecomment-1460611338
> there's no way to do chunking on the hash creation

This is correct when referring to the SubtleCrypto API.  Using a different JavaScript library would allow you to perform "updates" using the chunks as long as they are in sequential order.  It could also bring down the space complexity since it looks like you're currently loading the entire image into memory before validating the hash.

From https://github.com/elementary/website/issues/3179#issuecomment-1434969899
> It's a little slow since it's running in the browser under JavaScript

Would it be worth discussing a move to WASM for faster results?  I think JavaScript is the winner here since it lessens the number of languages in the code-base, but I think it's worth discussing since this could bring the verification time down significantly.

RMcNeely:
Personally, I never like to add a library that would need to be maintained when you have a capable alternative in Web API's. I don't think that configuring WASM for just this is going to make an incredible difference. 

For the benchmarking when you say `shasum` are you just running that locally on your computer? And is `tails` the time it took the tails to check an example file from the first post?

JBanks:
@RMcNeely, I guess I should have benchmarked with the elementary file originally, because I just got this error using your method:
![image](https://user-images.githubusercontent.com/453768/229167250-bc772809-117d-4633-81a1-209140e1d694.png)
Elementary is just under 3GB.  Here's the updated benchmark results with the elementary image:

| method        | time (s) |
|---------------|----------|
| RMcNeely      | Error    |
| shasum -a 256 | 14.30     |
| tails         | 104.119   |

[...]

PR #3189, opened 2023-03-01
Simple ISO verification hashing done on the client

RMcNeely:
Fixes #

Initial pass at #3179 

### Changes Summary

- Adds a page to verify the checksum of the download

This pull request is a draft that needs UI review

### DIFF
--- a/download.php
+++ b/download.php
@@ -1,0 +1,49 @@
+<?php
+    require_once __DIR__.'/_backend/preload.php';
+
+    $page['name'] = 'Verify Download';
+    $page['title'] = 'Verify your OS Download';
+
+    include $template['header'];
+?>
+
+<div style='margin: 16rem 8rem;'>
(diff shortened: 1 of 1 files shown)


4jNsY6fCVqZv:
@RMcNeely I am curious. Could you imagine showing a visual glimpse of the actual state of your POC?

RMcNeely:
I can add _something_ to show progress but I'm not sure that the way it is written I can easily add a progress bar given that the APIs are async. I would need to rewrite the file read operation to use chunks but to my knowledge, there's no way to do chunking on the hash creation so at best you would have a progress bar for only one part of the workflow.

I think it would be simpler to add a message that changes with the different steps of the verification process. Something like: 
1. Reading ISO file
2. Creating Hash
3. Other? Stringifiying the hash and comparing it to an expected value would both be very quick operations. I'm not sure thats worth displaying something for.

PR #3359, opened 2023-09-26
Increased navbar icons size and decreased gap

### LINKED ISSUE #3319: Icons need a minor updates
### Problem

- Icons at the top of the navbar.

### Proposal

- The icons at the top of the navbar are small and too hard to see, maybe if you can increase their size a little bit would be better. (1.3em) does the job I've tested it on Chrome dev tools.

- If possible also decrease the gap between icons.

### Prior Art (Optional)

_No response_

bhavikrajpal49:
Fixes #3319

### Changes Summary

Changes made in main.css

nav a:visited {
  color: inherit;
  display: inline-block;
  fill: inherit;
  margin: 0;
  padding: 0 10px; /* Decreased the padding to reduce the gap between icons */
  font-size: 1.3em; /* Increase the font size to make icons larger */
}

### DIFF
--- a/_styles/main.css
+++ b/_styles/main.css
@@ -324,7 +324,8 @@ nav a:visited {
   display: inline-block;
   fill: inherit;
   margin: 0;
-  padding: 0 12px;
+  padding: 0 10px;
+  font-size: 1.3em;
 }
 
 nav a:focus,


lewisgoddard (changes requested):
This also changes the size of the navigation links that are text-based on the left. This issue isn't confirmed, so I wouldn't spend too much time on this.

PR #3360, opened 2023-09-29
Remove custom mailchimp emails

davidmhewitt:
Will need the receipts turning on in the Stripe dashboard once we've deployed this.


danirabbit (approved):
LGTM!

PR #3367, opened 2023-09-30
Bump stripe API version

davidmhewitt:
Makes the website explicitly use the current latest version of the Stripe API so we'll be able to see in the dashboard if there's still anything using an older version by default.

This no longer sends a re-download link in the Stripe receipt email as the API changes mean we don't get a PaymentIntent ID up front, so we can't use that to uniquely ID the payment. I've not removed the code to honor old re-download links, though, this will become defunct in the future.


danirabbit (approved):
Not exactly my wheelhouse but nothing seems glaringly wrong to me :sweat_smile:

Issue #3591, opened 2024-07-21
Accessibility issues on the website


NickColley:
## High level

Off the back of reading the [It's Disability Pride Month! Let's Get Accessible ](https://blog.elementary.io/updates-for-july-2024/) blog post I noticed that the navigation used on the website had poor colour contrast, I then ran some automated testing and found more issues.

My recommendation is to adopt an end-to-end accessibility policy where access is prioritised from people hearing about Elementary on the web all the way to them downloading it and using it.

While I understand this is a project primarily ran by volunteers more care can be taken to triaging and prioritising access barriers raised by real users and to track them in the overall project: https://github.com/orgs/elementary/projects/111.
Issues like https://github.com/elementary/website/issues/3378 should be considered high priority bugs as the user is describing the download functionality to be completely broken.

## Website accessibility

Through some very quick reviewing examples of issues look like:

### Colours used are hard to read

In this example one of the most important elements, the download button does not have sufficient contrast between the text colour and the background, this could mean that some users with visual impairments struggle to read what the button says.

![Screenshot of the download button on the homepage with a failing colour contrast check at 3.28, with white on light blue](https://github.com/user-attachments/assets/6a524433-2b6d-44c2-a64c-fe3f6df767e8)

This sort of issue can be seen elsewhere for example the blog post page has particularly low contrast (2.10) for the main navigation

![Blog post navigation links with white on very light blue](https://github.com/user-attachments/assets/803476d4-bd88-409f-a1ab-8c3dde968a0e)

Potential solution here is to increase the contrast so that these meet 4.5 contrast ratio at minimum.

### Focus indicators are hidden
If you tab through the homepage some elements have focus styles and others have no styles at all which makes it very difficult to know where you are on the page if you're using a keyboard to navigate, this could mean some users with physical impairments will find it hard to get around the website.

Potential solution here is to add clear focus styles for all tabbable elements.

### Images with no alternative text
There are many images that do not have alternative text, this could mean some users of screen readers, for example blind users, will not know what the images mean.

Potential solution here is to add alternative text to all non-presentational images.

These examples are non-exhaustive I have spent very little time putting them together, to properly understand the amount of barriers that exist I propose the following next steps:

## Next steps
1. [ ] Go through existing GitHub issues and triage and label any access barriers already raised
1. [ ] Do a review of the most important web journeys
    - [ ] Create a list of the web pages most used and critical for using Elementary e.g. downloading the distro, asking for help.
[...]

PR #3592, opened 2024-07-21
Make focus indicators consistent

NickColley:
Focus indicators are important for keyboard users to navigate around the website, without them it is very difficult or impossible to know where they are on the page.

This change removes the styles that prevent focus indicators being shown and bring them inline with the rest of the website e.g. blog page.

Ideally we've have some sort of design system where we have considered focus styles that work at a component level but this is the minimum to make things consistent and remove a big barrier.

This resolves issues related to [Focus Visible (Level AA)](https://www.w3.org/WAI/WCAG22/Understanding/focus-visible.html), [Focus Appearance (Level AAA)](https://www.w3.org/WAI/WCAG22/Understanding/focus-appearance), [Non-text Contrast (Level AA)](https://www.w3.org/WAI/WCAG22/Understanding/non-text-contrast).
Relevant failure example which mirrors the issue we see in our project:
[Failure of Success Criterion 1.4.11, 2.4.7 and 2.4.13 due to styling element outlines and borders in a way that removes or renders non-visible the visual focus indicator](https://www.w3.org/WAI/WCAG22/Techniques/failures/F78)

As part of #3591

### Before
Donate button being focused, very slight colour change from light grey to darker grey

### After
Focus indicators will be the default for the operating system and browser, here's how it looks on Ubuntu Firefox. This is consistent with the majority of the website.

Donate button being focused, default focus indicators shown, in this case Ubuntu Firefox shoes a clear orange focus

Note that for some browsers focus styles will only appear when using the keyboard when using the default focus styles, following this up with custom focus indicators in the future may be an opportunity to add them to people using pointer devices as well e.g. a mouse.


RMcNeely:
Looks good to me but its already out of date (sorry!) If you can get this updated then I think it will be good. I did some greepping for other places that had a value set for `outline` but couldn't find any so I think we're good.

NickColley:
@RMcNeely no worries, rebased :)

zeebok (approved):

RMcNeely (approved):

PR #3593, opened 2024-07-21
Replace leanmodal2 with native dialog element

NickColley:
Adds a lot of good features that make it way more accessible like preventing outside elements from being focused, which can be difficult to do otherwise.

Removes dependency on leanmodel2 so quicker page loads.

TODO before ready:
- Fallback gracefully in browsers that dont support it
- Test in real screen readers

Fixes #

### Changes Summary

- 
- 
- 


index.php
```diff
@@ -467,21 +467,23 @@
 
     <span id="translate-download" style="display:none;" hidden>Download elementary OS</span>
     <span id="translate-purchase" style="display:none;" hidden>Purchase elementary OS</span>
-    <div id="download-modal" class="dialog modal">
-      <img alt="Download elementary OS icon" src="images/icons/apps/48/system-os-installer.svg">
+
+    <dialog id="download-modal" class="dialog modal" aria-labelledby="download-modal-title">
+      <!-- <img alt="Download elementary OS icon" src="images/icons/apps/48/system-os-installer.svg"> -->
+      <img src="https://elementary.io/images/icons/apps/48/system-os-installer.svg" alt="">
```
NickColley, line 473: Images were not loading locally so I hardcoded this but needs to be changed back.

index.php
```diff
@@ -467,21 +467,23 @@
 
     <span id="translate-download" style="display:none;" hidden>Download elementary OS</span>
     <span id="translate-purchase" style="display:none;" hidden>Purchase elementary OS</span>
-    <div id="download-modal" class="dialog modal">
-      <img alt="Download elementary OS icon" src="images/icons/apps/48/system-os-installer.svg">
+
+    <dialog id="download-modal" class="dialog modal" aria-labelledby="download-modal-title">
```
NickColley, line 471: Using aria-labelledby to tie the heading to the dialog itself so it's clear what it's for, need to test screen reader behaviour, I think we may need to focus the dialog when it opens...

index.php
```diff
@@ -467,21 +467,21 @@
 
     <span id="translate-download" style="display:none;" hidden>Download elementary OS</span>
     <span id="translate-purchase" style="display:none;" hidden>Purchase elementary OS</span>
-    <div id="download-modal" class="dialog modal">
-      <img alt="Download elementary OS icon" src="images/icons/apps/48/system-os-installer.svg">
+    <dialog id="download-modal" class="dialog" aria-labelledby="download-modal-title">
+      <img src="images/icons/apps/48/system-os-installer.svg" alt=""/>
```
NickColley, line 471: Set the alt to nothing to mark the image as presentational as it is just for show.

index.php
```diff
@@ -467,21 +467,21 @@
 
     <span id="translate-download" style="display:none;" hidden>Download elementary OS</span>
     <span id="translate-purchase" style="display:none;" hidden>Purchase elementary OS</span>
-    <div id="download-modal" class="dialog modal">
-      <img alt="Download elementary OS icon" src="images/icons/apps/48/system-os-installer.svg">
+    <dialog id="download-modal" class="dialog" aria-labelledby="download-modal-title">
```
NickColley, line 470: We label the dialog based on the heading so that the dialog has a clear name "Choose a Download modal dialog" etc

PR #3602, opened 2024-07-25
Remove legacy warnings that are no longer correct or in use.

### LINKED ISSUE #3601: Legacy warning display on local development
The banner that shows when a browser is attempting to visit the site using HTTP shows when developing locally.

https://github.com/elementary/website/blob/master/_templates/legacy.php#L15

I thought maybe this was because the PHPENV variable that's checked for is not set to development here:

https://github.com/elementary/website/blob/master/dev/nginx.conf#L27

But when changing it and restarting things it didnt seem to change to development, might be something I did incorrectly though.

Main impact of this is it makes developing locally a bit annoying seeing the banner show up every refresh...

lewisgoddard:
Fixes #3601

### Prepositions
- HTTP-only connections from the web should be being redirected by the webserver.
- Local testing in HTTP-only environments does not break the site in any substantial way.
- beta.elementary.io no longer deploys testing branches, we have largely moved to local testing.

### Changes Summary

- Remove legacy warning for HTTP.
- Remove legacy warning for non-production environments.
- Remove branch support from `page.js`


NickColley:
If I could approve the pull request I would, nice work. 👍🏻

kgrubb (approved):

PR #3690, opened 2024-12-07
feat: add /api/sponsors_goal endpoint

davidmhewitt:
Adds an API endpoint to return the current GitHub sponsors goal and the percentage towards it. Example response:

```
{"percent":23,"target":7500}
```

Means we can keep the GitHub token in config on the server and not expose it in a compiled package on the system.

### DIFF
--- a/_backend/config.example.php
+++ b/_backend/config.example.php
@@ -21,6 +21,9 @@ return array(
   'stripe_sk' => 'sk_test_hoigesrjgoisrhgilgjrsfjs',
   'stripe_pk' => 'pk_test_hoigesrjgoisrhgilgjrsfjs',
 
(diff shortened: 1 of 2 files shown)


alainm23:
Great, is there anything else missing to approve this PR? Honestly, I don’t know PHP, and I’ve always refused to learn it 😅️.

davidmhewitt:
This should be ready to go, I've tested locally and confirmed it's working. I don't know if someone like @lewisgoddard wants to review the code.

Ideally we would have some caching here as this data won't change so often so we don't need to call the GitHub API every time, maybe just every hour or so. But I can configure some caching in cloudflare to cache the responses if necessary.

lewisgoddard (approved):

Issue #3693, opened 2024-12-10
Replace Translations.svg with a fixed world map


Kaleidosium:
### What Happened?

The world map in https://github.com/elementary/website/blob/master/_images/get-involved/translations.svg shows a bugged out map where the Borneo territories of Indonesia and Malaysia are separated from one another, this SVG is the culprit.

Perhaps we could replace it with either of these?
- https://en.wikipedia.org/wiki/File:Simple_world_map.svg (CC0)
- https://commons.wikimedia.org/wiki/File:WorldMap-Blank-Noborders.svg (CC0)

### Steps to Reproduce

1. View the SVG in the latest Firefox or Chrome

### Expected Behavior

Viewing a World Map where Borneo isn't separated like this.

PR #3739, opened 2025-03-05
Fix translations.svg

randomints:
Fixes #
this fixes issue #3693 

### Changes Summary

- I replaced translations.svg with a more accurate .png image


tintou:
This has to be an .svg

tintou:
Why do you think that we need an very accurate image here? What is the issue you are trying to fix?

randomints:
I changed it to an svg

randomints:
I'm fixing #3693 because I'd like some green squares on my github calendar

PR #3766, opened 2025-04-26
Update Shortcuts Overlay tip in Learning the Basics

1280px:
Very small change, since Super key now opens apps menu by default.


ryonakano (approved):
Good catch!

lewisgoddard (approved):

PR #3824, opened 2025-07-18
Update icons for HIG

wpkelso:
Related: elementary/hig#54
### Changes Summary

- Duplicated relevant icons from the icons repo
- Remade examples with current icons

Linked with elementary/hig#62


danirabbit (changes requested):
- [ ] It looks like `contrast-example2.png` is the wrong size of image. The goal of this section is to show the stroke style and it looks like we overlay a grid. The replacement image here has strokes that are half as wide as the original image. I'm guessing that the 128px image was exported at 4x here in the replacement, but it's probably 64px @ 8x in the original
- [ ] in `example-icon1.png`, the text here is about showing the concept of the "mean line" across different icons with different x heights. The arrow in the replacement icon no longer aligns on this mean line, so it makes the accompanying text not make sense. We probably should just get rid of this "mean line" section since I don't think we really adhere to that anymore, but either way this replacement image doesn't really make sense here
- [ ] We're missing the replacements for `shadow-example1.png` etc that go with the updated Music icon. I'm okay with not updating this if you intend to remove that section in anticipation of https://github.com/elementary/icons/issues/1353

wpkelso:
@danirabbit Yes, `shadow-example1.png` was left as-is in anticipation of the shadow's removal. Currently, elementary/icons#1362 exists to solidify that, but as it's not been merged yet, I've left the shadows in all the icons updated here.

I wasn't intending to fully remove sections yet, just leave them as vestigial here and come back to them later, but perhaps it's better to change the coolant while we're changing the oil, so to speak. I can make a proposal to remove/deprecate relevant sections from the HIG.

danirabbit (approved):
I'm happy to merge this now, and then yeah I think we should probably remove the section about the "mean line" as a follow up

danirabbit:
Thanks for updating these!

PR #3920, opened 2025-12-13
Add download options, update readme to add building and running instr…

### LINKED ISSUE #3845: Architecture selector
### Problem

We'll soon have UEFI ARM64 images, but we currently don't have a way in the downloads modal to select processor architecture 

### Proposal

We need to add some kind of interface for choose an ARM64 or AMD64 download

### Prior Art (Optional)

Fedora's download page has a pretty nice little boxed list thing going on: https://fedoraproject.org/workstation/download

barunespadhy:
Fixes #3845

Tried to create a modal box to support downloading the new architecture.

**Note:** The download and magnet links for ARM 64 has to be added, currently it has the AMD 64's link as a placeholder.

Screenshot of view:

image


barunespadhy:
@NickColley Restructured to use the new system :)
It looks closer to the current dialog box now:

image

Where can I find the configs for the arm release details? Currently it shows amd for both, because i couldnt find the config names for arm.

PR #4136, opened 2026-08-30
Get Involved: remove PayPal

danirabbit:
PayPal banned our account for "suspicious activity". Folks can still sponsor us with PayPal through Liberapay so update the copy there

### DIFF
--- a/get-involved.php
+++ b/get-involved.php
@@ -52,7 +52,6 @@ include $template['alert'];
             <a class="button flat" href="https://www.patreon.com/elementary" target="_blank" rel="noopener">Back on Patreon</a>
         </div>
     </div>
-
   
     <div class="grid">
         <div class="half">
@@ -61,7 +60,6 @@ include $template['alert'];
             <a class="button flat" title="Liberapay" href="https://liberapay.com/elementary/" target="_blank" rel="noopener">Contribute with Liberapay</a>
         </div>
     </div>
-
 </section>
 
 <section id="translations">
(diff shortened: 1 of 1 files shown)


get-involved.php
```diff
@@ -54,19 +54,9 @@
...
-            <input type="hidden" name="hosted_button_id" value="LG382EHQVTDYN">
-            <input type="submit" value="Use PayPal" name="submit" title="PayPal - The safer, easier way to pay online!" class="button flat">
-            <img alt="" border="0" src="https://www.paypalobjects.com/en_US/i/scr/pixel.gif" width="1" height="1">
-            </form>
-        </div>
-        <div class="third">
+        <div class="half">
             <img class="logo" alt="Liberapay" src="images/get-involved/liberapay_logo_black.svg">
-            <p>Set up a recurring contribution through Liberapay, the open source and non-profit funding platform.</p>
+            <p>Set up sponsorships with debit, credit, SEPA, or PayPal through Liberapay: the open source and non-profit funding platform.</p>
```
RMcNeely: Shouldn't we be removing the PayPal reference here?

danirabbit: *our* PayPal account is disabled, but if *you* have a PayPal account you can still use it through Liberapay instead of to our now banned account

danirabbit: I think, lemme double check

danirabbit: Okay nvm you're right, no PayPal through Liberapay without our own paypal account

lewisgoddard (approved):

